PR #14, opened 2022-02-15
chore: Add customer facing error

### LINKED ISSUE #7: Handle gRPC status codes (e.g. `RESOURCE_EXHAUSTED`)
Currently, we're not exposing errors to customers but we need to create customer facing errors and pull it up to the top level (under the `momento` package).

This is how the Java SDK handles customer facing error: https://github.com/momentohq/client-sdk-java/issues/106
Python example: https://github.com/momentohq/client-sdk-python/issues/27
Reference how AWS exposes error: https://github.com/aws/aws-sdk-go/tree/main/aws/awserr

poppoerika:
This is a follow-up pr for https://github.com/momentohq/client-sdk-go/pull/11
@eaddingtonwhite I wasn't 100% sure how I can create customer facing error based on `MomentoSvcErr`, so I brought up `MomentoError` to the top level (`momento`). So your feedback would be much appreciated. 

Closes #7


momento/error.go
```diff
@@ -0,0 +1,62 @@
...
+	PermissionDenied     = "PermissionDenied"
+	Unauthenticated      = "Unauthenticated"
+	ResourceExhausted    = "ResourceExhausted"
+	NotFound             = "NotFound"
+	AlreadyExists        = "AlreadyExists"
+	Unavailable          = "Unavailable"
+)
+
+type MomentoError interface {
+	momentoerrors.MomentoSvcErr
```
eaddingtonwhite, line 25: Rule thumb when making sdk is any time you are returning something to user from internal package something is not right. You should have new fields here. Check out https://github.com/aws/aws-sdk-go/blob/main/aws/awserr/error.go#L31-L44 for inspo.

Then can have some helper mehtods as well to help you convert.

poppoerika:
@eaddingtonwhite - I'd like to close https://github.com/momentohq/client-sdk-go/pull/11 since this pr is evolved from that pr, so we can just merge this pr to reflect both changes.

momento/simple_cache_client_test.go
```diff
@@ -81,7 +77,7 @@ func TestCreateCacheGetSetValueAndDeleteCache(t *testing.T) {
 		t.Error(err.Error())
 	}
 
-	_, err = client.Set(&CacheSetRequest{
+	_, err := client.Set(&CacheSetRequest{
```
eaddingtonwhite, line 80: can we have one test where we try parsing from raw error object returned from sdk via interface make sure this works as expected.

Similar to how you commonly do w/ aws sdk
```go
if err != nil {
    if awsErr, ok := err.(awserr.Error); ok {
        // process SDK error
    }
}
```

Would expect something like this.
```go
if err != nil {
    if momentoErr, ok := err.(momento.Error); ok {
        assert.Equal(t, InvalidArgumentError, err.Code())
    }else{
        t.Errorf("unexpected error: %v", err)
    }
}
```

eaddingtonwhite (commented):
@poppoerika this is looking alot better. Can we just add one more unit test making sure can marshal nicely into error interface. After that I think this looks good to merge.

momento/simple_cache_client_test.go
poppoerika: It seems that the linter does not like we are doing type assertion on the same type...
image

eaddingtonwhite: ahh I see. I think we should still just return type error from our sdk functions not type MomentoError.  I think that is pretty common convention in golang  I think maybe cause there are scenarios where core golang error types may need to be returned. Not totally sure on this I think lets just copy how AWS did though with their sdks and return error for now from all our signatures. I think all the main SDKs ive used return just `error` not their higher level wrapper.

This will get rid of this link warning and force you to handle marshaling error via interface similar to example above.

So would now be like
`func (client *ScsClient) Set(request *CacheSetRequest) (*SetCacheResponse, error) {`
https://github.com/momentohq/client-sdk-go/blob/5a767331c78c90e17ad1c7cc339c371726cd4dac/momento/simple_cache_client.go#L99

poppoerika: I got rid of the specific test for type assertion since there are already two places testing this: 
https://github.com/momentohq/client-sdk-go/blob/b7199087d95ffd6d6bf2234ae66cab4ec4c34f7b/momento/simple_cache_client_test.go#L52

https://github.com/momentohq/client-sdk-go/blob/b7199087d95ffd6d6bf2234ae66cab4ec4c34f7b/momento/simple_cache_client_test.go#L131

momento/simple_cache_client_test.go
```diff
@@ -130,7 +128,9 @@ func TestZeroRequestTimeout(t *testing.T) {
 		DefaultTtlSeconds:     DefaultTtlSeconds,
 		RequestTimeoutSeconds: &timeout,
 	})
-	if assert.Error(t, err) {
-		assert.Equal(t, "InvalidInputError: Request timeout must be greater than zero.", err.Error())
+	if momentoErr, ok := err.(MomentoError); ok {
```
eaddingtonwhite, line 131: nice!!

eaddingtonwhite (approved):

PR #15, opened 2022-02-16
feat: usability improvements and testing updates

### LINKED ISSUE #16: Update `README` Momento Usage
With the updates made in https://github.com/momentohq/client-sdk-go/pull/15, need to update the Momento usage example.
https://github.com/momentohq/client-sdk-examples/issues/71

eaddingtonwhite:
@poppoerika I sat down played with sdk a bit more after latest error changes and made few tweaks to help with end user usability and code readability in few spots. Also tried to write good example of table test set up for reference. Open to changing any parts so please let me know what you think or if we can improve further after looking over.

Closes #16

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,11 @@ Go SDK for Momento, a serverless cache that automatically scales without any of 
 
 - [Go version 1.17.\*](https://go.dev/dl/)
 - A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
+- golang
+  - `brew install go`
+- golint
+  - `go get -u golang.org/x/lint/golint`
+
 
 <br/>
 
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 :warning: Experimental SDK :warning:
 
-Go SDK for Momento is experimental and under active development.
-There could be non-backward compatible changes or removal in the future.
-Please be aware that you may need to update your source code with the current version of the SDK when its version gets upgraded.
+Go SDK for Momento is experimental and under active development. There could be non-backward compatible changes or
+removal in the future. Please be aware that you may need to update your source code with the current version of the SDK
+when its version gets upgraded.
 
 ---
 
@@ -14,7 +14,8 @@ Please be aware that you may need to update your source code with the current ve
     <img src="images/gopher.png" alt="Logo" width="200" height="150">
 </div>
 
-Go SDK for Momento, a serverless cache that automatically scales without any of the operational overhead required by traditional caching solutions.
+Go SDK for Momento, a serverless cache that automatically scales without any of the operational overhead required by
+traditional caching solutions.
 
 <br/>
 
@@ -23,12 +24,12 @@ Go SDK for Momento, a serverless cache that automatically scales without any of 
 ## Requirements
 
 - [Go version 1.17.\*](https://go.dev/dl/)
-- A Momento Auth Token is required, you can generate one using the [Momento CLI](https://github.com/momentohq/momento-cli)
+- A Momento Auth Token is required, you can generate one using
+  the [Momento CLI](https://github.com/momentohq/momento-cli)
 - golang
-  - `brew install go`
+    - `brew install go`
 - golint
-  - `go get -u golang.org/x/lint/golint`
-
+    - `go get -u golang.org/x/lint/golint`
 
 <br/>
 
@@ -41,18 +42,21 @@ Check out our [Go SDK example repo](https://github.com/momentohq/client-sdk-exam
 ## Using Momento
 
 ```go
+package main
+
 import (
+	"fmt"
 	"log"
 	"os"
 
 	"github.com/google/uuid"
-	"github.com/momentohq/client-sdk-go/momento"
+	. "github.com/momentohq/client-sdk-go/momento"
 )
 
 func main() {
 	authToken := os.Getenv("MOMENTO_AUTH_TOKEN")
 	const (
-		cacheName             = "cache"
+		cacheName             = "my-first-cache-😊"
 		itemDefaultTtlSeconds = 60
 	)
 
@@ -61,51 +65,71 @@ func main() {
 	}
 
 	// Initializes Momento
-	client, err := momento.NewSimpleCacheClient(authToken, itemDefaultTtlSeconds)
+	client, err := NewSimpleCacheClient(authToken, itemDefaultTtlSeconds)
 	if err != nil {
-		panic(err)
+		log.Fatal(fmt.Sprintf("failed initilizing cache client with err %+v", err))
 	}
 
-	// Create Cache and check if CacheName exists
-	err = client.CreateCache(&momento.CreateCacheRequest{
-		CacheName: CacheName,
+	// Create Cache Ignore if Cache already exists
+	err = client.CreateCache(&CreateCacheRequest{
+		CacheName: cacheName,
 	})
-	if momentoErr, ok := err.(momento.MomentoError); ok {
-        if momentoErr.Code() != AlreadyExists {
-            return nil, err
-        }
+	if err != nil {
+		if momentoErr, ok := err.(MomentoError); ok {
+			if momentoErr.Code() != AlreadyExists {
+				log.Fatal(fmt.Sprintf(
+					"failed creating cache %s with err %+v",
+					cacheName, momentoErr,
+				))
+			}
+		}
 	}
-	log.Printf("Cache named %s is created\n", CacheName)
+	log.Println(fmt.Sprintf("Cache %s is created", cacheName))
 
-	// Sets key with default TTL and gets value with that key
-	key := []byte(uuid.NewString())
-	value := []byte(uuid.NewString())
-	log.Printf("Setting key: %s, value: %s\n", key, value)
-	_, err = client.Set(&momento.CacheSetRequest{
-		CacheName: CacheName,
-		Key:       key,
-		Value:     value,
-	})
-	if err != nil {
-		panic(err)
-	}
-	log.Printf("Getting key: %s\n", key)
-	resp, err := client.Get(&momento.CacheGetRequest{
-		CacheName: CacheName,
-		Key:       key,
-	})
-	if err != nil {
-		panic(err)
-	}
-	log.Printf("Lookup resulted in a : %s\n", resp.Result())
-	log.Printf("Looked up value: %s\n", resp.StringValue())
+	// Sets key with default TTL and custom ttl and then retrieves the items from cache
+	key1 := []byte(uuid.NewString())
+	key2 := []byte(uuid.NewString())
+	value1 := []byte(uuid.NewString())
+	value2 := []byte(uuid.NewString())
 
-	// Permanently delete the cache
-	err = client.DeleteCache(&momento.DeleteCacheRequest{CacheName: CacheName})
+	log.Println(fmt.Sprintf("Setting key: %s, value: %s", key1, value1))
+	_, err = client.Set(&CacheSetRequest{
+		CacheName: cacheName,
+		Key:       key1,
+		Value:     value2,
+	})
 	if err != nil {
-		panic(err)
+		log.Fatal(fmt.Sprintf("failed setting key %s with err %+v", key1, err))
 	}
-	log.Printf("Cache named %s is deleted\n", CacheName)
+	log.Println(fmt.Sprintf("Setting key with custom ttl key: %s, value: %s", key2, value2))
+	_, err = client.Set(&CacheSetRequest{
+		CacheName:  cacheName,
+		Key:        key2,
+		Value:      value2,
+		TtlSeconds: TTL(60),
+	})
+	if err != nil {
+		log.Fatal(fmt.Sprintf("failed setting key %s with err %+v", key2, err))
+	}
+	log.Println(fmt.Sprintf("Getting key: %s", key1))
+	resp, err := client.Get(&CacheGetRequest{
+		CacheName: cacheName,
+		Key:       key1,
+	})
+	if err != nil {
+		log.Fatal(fmt.Sprintf("failed getting key %s with err %+v", key1, err))
+	}
+	log.Println(fmt.Sprintf(
+		"Get request succeded result: %s value: %s",
+		resp.Result(), resp.StringValue(),
+	))
+
+	// Delete the cache
+	err = client.DeleteCache(&DeleteCacheRequest{CacheName: cacheName})
+	if err != nil {
+		log.Fatal(fmt.Sprintf("failed deleting cache %s with err %+v", cacheName, err))
+	}
+	log.Printf(fmt.Sprintf("Cache %s is deleted", cacheName))
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -50,30 +50,30 @@ import (
 )
 
 func main() {
-	var AuthToken = os.Getenv("MOMENTO_AUTH_TOKEN")
+	authToken := os.Getenv("MOMENTO_AUTH_TOKEN")
 	const (
-		CacheName             = "cache"
-		ItemDefaultTtlSeconds = 60
+		cacheName             = "cache"
+		itemDefaultTtlSeconds = 60
 	)
 
-	if AuthToken == "" {
+	if authToken == "" {
 		log.Fatal("Missing required environment variable MOMENTO_AUTH_TOKEN")
 	}
 
 	// Initializes Momento
-	client, err := momento.SimpleCacheClient(&momento.SimpleCacheClientRequest{
-		AuthToken:         AuthToken,
-		DefaultTtlSeconds: ItemDefaultTtlSeconds,
-	})
+	client, err := momento.NewSimpleCacheClient(authToken, itemDefaultTtlSeconds)
 	if err != nil {
 		panic(err)
 	}
+
 	// Create Cache and check if CacheName exists
 	err = client.CreateCache(&momento.CreateCacheRequest{
 		CacheName: CacheName,
 	})
-	if err != nil && !strings.Contains(err.Error(), "AlreadyExists") {
-		panic(err)
+	if momentoErr, ok := err.(momento.MomentoError); ok {
+        if momentoErr.Code() != AlreadyExists {
+            return nil, err
+        }
 	}
 	log.Printf("Cache named %s is created\n", CacheName)
 
--- a/internal/resolver/end_point_resolver.go
+++ b/internal/resolver/end_point_resolver.go
@@ -10,10 +10,10 @@ import (
 )
 
 const (
-	MOMENTO_CONTROL_ENDPOINT_PREFIX = "control."
-	MOMENTO_CACHE_ENDPOINT_PREFIX   = "cache."
-	CONTROL_ENDPOINT_CLAIM_ID       = "cp"
-	CACHE_ENDPOINT_CLAIM_ID         = "c"
+	momentoControlEndpointPrefix = "control."
+	momentoCacheEndpointPrefix   = "cache."
+	controlEndpointClaimId       = "cp"
+	cacheEndpointClaimId         = "c"
 )
 
 type Endpoints struct {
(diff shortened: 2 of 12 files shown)


internal/resolver/end_point_resolver.go
```diff
@@ -10,10 +10,10 @@ import (
...
 
 const (
-	MOMENTO_CONTROL_ENDPOINT_PREFIX = "control."
-	MOMENTO_CACHE_ENDPOINT_PREFIX   = "cache."
-	CONTROL_ENDPOINT_CLAIM_ID       = "cp"
-	CACHE_ENDPOINT_CLAIM_ID         = "c"
+	momentoControlEndpointPrefix = "control."
+	momentoCacheEndpointPrefix   = "cache."
+	controlEndpointClaimId       = "cp"
+	cacheEndpointClaimId         = "c"
```
poppoerika, line 16: For Golang, it's okay to use just camel case to name constant variables?

momento/simple_cache_client.go
```diff
@@ -7,140 +7,144 @@ import (
 	"github.com/momentohq/client-sdk-go/internal/services"
 )
 
-const DeafultDataCtxTimeout = 5
+const defaultRequestTimeout = uint32(5)
 
-type ScsClient struct {
+type ScsClient interface {
```
poppoerika, line 12: I'm still try to grasp the interface usage... 
What is the purpose of using interface here?

eaddingtonwhite: This will make it much easier for end user to do proper testing on their end. By exporting an interface then a customer can make a `type MockScsClient struct {}` that implements our client interface and then they pass in to their code for unit testing to override certain errors or responses our SDK would return in their unit tests. Since golang is not the most flexible language by design having really good interfaces on any class or package that you want to be able to mock eventually is really important. This is generally how we do unit testing w/ our kotlin code as well instead of using a framework like mockitto for overiding sdk functionality in kotlin unit tests.

internal/resolver/end_point_resolver.go
eaddingtonwhite: yeah my IDE was yelling at me about the SCREAMING_SNAKE_CASE just a golang style thing. ¯\_(ツ)_/¯

momento/simple_cache_client_test.go
```diff
@@ -2,104 +2,97 @@ package momento
...
+	return client, nil
 }
 
-func cleanUp(client *ScsClient) {
-	err := client.Close()
-	if err != nil {
-		log.Fatal(err.Error())
-	}
+func cleanUp(client ScsClient) {
+	_ = client.Close()
```
poppoerika, line 51: We don't need to worry about closing client connections failure in the test?

momento/simple_cache_client_test.go
```diff
@@ -2,104 +2,97 @@ package momento
...
-	client        *ScsClient
-	err           error
 )
 
 const (
-	DefaultTtlSeconds = 60
+	defaultTtlSeconds = 60
 )
 
-func TestMain(m *testing.M) {
```
poppoerika, line 24: We don't want to use TestMain?
Is it because this has no flexibility to when to run `setUp` and  `cleanUp`? (I believe TestMain always run `setUp` once at the beginning and run `cleanUp` at the end before all the tests in this file, which means that we can run it every time before each tests starts/ends)

momento/simple_cache_client_test.go
```diff
@@ -109,28 +102,80 @@ func TestCreateCacheGetSetValueAndDeleteCache(t *testing.T) {
...
-	_, err := SimpleCacheClient(&SimpleCacheClientRequest{
-		AuthToken:             TestAuthToken,
-		DefaultTtlSeconds:     DefaultTtlSeconds,
-		RequestTimeoutSeconds: &timeout,
-	})
-	if momentoErr, ok := err.(MomentoError); ok {
-		assert.Equal(t, InvalidArgumentError, momentoErr.Code())
-	} else {
-		t.Errorf("Unexpected error: %v", err)
+func TestClientInitialization(t *testing.T) {
```
poppoerika, line 123: Ohh, this is really nice!

momento/simple_cache_client.go
```diff
@@ -7,140 +7,144 @@ import (
...
+		return nil, convertMomentoSvcErrorToCustomerError(momentoerrors.ConvertSvcErr(err))
 	}
-	err = validateRequestTimeout(request.RequestTimeoutSeconds)
-	if err != nil {
-		return nil, NewMomentoError(err.Code(), err.Message(), err.OriginalErr())
+
+	requestTimeoutToUse := defaultRequestTimeout
+	if request.RequestTimeoutSeconds > 1 {
+		requestTimeoutToUse = request.RequestTimeoutSeconds
 	}
```
poppoerika, line 48: I believe Python and other SDKs throw an exception if the provided timeout is 0 or less.
For out case, we're setting the default timeout if the provided timeout is 0 or 1 without warning the user?

momento/simple_cache_client_test.go
```diff
@@ -109,28 +102,80 @@ func TestCreateCacheGetSetValueAndDeleteCache(t *testing.T) {
...
+			}
+
+			if tt.expectedErr != "" && err != nil {
+				if momentoErr, ok := err.(MomentoError); ok {
+					if momentoErr.Code() == tt.expectedErr {
+						return // Success end test we expected this
+					}
+				}
+				t.Errorf(
+					"unexpected error occured initilizing client got=%+v expected=%+v",
```
poppoerika, line 171: `initializing`

momento/simple_cache_client_test.go
```diff
@@ -109,28 +102,80 @@ func TestCreateCacheGetSetValueAndDeleteCache(t *testing.T) {
...
+			TestCacheName, existingCacheResp.StringValue(),
+		)
 	}
 
 	err = client.DeleteCache(&DeleteCacheRequest{
 		CacheName: cacheName,
 	})
 	if err != nil {
-		t.Error(err.Error())
+		t.Errorf("error occured deleting cache err=%+v", err)
```
poppoerika, line 117: `occurred`

momento/simple_cache_client_test.go
```diff
@@ -2,104 +2,97 @@ package momento
...
 
 // Basic happy path test - create a cache, operate set/get, and delete the cache
-func TestCreateCacheGetSetValueAndDeleteCache(t *testing.T) {
+func TestBasicHappyPathSDKFlow(t *testing.T) {
 	cacheName := uuid.NewString()
 	key := []byte(uuid.NewString())
 	value := []byte(uuid.NewString())
-
+	client, err := setUp()
+	if err != nil {
```
poppoerika, line 60: `occurred`

momento/simple_cache_client_test.go
```diff
@@ -2,104 +2,97 @@ package momento
...
+		t.Error(fmt.Errorf("error occured setting up client err=%+v", err))
```
poppoerika, line 61: `occurred`

momento/simple_cache_client_test.go
```diff
@@ -2,104 +2,97 @@ package momento
...
+	}
 	err = client.CreateCache(&CreateCacheRequest{
 		CacheName: cacheName,
 	})
 	if err != nil {
-		t.Error(err.Error())
+		t.Error(fmt.Errorf("error occured creating cache err=%+v", err))
```
poppoerika, line 67: `occurred`

momento/simple_cache_client_test.go
```diff
@@ -2,104 +2,97 @@ package momento
...
 
-	_, err := client.Set(&CacheSetRequest{
+	_, err = client.Set(&CacheSetRequest{
 		CacheName: cacheName,
 		Key:       key,
 		Value:     value,
 	})
 	if err != nil {
-		t.Error(err.Error())
+		t.Errorf("error occured setting key err=%+v", err)
```
poppoerika, line 76: `occurred`

momento/simple_cache_client_test.go
poppoerika: This is what you are mentioning last night about Golang common test pattarn?

eaddingtonwhite: yes! https://dave.cheney.net/2019/05/07/prefer-table-driven-tests <- good quick read. Dave Cheney is great

momento/simple_cache_client.go
eaddingtonwhite: Hmm so first off good catch. What I was intending here was `> 0` not 1.  I felt the optional *unit32 added more pain then was worth. So switched to just uint32 So now it defaults to 0 under the hood if not set and its impossible to tell if user passed or if it just wasnt set. Before logic was set to only fail if was less then 1  and because we use uint32 for type now they cant pass negative number or something bad.

I think the scenario that concerns me is user explicitly passes 0 expecting idk what exactly and then we just silently use default request timeout under the hood....

curious maybe @gautamomento thoughts here also. I can try adding back to throw error if they pass 0 but we will need to go back to *uint32 then vs uint32 pretty sure.

momento/simple_cache_client_test.go
eaddingtonwhite: I was having problems with client not being initilized and was null when tests were running I think maybe because of parallel tests... was not 100 % sure on this though. I think we may need to set instance of client for each test vs a global one. I was getting pretty tired towards end though let me revisit this and play a bit more this morning with this will let u know.

momento/simple_cache_client_test.go
eaddingtonwhite: idk just seemed like over kill in the moment. but looking again yeah maybe we should make sure exits cleanly to ur point. I can re-add check

momento/simple_cache_client.go
poppoerika: Yes, so the reason why I used a pointer to pass (even though it felt weird since that was the only field in the struct to pass as a pointer) to distinguish actual 0 timeout passed by the user and default 0 set by the struct.

>  user explicitly passes 0 expecting idk what exactly

I just followed other SDK patters to check if user-provided timeout or default timeout is 0 but I felt the same as your comment.

momento/simple_cache_client_test.go
poppoerika: Okay! Sounds good.

momento/simple_cache_client.go
eaddingtonwhite: idk I think my preference would be to just keep this not as pointer and wait and see if any real customers have issue with this approach? Could be convinced otherwise for sure though. Lets let @gautamomento make final call since he has been owning e2e sdk experience. Im fine either way but slight preference to no pointer I think after playing w/ sdk last night.

momento/simple_cache_client_test.go
poppoerika: For this test, we don't need to close control/data client connections?

eaddingtonwhite: probably should... let me look into that

momento/simple_cache_client_test.go
poppoerika: 🙇‍♀️

poppoerika:
@eaddingtonwhite Thank you for all the improvement changes!
The SDK looks much better and more user friendly. 
After this gets merged and re-published, I want to work on updating the README Momento usage for this repo and the SDK example in the client-sdk-examples repo.
Created tickets:
- `client-sdk-go` repo https://github.com/momentohq/client-sdk-go/issues/16
- `client-sdk-examples` repo https://github.com/momentohq/client-sdk-examples/issues/71

momento/simple_cache_client.go
gautamomento: Would it be any different if we used some `time.Duration` class assuming that it is a first class citizen in Golang? That way we have clear indication of whether it is provided or not. 

I have chosen Duration class in Java as it is a first class citizen  - customers can do whatever they prefer millis, nanos, seconds, hours etc. I would prefer at some point to make our ttls also comply to this but that is a topic for some other time.

poppoerika: It seems that `time.Duration` results in `int64` so if the user doesn't pass anything in `timeout` field, it'll be 0 and still not way for us to distinguish the user provided 0 from default value 0?
https://pkg.go.dev/time#Duration

gautamomento: I am not sure what the right way out here. Not entering anything and using defaults make sense to me. 

Entering zero (even when, you shouldn't be) and then using defaults just seems wrong to me.  I am in favor of whatever mechanism we need to use to not make this happen.

poppoerika: @eaddingtonwhite With Gautam's feedback, do we want to go back to the pointer `timeout` so that we can check if the user passed 0 or nil?

eaddingtonwhite: per our zoom call have changed up how this works per that convo. Going to resolve this thread and we can open new comments on new approach.

momento/simple_cache_client_test.go
eaddingtonwhite: @poppoerika tried having clean up try always close and fail if got error. The problem I think with this approach is that there are some cases where client will close will always fail. I think would make more sense to have an explicit test for calling Close in different scenarios then trying to assert always closes properly in our generic test tear down step. I think we can add that explicit test in follow up pr. WDUT?

momento/simple_cache_client_test.go
eaddingtonwhite: Ok played w/ this a bit and was not finding great way to do across multiple tests especially that works when running tests in parallel. I will need to play with this some more. I renamed set up and tear down functions a bit more though so was bit cleaner. I think we should revisit the TestMain functionality again in another follow up pr.

momento/simple_cache_client_test.go
eaddingtonwhite: see other comment about tear down and validating failures in here generically.

eaddingtonwhite: @poppoerika tried having clean up try always close and fail if got error. The problem I think with this approach is that there are some cases where client will close will always fail. I think would make more sense to have an explicit test for calling Close in different scenarios then trying to assert always closes properly in our generic test tear down step. I think we can add that explicit test in follow up pr. WDUT?

momento/simple_cache_client_test.go
eaddingtonwhite: added close to this test

momento/simple_cache_client_test.go
eaddingtonwhite: ok played with this for a while and was having trouble getting to work now w/ test running in parallel and initing client properly. I think we should maybe leave how I have for now  and can revisit TestMain in follow up pr. I did try to clean up names of set up and tear down functions to make it bit more clear.

README.md
```diff
@@ -45,30 +50,30 @@ import (
 )
 
 func main() {
-	var AuthToken = os.Getenv("MOMENTO_AUTH_TOKEN")
+	authToken := os.Getenv("MOMENTO_AUTH_TOKEN")
```
poppoerika, line 53: Thank you for updating the example!

momento/simple_cache_client_test.go
```diff
@@ -2,135 +2,195 @@ package momento
...
 	if err != nil {
-		t.Error(err.Error())
+		t.Errorf("error occurred setting key err=%+v", err)
+	}
+
+	_, err = client.Set(&CacheSetRequest{
+		CacheName:  cacheName,
+		Key:        uuid.NewString(),
+		Value:      value,
+		TtlSeconds: TTL(1),
```
eaddingtonwhite, line 51: example using new custom `TTL` helper func and object to wrap `*uint32` on set call. Definetly like this approach more for end user I think.

 @gautamomento @poppoerika

momento/simple_cache_client.go
```diff
@@ -7,140 +7,187 @@ import (
...
+
+type DefaultScsClient struct {
+	authToken             string
+	controlClient         *services.ScsControlClient
+	dataClient            *services.ScsDataClient
+	defaultTtlSeconds     uint32
+	defaultRequestTimeout uint32
+}
+
+type Option func(*DefaultScsClient) MomentoError
```
poppoerika, line 31: How does this function work when it's called [here](https://github.com/momentohq/client-sdk-go/pull/15/files#diff-7aaa7ca147a29523feaeed77cc4fc404a052560b64fd304d1e075dcd6e024a62R72)?
It checks if it only contains the fields declared [here](https://github.com/momentohq/client-sdk-go/pull/15/files#diff-7aaa7ca147a29523feaeed77cc4fc404a052560b64fd304d1e075dcd6e024a62R23-R29)?

momento/simple_cache_client.go
```diff
@@ -7,140 +7,187 @@ import (
...
 	endpoints, err := resolver.Resolve(&models.ResolveRequest{
-		AuthToken: request.AuthToken,
+		AuthToken: authToken,
 	})
 	if err != nil {
-		return nil, NewMomentoError(err.Code(), err.Message(), err.OriginalErr())
+		return nil, convertMomentoSvcErrorToCustomerError(err)
+	}
+
+	if defaultTtlSeconds < 1 {
```
poppoerika, line 55: Do we not want to allow 0 for `defaultTTlSeconds` based on Gatuam's comment [here](https://gomomento.slack.com/archives/C034513MJ1E/p1645056278635209)?

momento/simple_cache_client_test.go
poppoerika: Nice! With this way, users don't need to dealt with passing an uint32 pointer.

momento/simple_cache_client.go
```diff
@@ -7,140 +7,187 @@ import (
...
+				nil,
+			)
+		}
+		c.defaultRequestTimeout = requestTimeout
+		return nil
+	}
 }
 
-func SimpleCacheClient(request *SimpleCacheClientRequest) (*ScsClient, error) {
+func NewSimpleCacheClient(authToken string, defaultTtlSeconds uint32, opts ...Option) (ScsClient, error) {
```
poppoerika, line 47: I think this is just me not understanding Golang well but how do users know they can pass `WithRequestTimeout` as `...Option`?

momento/simple_cache_client.go
eaddingtonwhite: The Option functions just are a nice way to be able to have a set of optional "setter" functions passed into a constructor. You can see NewSimpleCacheClient now takes a list of `opts ...Option` leveraging Varadic Functions https://gobyexample.com/variadic-functions  

Since go doesnt support method overload this is common pattern you will see to allow you to pass in optional configuration settings into a client package you are developing.

So we just have one Option now `WithRequestTimeout` which we check for if it was set after applying any passed options to initilized client here
https://gobyexample.com/variadic-functions

here is example of me using this option in tests
https://github.com/momentohq/client-sdk-go/pull/15/files#diff-adbc2340f9c72f714ae731985c022e0b1ab31072d90f6409916df27d4010bad5R143


Down line we will support more Options for user to be pass us on init. Things like `WithLogger`, `WithMetrics` etc...

momento/simple_cache_client.go
eaddingtonwhite: Was thinking about this more. I get it for us allowing for a Set request but seems weird to me to allow you to set it for overall client default.... I can remove though just felt like we shouldnt allow overall defaultTTL to be 0. @gautamomento wdut?

momento/simple_cache_client.go
eaddingtonwhite: We need to add proper godocs to package and better examples showing different options available when init client in package. `WithRequestTimeout` is exported out of this package so users should be able to discover and use. Once we add proper godocs to all exported parts of this package will show up nicely on our docs page with docs explain what each Option function does.

Check out Config sub package of aws sdk for good example what might look like for us
https://pkg.go.dev/github.com/aws/aws-sdk-go/aws#Config 

Can see a similar pattern w/ lots of `With*` functions exported to be used as optional config settings.

momento/simple_cache_client.go
gautamomento: I would like to keep this consistent across SDKs. Let us allow zero and we can revisit this conversation and just apply the change in one go for all SDKs

momento/simple_cache_client_test.go
gautamomento: Love this.

gautamomento (commented):
This is looking nice. 

Just one comment around zero TTL but otherwise this LGTM.

PR #21, opened 2022-02-18
chore: Add documentation for momento package

### LINKED ISSUE #17: Add documentation
Useful link: https://elliotchance.medium.com/godoc-tips-tricks-cda6571549b

poppoerika:
These documentation will be reflected to our package [here](https://pkg.go.dev/github.com/momentohq/client-sdk-go@v0.2.1/momento#section-documentation).

Closes #17

### DIFF
--- a/momento/cache_operation_responses.go
+++ b/momento/cache_operation_responses.go
@@ -1,36 +1,45 @@
 package momento
 
+// Initializes ListCacheResponse to handle list cache response.
 type ListCachesResponse struct {
 	nextToken string
 	caches    []CacheInfo
 }
 
+// Returns next token.
 func (resp *ListCachesResponse) NextToken() string {
 	return resp.nextToken
 }
 
+// Returns all caches.
 func (resp *ListCachesResponse) Caches() []CacheInfo {
 	return resp.caches
 }
 
+// Initializes CacheInfo to handle caches returned from list cache operation.
 type CacheInfo struct {
 	name string
 }
 
+// Returns cache's name.
 func (ci CacheInfo) Name() string {
 	return ci.name
 }
 
 const (
-	HIT  string = "HIT"
+	// Represents cache hit.
+	HIT string = "HIT"
+	// Represents cache miss.
 	MISS string = "MISS"
 )
 
+// Initializes GetCacheResponse to handle gRPC get response.
 type GetCacheResponse struct {
 	value  []byte
 	result string
 }
 
+// Returns value stored in cache as string if there was Hit. Returns an empty string otherwise.
 func (resp *GetCacheResponse) StringValue() string {
 	if resp.result == HIT {
 		return string(resp.value)
@@ -38,6 +47,7 @@ func (resp *GetCacheResponse) StringValue() string {
 	return ""
 }
 
+// Returns value stored in cache as bytes if there was Hit. Returns nil otherwise.
 func (resp *GetCacheResponse) ByteValue() []byte {
 	if resp.result == HIT {
 		return resp.value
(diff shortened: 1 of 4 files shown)


momento/cache_operation_responses.go
```diff
@@ -1,62 +1,76 @@
 package momento
 
+// Initializes ListCacheResponse to handle list cache response.
```
gautamomento, line 3: I don't think we need Initialize here. Customer will never initialize this object. They are going to get this type and we just need to tell them what the type represents. From MomentoSDK developer perspective, initializes makes sense but not for the customer. I would revisit other descriptions too to see if they need `Initializes` anywhere in the description.

Output of the List caches operation

momento/cache_operation_responses.go
```diff
@@ -1,62 +1,76 @@
...
 func (resp *ListCachesResponse) NextToken() string {
 	return resp.nextToken
 }
 
+// Returns all caches.
 func (resp *ListCachesResponse) Caches() []CacheInfo {
 	return resp.caches
 }
 
+// Initializes CacheInfo to handle caches returned from list cache operation.
```
gautamomento, line 19: Information about the Cache.

momento/cache_operation_responses.go
```diff
@@ -1,62 +1,76 @@
 package momento
 
+// Initializes ListCacheResponse to handle list cache response.
 type ListCachesResponse struct {
 	nextToken string
 	caches    []CacheInfo
 }
 
+// Returns next token.
```
gautamomento, line 9: This warrants a more elaborate comment.

This probably warrants a comment.

Next Page Token returned by Simple Cache Service along with the list of caches.

If nextToken is present, then this token must be provided in the next call to continue paginating through the list. This is done by setting this value in ListCachesRequest

momento/error.go
```diff
@@ -19,9 +20,13 @@ const (
 )
 
 type MomentoError interface {
+	// Satisfy the generic error interface.
```
gautamomento, line 23: I would again skip this. From customer perspective, what does this field mean? What information does this convey?

momento/simple_cache_client.go
```diff
@@ -117,6 +134,8 @@ func (c *DefaultScsClient) DeleteCache(request *DeleteCacheRequest) error {
 	return nil
 }
 
+// Lists all caches in your Momento account.
+// The following is a possible error that can be returned:
```
gautamomento, line 138: This description seems incomplete.

momento/simple_cache_client.go
```diff
@@ -175,6 +204,7 @@ func (c *DefaultScsClient) Get(request *CacheGetRequest) (*GetCacheResponse, err
 	}, nil
 }
 
+// Closes both control and data gRPC client connections.
```
gautamomento, line 207: If thus customer facing then just say closes the client. 

They need not know about if we use gRPC or have separate data and control clients.

gautamomento (commented):
Would be nice if @eaddingtonwhite also takes a pass at this. Go comments seem very confusing to me. Seems like they don't differentiate between internal and external comments?

Like in Java and python, there is a differentiation between what qualifies as an internal comment vs what is included in the docs.

momento/error.go
poppoerika: I referenced this comment from aws golang sdk [here](https://github.com/aws/aws-sdk-go/blob/main/aws/awserr/error.go#L32). 
But I'm okay with not adding an extra information about this since `error` is in the `MomentoError` interface to satisfy the generic error interface. I'm not sure if additional information is needed.

eaddingtonwhite (approved):
Let keep iterating on this in future prs.

PR #25, opened 2022-04-21
feat: add support for signing keys APIs

tylerburdsall:
Adds support for `CreateSigningKey`, `RevokeSigningKey`, and `ListSigningKeys` APIs


internal/models/responses.go
```diff
@@ -38,6 +41,54 @@ func NewCacheInfo(cache *pb.XCache) CacheInfo {
...
+type CreateSigningKeyResponse struct {
+	KeyId     string
+	Endpoint  string
+	Key       string
+	ExpiresAt time.Time
+}
+
+func NewCreateSigningKeyResponse(endpoint string, resp *pb.XCreateSigningKeyResponse) *CreateSigningKeyResponse {
+	keyObj := map[string]string{}
+	json.Unmarshal([]byte(resp.GetKey()), &keyObj)
```
eaddingtonwhite, line 57: for sure want to handle error here and have this function return an error as well.

internal/models/responses.go
```diff
@@ -38,6 +41,54 @@ func NewCacheInfo(cache *pb.XCache) CacheInfo {
...
+
+type CreateSigningKeyResponse struct {
+	KeyId     string
+	Endpoint  string
+	Key       string
+	ExpiresAt time.Time
+}
+
+func NewCreateSigningKeyResponse(endpoint string, resp *pb.XCreateSigningKeyResponse) *CreateSigningKeyResponse {
+	keyObj := map[string]string{}
```
eaddingtonwhite, line 56: I think you can define this more cleanly like
```go
var keyObj map[string]string
```

internal/models/responses.go
```diff
@@ -38,6 +41,54 @@ func NewCacheInfo(cache *pb.XCache) CacheInfo {
...
+	NextToken string
+}
+
+type ListSigningKeysResponse struct {
+	NextToken   string
+	SigningKeys []SigningKey
+}
+
+func NewListSigningKeysResponse(endpoint string, resp *pb.XListSigningKeysResponse) *ListSigningKeysResponse {
+	var signingKeys = []SigningKey{}
```
eaddingtonwhite, line 75: same here can just define nil slice like
```go
var signingKeys []SigningKey 
```

should be same

internal/models/responses.go
```diff
@@ -38,6 +41,54 @@ func NewCacheInfo(cache *pb.XCache) CacheInfo {
...
+}
+
+type SigningKey struct {
+	KeyId     string
+	Endpoint  string
+	ExpiresAt time.Time
+}
+
+func NewSigningKey(endpoint string, signingKey *pb.XSigningKey) SigningKey {
+	return SigningKey{KeyId: signingKey.GetKeyId(), Endpoint: endpoint, ExpiresAt: time.Unix(int64(signingKey.GetExpiresAt()), 0)}
```
eaddingtonwhite, line 89: This just me personally but I like expanding stuff like this out I think easier to read. Hard to quickly see fields on object being returned when all on one line. Take as you will dont feel super strongly on this.

```go
return SigningKey{
  KeyId: signingKey.GetKeyId(), 
  Endpoint: endpoint, 
  ExpiresAt: time.Unix(int64(signingKey.GetExpiresAt()), 0)
}
```

internal/utility/utility.go
```diff
@@ -23,3 +23,10 @@ func IsValueValid(value interface{}) momentoerrors.MomentoSvcErr {
 	}
 	return nil
 }
+
+func IsTtlValid(ttl uint64) momentoerrors.MomentoSvcErr {
+	if ttl < 0 {
```
eaddingtonwhite, line 28: like lint error said unsigned int cannot be less then 0 I did think we wanted to try and fail on 0 ttl moving forward though in sdks so maybe can make `ttl < 1`. Erica can confitm.
cc/ @poppoerika

momento/simple_cache_client.go
```diff
@@ -150,6 +154,57 @@ func (c *DefaultScsClient) ListCaches(request *ListCachesRequest) (*ListCachesRe
 	}, nil
 }
 
+// Creates a Momento signing key in your Moemnto account
```
eaddingtonwhite, line 157: Moemnto

momento/simple_cache_client.go
```diff
@@ -150,6 +154,57 @@ func (c *DefaultScsClient) ListCaches(request *ListCachesRequest) (*ListCachesRe
...
+	}
+	return &CreateSigningKeyResponse{
+		keyId:     rsp.KeyId,
+		endpoint:  rsp.Endpoint,
+		key:       rsp.Key,
+		expiresAt: rsp.ExpiresAt,
+	}, nil
+}
+
+// Revokes a Momento signing key in your Moemnto account, all tokens signed by which will be invalid
```
eaddingtonwhite, line 177: Moemnto

momento/simple_cache_client.go
```diff
@@ -150,6 +154,57 @@ func (c *DefaultScsClient) ListCaches(request *ListCachesRequest) (*ListCachesRe
...
+	err := c.controlClient.RevokeSigningKey(&models.RevokeSigningKeyRequest{
+		KeyId: request.KeyId,
+	})
+	if err != nil {
+		return convertMomentoSvcErrorToCustomerError(err)
+	}
+	return nil
+}
+
+// Lists all Momento signing keys in your Moemnto account
```
eaddingtonwhite, line 191: Moemnto

eaddingtonwhite (dismissed):
@tylerburdsall this is awesome!! ❤️ were adding this to go sdk now also. Left some feed back but overall looking good. 

Could we also try running golint and gofmt locally and check out put. It seems like were not totally compliant to me but hard to say for sure. Not sure if we need an extra cicd step maybe need to audit.

internal/utility/utility.go
tylerburdsall: 🤦 thank you!

wandaitzuchen (approved):

PR #30, opened 2022-09-26
chore: add README generator

poppoerika:
https://github.com/momentohq/dev-eco-issue-tracker/issues/109

@cprice404 @eaddingtonwhite @pgautier404 
With the current workflow setup, we need to manually release a new Go package to see if the README generator creates a README properly. 
Is it okay to do so?
Also, this is the same case for the Java SDK. So even though I merged [my pr](https://github.com/momentohq/client-sdk-java/pull/148), we need to manually release this as well to see if any fixes are needed.


.github/workflows/manual_release.yml
```diff
@@ -61,6 +61,15 @@ jobs:
...
           go test -v ./momento
 
+      - name: Generate README
+        uses: momentohq/standards-and-practices/github-actions/generate-and-commit-oss-readme@gh-actions-v1
+        with:
+          project_status: official
+          project_stability: experimental
+          project_type: sdk
+          sdk_language: Go
+          usage_example_path: ./examples/main.go
```
cprice404, line 71: @poppoerika can you create a new workflow called `on-push-to-main` or similar (let's just choose a naming convention from one of the repos where we've already done this like javascript or python or dotnet) and put the README generation stanza into that?

cprice404 (commented):
lgtm except I think we probably want to have the readme generation happen on push to main rather than only on releases.

eaddingtonwhite (approved):
Looks great!

PR #34, opened 2022-12-06
chore: adds code scanning workflow

eaddingtonwhite:


eaddingtonwhite:
Looks good.
Screenshot 2022-12-05 at 5 11 01 PM

bruuuuuuuce:
could this potentially go in the shared workflow https://github.com/momentohq/client-sdk-go/blob/main/.github/workflows/build.yml#L23?

eaddingtonwhite:
@bruuuuuuuce potentially... it can be bit different depending on repo is only thing (see java sdk for example). We can revisit though down line I wanted to get feel for this check first and break out to start.

bruuuuuuuce (approved):
:shipit:

PR #42, opened 2023-01-19
feat: break out incubating package and adds basic pubsub support

eaddingtonwhite:


internal/grpcmanagers/data_manager.go
```diff
@@ -3,30 +3,62 @@ package grpcmanagers
...
+	"google.golang.org/grpc/credentials/insecure"
 )
 
-type ScsDataGrpcManager struct {
+type DataGrpcManager struct {
 	Conn *grpc.ClientConn
 }
 
-func NewScsDataGrpcManager(request *models.DataGrpcManagerRequest) (*ScsDataGrpcManager, momentoerrors.MomentoSvcErr) {
+func NewUnaryDataGrpcManager(request *models.DataGrpcManagerRequest) (*DataGrpcManager, momentoerrors.MomentoSvcErr) {
```
poppoerika, line 19: Just a question.
We have both `NewUnaryDataGrpcManager` and `NewStreamDataGrpcManager` so that we can set different config and grpc dial options if we want to?

poppoerika: Ah I see that they use different interceptors.

pgautier404 (approved):

internal/grpcmanagers/data_manager.go
eaddingtonwhite: yeah correct... I probably can just make single function with this and just have caller pass down `...DialOption` might be cleaner here but just wanted to be bit more explicit about different "managers" we support and keep thin in clients provisioning was my thought process breaking out like this. 

The differences currently:
`NewLocalDataGrpcManager` - No TLS, No Auth, 
`NewUnaryDataGrpcManager` - unary authheader interceptor
`NewStreamDataGrpcManager` - stream authheader interceptor

internal/services/scs_control_service.go
```diff
@@ -48,6 +48,32 @@ func (client *ScsControlClient) CreateCache(request *models.CreateCacheRequest)
 	return nil
 }
 
+func (client *ScsControlClient) CreateTopic(ctx context.Context, request *models.CreateTopicRequest) momentoerrors.MomentoSvcErr {
+	if !utility.IsCacheNameValid(request.TopicName) {
+		return momentoerrors.NewMomentoSvcErr(momentoerrors.InvalidArgumentError, "Topic name cannot be empty", nil)
+	}
+	ctx, cancel := context.WithTimeout(ctx, ControlCtxTimeout)
+	defer cancel()
+	_, err := client.grpcClient.CreateCache(ctx, &pb.XCreateCacheRequest{CacheName: "topic-" + request.TopicName})
```
poppoerika, line 57: Another question: `CreateTopic` = `CreateCache` 
So a topic is a cache in Momento?

poppoerika (approved):

internal/services/scs_control_service.go
eaddingtonwhite: yes. will DM u can chat bit further some of this is in flight still might change a bit.

PR #54, opened 2023-01-20
chore: add `context.Context` as a parameter to public APIs

### LINKED ISSUE #47: fix how context.Context is passed from SDK calls be consistent
All our SDK signatures should take a parameters like: 
`CreateTopic(ctx context.Context, request *CreateTopicRequest)`

For several functions were not allowing user to pass down a context and injecting our own context under the hood. 
[Example](https://github.com/momentohq/client-sdk-go/blob/main/internal/services/scs_control_service.go#L42)

This is not good practice we should update to consistently allow request context to be passed down to each of our exposed methods as first parameter.

poppoerika:
Following what the pubsub public APIs do, allow users to pass their own `context.Context` to the public cache APIs rather than injecting our own context under the hood.

Closes #47

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,6 +43,7 @@ jobs:
         with:
           version: latest
           only-new-issues: true
+          skip-pkg-cache: true
 
       # If there are any diffs from goimports or go mod tidy, fail.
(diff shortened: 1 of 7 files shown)


eaddingtonwhite (approved):
@poppoerika Looks good thanks for jumping on this.

PR #57, opened 2023-01-24
feat: makes incubating package inherit cache functionality

eaddingtonwhite:


incubating/simple_cache_client.go
```diff
@@ -0,0 +1,149 @@
...
+// TODO figure out better way to dry this up is copy pasta from simple cache client
+func convertMomentoSvcErrorToCustomerError(e momentoerrors.MomentoSvcErr) momento.MomentoError {
+	if e == nil {
+		return nil
+	}
+	return momento.NewMomentoError(e.Code(), e.Message(), e.OriginalErr())
+}
+
+func (c *DefaultScsClient) CreateCache(ctx context.Context, request *momento.CreateCacheRequest) error {
+	return c.internalClient.CreateCache(ctx, request)
```
cprice404, line 116: this is how I would do this if golang doesn't support inheritance or delegation/decorator pattern via a built-in language construct, so this seems fine.  but... does golang not support inheritance?

cprice404 (approved):
one question but lgtm

.github/workflows/build.yml
```diff
@@ -7,7 +7,6 @@ permissions:
   pull-requests: read
 jobs:
   test:
-    strategy:
```
poppoerika, line 10: Thank you!

incubating/simple_cache_client.go
eaddingtonwhite: We could embed the `momento.DefaultScsClient` struct directly on `incubating.DefaultScsClient` potentially to make this bit cleaner and avoid pass through function calls like this.
```golang
type DefaultScsClient struct {
	*momento.DefaultScsClient
	controlClient *services.ScsControlClient
	pubSubClient  *services.PubSubClient
}
```

I think will require us to change up how we initialize the regular client a bit more and what we expose to other packages/users. Its internal detail to us at the moment. I kind of want to do config refactor since might simplify how we construct clients and then can come back and revisit if easier way to make it inherit functionality from other client in incubating SDK.

PR #58, opened 2023-01-24
feat: adds standard momento client configuration

eaddingtonwhite:


config/config.go
```diff
@@ -0,0 +1,51 @@
...
+type Configuration interface {
+	// GetTransportStrategy Returns the current configuration options for wire interactions with the Momento service
+	GetTransportStrategy() TransportStrategy
+
+	// WithTransportStrategy Copy constructor for overriding TransportStrategy returns a new Configuration object
+	//with the specified momento.TransportStrategy
+	WithTransportStrategy(transportStrategy TransportStrategy) Configuration
+
+	// GetClientSideTimeoutMillis Returns the current configuration options for client side timeout with the Momento service
+	GetClientSideTimeoutMillis() uint32
```
pgautier404, line 16: Just FYI, we're trying to standardize on using timespan/duration constructs instead of ints for deadlines and TTLs. I don't want to hold this PR up for that, especially since we'll need to also correct TTLs all over the place, but did want you to be aware.

eaddingtonwhite: thats good call Let me swap out quick

eaddingtonwhite: @pgautier404 I think I got them all. Let me know if missed anything though! Great catch! 🦅 👁️ thanks!

incubating/simple_cache_client_test.go
```diff
@@ -78,32 +82,32 @@ func TestBasicHappyPathPubSubIntegrationTest(t *testing.T) {
 	}
 
 	sub, err := client.SubscribeTopic(ctx, &TopicSubscribeRequest{
-		CacheName: "test-cache",
+		CacheName: "default",
```
pgautier404, line 85: We're creating a cache named `test-cache` just above this. I assume you want to use the same here? This test failed for me locally with a `NotFound` error because I didn't have a cache named `default`.

incubating/simple_cache_client_test.go
```diff
@@ -78,32 +82,32 @@ func TestBasicHappyPathPubSubIntegrationTest(t *testing.T) {
...
-	//		Value:     fmt.Sprintf("hello %d", i),
-	//	})
-	//	if err != nil {
-	//		panic(err)
-	//	}
-	//	time.Sleep(time.Second)
-	//}
+	for i := 0; i < 10; i++ {
+		err := client.PublishTopic(ctx, &TopicPublishRequest{
+			CacheName: "default",
```
pgautier404, line 104: Same as above regarding cache name.

incubating/simple_cache_client_test.go
eaddingtonwhite: oops yeah good catch

pgautier404 (approved):
:shipit:

PR #61, opened 2023-01-26
chore: add a string token provider for use with canary

pgautier404:
This adds a `NewStringMomentoTokenProvider` method to construct a credential provider from an auth token.

### DIFF
--- a/auth/credential_provider.go
+++ b/auth/credential_provider.go
@@ -55,6 +55,12 @@ func NewEnvMomentoTokenProvider(envVariableName string) (CredentialProvider, err
 			errors.New("invalid argument"),
 		)
 	}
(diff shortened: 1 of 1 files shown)


poppoerika (approved):
Looks good to me!

One question: I know this new string token provider is for the canary and I'm not familiar with the canary setup, but if we cannot use the existing env var token provider, do we want to expose this new string token provider?
I don't think we have it in either the .NET/PHP SDK?

pgautier404:
> Looks good to me!
> 
> One question: I know this new string token provider is for the canary and I'm not familiar with the canary setup, but if we cannot use the existing env var token provider, do we want to expose this new string token provider? I don't think we have it in either the .NET/PHP SDK?

It's being added to other SDKs as well as needed.

PR #63, opened 2023-01-27
feat: get sdk response types up to standards

eaddingtonwhite:
* removes signing key impl to break into its own package

### DIFF
--- a/internal/models/requests.go
+++ b/internal/models/requests.go
@@ -18,8 +18,10 @@ type LocalDataGrpcManagerRequest struct {
 	Endpoint string
 }
 type CacheGetRequest struct {
+	// CacheName Name of the cache to get the item from
 	CacheName string
-	Key       interface{}
+	// Key string ot []byte key to be used to retrieve the item.
+	Key interface{}
 }
 
 type ControlClientRequest struct {
--- a/internal/models/requests.go
+++ b/internal/models/requests.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/momentohq/client-sdk-go/auth"
+	"github.com/momentohq/client-sdk-go/config"
 	"github.com/momentohq/client-sdk-go/internal/momentoerrors"
 	pb "github.com/momentohq/client-sdk-go/internal/protos"
 )
@@ -21,18 +22,19 @@ type LocalDataGrpcManagerRequest struct {
 }
 
 type ControlClientRequest struct {
+	Configuration      config.Configuration
 	CredentialProvider auth.CredentialProvider
 }
 
 type DataClientRequest struct {
-	CredentialProvider    auth.CredentialProvider
-	DefaultTtlSeconds     uint32
-	RequestTimeoutSeconds uint32
+	Configuration      config.Configuration
+	CredentialProvider auth.CredentialProvider
+	DefaultTtlSeconds  uint32
 }
 
 type PubSubClientRequest struct {
+	Configuration      config.Configuration
 	CredentialProvider auth.CredentialProvider
-	// TODO think about timeout settings more
 }
 
 type NewLocalPubSubClientRequest struct {
--- a/internal/models/requests.go
+++ b/internal/models/requests.go
@@ -1,11 +1,8 @@
 package models
 
 import (
-	"fmt"
-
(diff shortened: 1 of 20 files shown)


momento/cache_responses.go
```diff
@@ -1,148 +1,85 @@
...
 )
 
-// Initializes GetCacheResponse to handle gRPC get response.
-type GetCacheResponse struct {
-	value  []byte
-	result string
+// CacheGetResponse Base type for possible responses a cache GET can return. Miss || Hit
+type CacheGetResponse struct {
+	responseType cacheGetResponseTypes
+	value        []byte
```
cprice404, line 41: so these are not accessible to users because they are lowercase, correct?

momento/cache_responses.go
```diff
@@ -1,148 +1,85 @@
...
-func (resp *SetCacheResponse) StringValue() string {
+// StringValue Returns value stored in cache as string if there was Hit. Returns an empty string otherwise.
+func (resp *CacheGetHitResponse) StringValue() string {
 	return string(resp.value)
 }
 
-// Returns byte value set in cache.
-func (resp *SetCacheResponse) ByteValue() []byte {
+// ByteValue Returns value stored in cache as bytes if there was Hit. Returns nil otherwise.
+func (resp *CacheGetHitResponse) ByteValue() []byte {
```
cprice404, line 83: We've been preferring to prefix all value accessors with the word `Value`, so that they are logically grouped in the IDE.  So I think you want to rename these to `ValueString` and `ValueBytes`.

cprice404 (commented):
couple nits and I think maybe get @pgautier404 to review, otherwise lgtm

tylerburdsall (dismissed):
LGTM, but should probably get another set of eyes on this

momento/cache_responses.go
eaddingtonwhite: correct :) golang!

momento/cache_responses.go
eaddingtonwhite: oo good catch nice will do

internal/models/requests.go
```diff
@@ -20,6 +17,12 @@ type DataGrpcManagerRequest struct {
 type LocalDataGrpcManagerRequest struct {
 	Endpoint string
 }
+type CacheGetRequest struct {
+	// CacheName Name of the cache to get the item from
+	CacheName string
+	// Key string ot []byte key to be used to retrieve the item.
```
pgautier404, line 23:
```suggestion
	// Key string or []byte key to be used to retrieve the item.
```

internal/services/scs_data_service.go
```diff
@@ -36,7 +35,7 @@ func NewScsDataClient(request *models.DataClientRequest) (*ScsDataClient, moment
 	if request.Configuration.GetClientSideTimeoutMillis() < 1 {
 		timeout = time.Duration(defaultRequestTimeoutSeconds) * time.Second
 	} else {
-		timeout = time.Duration(request.Configuration.GetClientSideTimeoutMillis()) * time.Second
+		timeout = request.Configuration.GetClientSideTimeoutMillis() * time.Second
```
pgautier404, line 38: This doesn't seem right. The client side timeout should already be a `time.Duration` and shouldn't need this conversion I think. It might be best to remove the `Millis` from the name of this property completely to avoid confusion.

eaddingtonwhite: oo good catch. Let me fix this I think just missed here when cutting over to duration.

momento/simple_cache_client.go
```diff
@@ -243,18 +146,31 @@ func (c *DefaultScsClient) Delete(ctx context.Context, request *CacheDeleteReque
...
 	defer c.dataClient.Close()
 }
 
+func convertCacheGetResponse(r *models.CacheGetResponse) *CacheGetResponse {
+	var response *CacheGetResponse
+	switch r.Result {
+	case models.MISS:
+		response = &CacheGetResponse{
+			responseType: miss,
+			value:        r.Value,
```
pgautier404, line 163: `value` should be omitted for a miss.

eaddingtonwhite: good catch!

pgautier404 (approved):
Looks great, thanks!

PR #67, opened 2023-01-29
feat: add simple logger

### LINKED ISSUE #51: Get SDK Logging Up to standards
Should have standard Momento SDK Info and Debug logging configurable

poppoerika:
Added a logger using `log` go built-in package. 
For now, users can configure a message key to display along with a message. 
Users can choose a log level to log by calling either `INFO` or `DEBUG`.
Output example from running `examples/main.go` looks like this:
```
{"level": "INFO", "message": "Creating cache"}
{"level": "INFO", "message": "default"}
{"level": "INFO", "message": "erika-cache"}
{"level": "INFO", "message": "cache"}
{"level": "INFO", "message": "Setting key: 28f8d707-171f-4792-ad09-f7a9855dfca3, value: 99da0344-79f4-4794-90a0-3df1b29a1653"}
{"level": "INFO", "message": "Getting key: 28f8d707-171f-4792-ad09-f7a9855dfca3"}
{"level": "INFO", "message": "Lookup resulted in cahce HIT. value=99da0344-79f4-4794-90a0-3df1b29a1653"}
{"level": "INFO", "message": "Cache named cache is deleted"}
```

I referenced this logger package called [zap](https://pkg.go.dev/go.uber.org/zap) which displays logs in a JSON format.
If more detailed logger is requested/preferred, we could use this package under the hood?

Closes #51


config/configurations.go
```diff
@@ -32,6 +36,9 @@ type InRegion struct {
 func LatestInRegionConfig() *InRegion {
```
cprice404, line 36: this isn't related to your PR but do we not want to have InRegion.Default and InRegion.LowLatency like we do in the other languages?  cc: @eaddingtonwhite

config/configurations.go
```diff
@@ -10,10 +10,14 @@ const defaultMaxSessionMemoryMb = 256
 
 // 4 minutes.  We want to remain comfortably underneath the idle timeout for AWS NLB, which is 350s.
 const defaultMaxIdle = 4 * time.Minute
+const defaultLoggerName = "message"
 
 func LatestLaptopConfig() *Laptop {
```
cprice404, line 15: Logging configuration is something that the user is almost always going to want to customize, so the pattern I have been following is to allow the logging configuration to be passed in as an argument to the factory function for the pre-built configs, so that they won't have to worry about finding copy constructors afterward:

https://github.com/momentohq/client-sdk-dotnet/blob/main/src/Momento.Sdk/Config/Configurations.cs#L68

config/configurations.go
```diff
@@ -10,10 +10,14 @@ const defaultMaxSessionMemoryMb = 256
 
 // 4 minutes.  We want to remain comfortably underneath the idle timeout for AWS NLB, which is 350s.
 const defaultMaxIdle = 4 * time.Minute
+const defaultLoggerName = "message"
```
cprice404, line 13: Is this actually a logger name?  Or is this controlling the field name for the message that gets produced?

If the latter, does go's built-in logging library have the concept of a logger name?

config/config.go
```diff
@@ -1,12 +1,22 @@
...
 	TransportStrategy TransportStrategy
 }
 
 type Configuration interface {
+	// GetLogger Returns the currently configured logger with the Momento service
+	GetLogger() Logger
+
+	// WithCustomLogger Copy constructor for overriding Logger returns a new Configuration object
+	// with the specified logger name.
+	WithCustomLogger(logger Logger) Configuration
```
cprice404, line 18: In this API I believe you are specifying a concrete type, `Logger`, from the built-in golang log library.  We need to think about how this would work if a user is already using `zap` or `logrus` or something else.

With logging it's really important that we give them a way to configure us to use their existing logging setup if that's what they want to do.  I don't think these signatures would support that today but I could be wrong.  cc: @eaddingtonwhite 

I think what we need to do is provide an Interface here for a MomentoLogger and a MomentoLoggerFactory, and accept that MomentoLoggerFactory as part of the configuration.  For now our MomentoLogger interface might only have a handful of basic functions defined on it: Info, Debug, etc.

The MomentoLoggerFactory would have one fn, `createLogger(loggerName String)`.  

Then we can implement concrete instances of those interface using the built-in `log.Logger`, but if someone is already using Zap they can define concrete instances that are backed by Zap (or we can) and then all of our logs will go to their normal logging destinations.

If this doesn't make sense we can circle up.  And if we don't have time for that right now that's okay but we shouldn't add logging configuration to the Configuration interface until we do, IMO.

cprice404 (commented):
need to consider how this will work for users who already have a preferred logging library.

config/config.go
poppoerika: @cprice404 Thank you for the through feedback. It makes sense to me and it cleared up what I was missing in adding logging configuration.
So this is my understanding of what we want to implement after reading your feedback:
- Create `MomentoLoggerFactory` that has one function called `createLogger` to create a logger factory
- Implement interface that would allow users to use their own logger if they already have one using. If not, use a simple `log.Logger` to log `INFO` and `DEBUG`

@eaddingtonwhite @cprice404 
My question here is that how can we make a common interface like C#'s `ILoggerFactory`?
My understanding is that Go doesn't have a common interface for logging so if we are to allow users to use their preferred/existing logger, do we need to implement multiple interfaces to cover those logging packages (`zap`, `logrus`, etc)?
Sorry if I misunderstood anything.

eaddingtonwhite: We should be able to define one common interface and then extend it with various different drivers eventually when someone needs that. Check out how we did in JavaScript SDK to see example of our simple logging interface and default impl using `pino`. 
https://github.com/momentohq/client-sdk-javascript/blob/main/src/utils/logging.ts#L8-L14

cprice404:
For @poppoerika or whomever else ends up picking this up; here is a PR against the node.js SDK that does the same refactor I'm suggesting here:

https://github.com/momentohq/client-sdk-javascript/pull/224

config/logger/builtin-logger.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func (l *BuitlinMomentoLogger) Info(message string, args ...any) {
+	str := ""
+	if args == nil {
+		str = fmt.Sprintf(`{"level": "INFO", "message": "%s", "name": "%s"}`, message, l.loggerName)
+	} else {
+		str = fmt.Sprintf(`{"level": "INFO", "message": "%s", "name": "%s", "%v"}`, message, l.loggerName, args)
+	}
+	log.SetFlags(0)
```
cprice404, line 19: Do we need to set flags on every log message?  I would assume that would be something we want to do in the constructor?

config/logger/builtin-logger.go
```diff
@@ -0,0 +1,43 @@
...
+	"log"
+)
+
+type BuitlinMomentoLogger struct {
+	loggerName string
+}
+
+func (l *BuitlinMomentoLogger) Info(message string, args ...any) {
+	str := ""
+	if args == nil {
```
cprice404, line 14: I'm not a go expert so I will defer to ellery as to whether there's any way to "spread" the args list so we don't need an if/else here, but it's not a blocker.

examples/main.go
```diff
@@ -71,14 +73,14 @@ func main() {
 
 	switch r := resp.(type) {
 	case *momento.CacheGetHit:
-		log.Printf("Lookup resulted in cahce HIT. value=%s\n", r.ValueString())
+		logger.Info(fmt.Sprintf("Lookup resulted in cahce HIT. value=%s\n", r.ValueString()))
 	case *momento.CacheGetMiss:
-		log.Printf("Look up did not find a value key=%s", key)
+		logger.Info(fmt.Sprintf("Look up did not find a value key=%s", key))
```
cprice404, line 78: Shouldn't you be able to avoid the `Sprintf` here since you are doing it inside of the logging impl?

cprice404 (commented):
this is very close now, thanks!  just a few more questions/comments

examples/main.go
poppoerika: `log.Printf` 😮‍💨

config/configurations.go
```diff
@@ -14,6 +17,7 @@ const defaultMaxIdle = 4 * time.Minute
 func LatestLaptopConfig() *Laptop {
 	return &Laptop{
 		Configuration: NewSimpleCacheConfiguration(&ConfigurationProps{
+			LoggerFactory: logger.NewBuitlinMomentoLoggerFactory(),
```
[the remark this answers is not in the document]
poppoerika: My understanding is that Go doesn't have the concept of optional parameters. 
We could make `NewSimpleCacheConfiguration` to take `any` and do some type matching once we get actual values and assign the "default" logger Factory then. But I think it could get messy and unclear to what kind of config props we support? cc @eaddingtonwhite

config/logger/builtin-logger.go
poppoerika: I can remove the if-else but the reason why I'm checking `if args == nil` us that if users don't pass anything to `args` then it's logged as
```
{"level": "INFO", "message": "Setting key: 43635825-03b2-43d8-8103-5355d6a83a65, value: 2e937128-d09b-4f84-8abf-395e6b1aaed9
", "name": "examples-main", "[]"}
```

Since I'm using `%v` to represent `args`, it's displayed as `[]` at the end I thought it's a bit confusing to include an empty list at the end.

cprice404: yes, most languages that accept variadic arguments have some kind of operator you can use to "spread" them so that you could prepend the message and logger name to the array and then spread it back out to pass to log.Printf.  (e.g. `...` operator in typescript).  I would have guessed that go would have something like that, but maybe it doesn't.  If it doesn't then the conditional is fine for now.

config/configurations.go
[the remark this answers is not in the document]
cprice404: okay we don't have to solve this on this PR but we do need to have a way for callers to pass in a loggerfactory.

cprice404 (dismissed):

config/configurations.go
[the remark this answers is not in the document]
poppoerika: Ah sorry I misunderstood what you mentioned above. I will update it 🙇‍♀️

poppoerika:
@cprice404 @eaddingtonwhite 
For [this](https://github.com/momentohq/client-sdk-go/pull/67#discussion_r1097930451), I made `LatestLaptopConfig` accepts `...logger.MomentoLoggerFactory` in that way users can pass zero or more logger factory.
But it only accepts one logger factory to be passed to a configuration constructor.
Since users could technically pass more than one logger factory, I though it's a bit counterintuitive but this is I think a clear/type safe way to achieve an "optional" parameter.
I'm no Go expert so any feedback would be appreciated!

config/configurations.go
```diff
@@ -29,9 +37,14 @@ type InRegion struct {
 	Configuration
 }
 
-func LatestInRegionConfig() *InRegion {
+func LatestInRegionConfig(loggerFactory ...logger.MomentoLoggerFactory) *InRegion {
```
eaddingtonwhite, line 40: why don't we always just default to Noop logger here and then for user to overide they have to use `WithLoggerFactory` on returned config object? Isnt this how you are using in your example? Seems a bit strange to me to just allow user to overide logging options when calling `LatestInRegionConfig()`? Do we plan on allowing them to pass other over rides like for transport strategy as well?

eaddingtonwhite: ahh ok see this message again now. https://github.com/momentohq/client-sdk-go/pull/67#discussion_r1090880809
I think lets just go w/ this way for now see how it goes as want to extend more. Can always refactor a bit. Maybe can you just update example to use this method and pass in logging config directly to `LatestInRegionConfig()` vs using the `WithLoggerFactory()` so that users do this way since is reason we added this.

poppoerika: Yup! Sounds good. I will update the example.

cprice404: @poppoerika I'm sorry for the thrash on this but can you just change it to require a single instance of MomentoLoggerFactory and remove the `WithLoggerFactory` for now?  There is a chicken and egg problem I keep forgetting about with `WithLoggerFactory`, I can explain it to you all on a zoom tomorrow if you like.

sorry this pr has taken so long to get in, I promise I am +1 if you just change it to single required loggerFactory instance for now :)

config/logger/builtin-logger.go
eaddingtonwhite: Are we trying to log key values pairs w/ json or just append on to basic string message and do non json logging as default for each additional passed in arg? Feels bit strange to me to just log as an additional array with every other field defaulting to json format. We might want to use a logging lib to help us deal with converting passed args to json if we want users to pass keyvalue pairs? Maybe quick zoom on this might be easier.

config/configurations.go
poppoerika: 👍  Yeah, I can remove that.

cprice404 (approved):
i'm okay with merging this as-is for now if @eaddingtonwhite is.

next step IMO is to get rid of `WithLoggerFactory` and change the signatures to only accept one LoggerFactory instead of multiple.  Then from there we may want to also tweak the interface based on my convo with Ellery this morning.  But I'm okay with those being handled in later PRs.

poppoerika:
Output with `BuiltinMomentoLogger` in `examples/main.go`
```
[2006-01-02T15:04:05Z07:00] INFO (examples-main): Setting key/value, key, 4f3cc609-9bb6-4ca1-ad84-62828a1cf9a1, value, cbee10a0-d3bc-41d1-b454-c6536929f937
[2006-01-02T15:04:05Z07:00] INFO (examples-main): Getting key, key, 4f3cc609-9bb6-4ca1-ad84-62828a1cf9a1
[2006-01-02T15:04:05Z07:00] INFO (examples-main): Lookup resulted in cahce HIT, key, 4f3cc609-9bb6-4ca1-ad84-62828a1cf9a1, value, cbee10a0-d3bc-41d1-b454-c6536929f937
[2006-01-02T15:04:05Z07:00] INFO (examples-main): Cache named my-test-cache is deleted
```

cc @eaddingtonwhite @cprice404

examples/main.go
```diff
@@ -71,14 +72,14 @@ func main() {
 
 	switch r := resp.(type) {
 	case *momento.CacheGetHit:
-		log.Printf("Lookup resulted in cahce HIT. value=%s\n", r.ValueString())
+		mLogger.Info("Lookup resulted in cahce HIT", []string{"key", key, "value", r.ValueString()}...)
```
eaddingtonwhite, line 75: Typo in `cahce` also shouldn't we be able to just pass additional args to log like this now?
```suggestion
		mLogger.Info(
		   "Lookup resulted in cache HIT", "key", key, "value", r.ValueString(),
	         )
```

poppoerika: Thank you for catching it!
Dah! I will update them.

examples/main.go
```diff
@@ -71,14 +72,14 @@ func main() {
...
-		log.Printf("Look up did not find a value key=%s", key)
+		mLogger.Info("Look up did not find a value", "key", key)
 	}
 
 	// Permanently delete the cache
 	if err = client.DeleteCache(ctx, &momento.DeleteCacheRequest{CacheName: cacheName}); err != nil {
 		panic(err)
 	}
-	log.Printf("Cache named %s is deleted\n", cacheName)
+	mLogger.Info(fmt.Sprintf("Cache named %s is deleted\n", cacheName))
```
eaddingtonwhite, line 84: nit we can change in next pr if we want but would normally write log message like this:
```suggestion
	mLogger.Info("Cache successfully deleted", "name", cacheName)
```

eaddingtonwhite (dismissed):
Ok yeah im good on this for now as is we can circle back and iterate in future prs. Thanks for pushing through on this @poppoerika .

eaddingtonwhite (approved):

PR #72, opened 2023-01-31
feat: rework response types to use interface and allow user to pass []byte or string messages to a topic 

eaddingtonwhite:


incubating/cache_response.go
```diff
@@ -1,13 +1,57 @@
...
+	byteType   topicResponseTypes = "bytes"
+)
+
+// TopicMessage Base type for possible messages received on a topic
+type TopicMessage struct {
+	responseType topicResponseTypes
+	value        []byte
+}
+
+type TopicMessageByte struct {
```
kvcache, line 16:
```suggestion
type TopicMessageBytes struct {
```

incubating/cache_response.go
```diff
@@ -1,13 +1,57 @@
...
+	value []byte
+}
+
+func (t *TopicMessageByte) Value() []byte {
+	return t.value
+}
+
+type TopicMessageStringResponse struct {
```
kvcache, line 24:
```suggestion
type TopicMessageString struct {
```

incubating/cache_response.go
```diff
@@ -1,13 +1,57 @@
...
-func (resp *TopicMessageReceiveResponse) StringValue() string {
-	return resp.value
+func (t *TopicMessageStringResponse) Value() string {
+	return t.value
+}
+
+// IsByteMessage returns true if the message received on a topic was of type bytes
+func (r *TopicMessage) IsByteMessage() bool {
+	return r.responseType == byteType
+}
```
kvcache, line 35: I can't shake that these string equality checks still feel wrong. Have you considered following grpc's golang pattern instead of reinventing dynamic dispatch via string-matching?

https://software-factotum.medium.com/protobuf-and-go-handling-oneof-field-type-172ca780ec47

```
switch response := response.Body.(type) {
case *typedefs.BytesBody:
	fmt.Printf("Bytes body: %s\n", response.BytesBody)
case *typedefs.StringBody:
	fmt.Printf("String body: %s\n", response.StringBody)
default:
	fmt.Println("No matching body type in this sdk")
}
```

If you read google's documentation, here is how they have you use variants: https://developers.google.com/protocol-buffers/docs/reference/go-generated
```
switch x := m.Avatar.(type) {
case *account.Profile_ImageUrl:
        // Load profile image based on URL
        // using x.ImageUrl
case *account.Profile_ImageData:
        // Load profile image based on bytes
        // using x.ImageData
case nil:
        // The field is not set.
default:
        return fmt.Errorf("Profile.Avatar has unexpected type %T", x)
}
```

eaddingtonwhite: @kvcache I tink your right that should just use reflection here instead of string match type especially since response type is just driven off if value is string or []byte. Let me update quick especially since this is internal to response types users wont see either. I had used string match for HIT/MISS response type on cache GET and just kept pattern going into this response type.

kvcache: It's probably worth changing that other string match to this strategy too.

Users of the subscribe function should probably get this as well, I don't think I'm suggesting that this should only be applied internally.

eaddingtonwhite: So internally to SDK I think I agree can just use reflection to do matching within response vs enum. Here for sure will go back look at how did it with hit/miss.

So for external hit/miss we were thinking that returning an interface and allowing user to match on it vs helper methods ended up being not as clear for end user vs defined helper methods in response object.
**Interface:**
```go

rsp, _ := client.Get(ctx, request)

// simple cast to hit
if r, ok := rsp.(momento.CacheGetHitResponse); ok {
	val := r.Value()
	fmt.Println(val)
}

// Pattern match on interface returned
switch rsp.(type) {
	case momento.CacheGetHitResponse:
		if r, ok := rsp.(momento.CacheGetHitResponse); ok {
			val := r.Value()
			fmt.Println(val)
		}
}
```
vs helper methods
```go
resp, _ := client.Get(ctx, request)

if resp.IsHit() {
	log.Printf("Lookup resulted in cahce HIT. value=%s\n", resp.AsHit().ValueString())
} else {
	log.Printf("Look up did not find a value key=%s", key)
}
```
If disagree w/ hit/miss approach let me sync w/ u off this can revisit some more. 

For topic recieve though are you proposing we just return single response object like `TopicMesage` with generic `interface{}` since those are only posible values and just let user Handle? 

Just export `value interface{}`
```go
_ = sub.Recv(context.Background(), func(ctx context.Context, m *TopicMessage) {
	if v, ok := m.Value.(string); ok {
		fmt.Printf("got a string! val=%s\n", m.Value)
	}
})

```
vs helper methods
```go
_ = sub.Recv(context.Background(), func(ctx context.Context, m *TopicMessage) {
	if m.IsStringMessage() {
		fmt.Printf("got a msg! val=%s\n", m.AsStringMessage().Value())
	}
}
```

kvcache: No I am proposing that we do like protocol buffers does:
```
switch x := m.Avatar.(type) {
case *account.Profile_ImageUrl:
        // Load profile image based on URL
        // using x.ImageUrl
case *account.Profile_ImageData:
        // Load profile image based on bytes
        // using x.ImageData
case nil:
        // The field is not set.
default:
        return fmt.Errorf("Profile.Avatar has unexpected type %T", x)
}
```

here i reproduced google's golang protocol buffers example using words that are conveniently in our vernacular:
```
// You can edit this code!
// Click here and start typing.
package main

import "fmt"

type GetResponse interface {
	isGetResponse()
}

type Hit struct {
	HitResult string
}

func (_ Hit) isGetResponse() {}

type Miss struct {
	Apology string
}

func (_ Miss) isGetResponse() {}

func get() GetResponse {
	return Hit{HitResult: "it hit"}
}

func main() {
	response := get()
	switch value := response.(type) {
	case Hit:
		fmt.Printf("look it was a typesafe hit: %s\n", value.HitResult)
	case Miss:
		fmt.Printf("look it was a typesafe miss: %s\n", value.Apology)
	default:
		fmt.Printf("I don't know what it was %s", value)
	}
}
```

In case it's unclear, I also think this pattern should be followed for `type TopicValue interface` and `TopicValueBytes` as well as `TopicValueString`. No internal string matching required, no manual typecasting required, no possibility of someone using the wrong type... This is how this is supposed to be done.

Consider:
```
func main() {
	response := get()
	switch value := response.(type) {
	case Hit:
		fmt.Printf("look it was a typesafe hit: %s\n", value.HitResult)
	case Miss:
		fmt.Printf("look it was a typesafe miss: %s\n", value.Apology)
	case string:
		fmt.Println("wut")
	default:
		fmt.Printf("I don't know what it was %s", value)
	}
}
```
which results in a compiler error:
```
./prog.go:34:7: impossible type switch case: string
	response (variable of type GetResponse) cannot have dynamic type string (missing isGetResponse method)

Go build failed.
```
which seems... very very good, like silver-bullet good.

I'm no go expert (I can't even recite hello world), I'm just reading documentation from google and typing what they say into https://go.dev/play/ and it seems like it's better than our homegrown ideas. If this proposal is unclear or if you disagree feel free to reach out on slack.

eaddingtonwhite: ahh ok I see now. I think I see what was missing now when was testing this before. setting `value` with variable initializer in switch makes this way cleaner does casting for you then.  Ok im alligned I agree w/ you will switch it over to use interface w/ types update examples. Good call out.

momento/simple_cache_client.go
```diff
@@ -153,18 +153,14 @@ func (c *DefaultScsClient) Close() {
...
+	var response CacheGetResponse
+	switch r := r.(type) {
+	case *models.CacheGetMiss:
+		response = &CacheGetMiss{}
+	case *models.CacheGetHit:
+		response = &CacheGetHit{
+			value: r.Value,
 		}
 	}
 	return response
```
kvcache, line 166:
```suggestion
func convertCacheGetResponse(model models.CacheGetResponse) CacheGetResponse {
	switch response := model.(type) {
	case *models.CacheGetMiss:
		return CacheGetMiss{}
	case *models.CacheGetHit:
		return CacheGetHit{
			value: response.Value,
		}
	default:
		// log.error?
		// return nil?
		// throw?
		// return miss?
	}
```

kvcache (commented):
Noticed a few things. I am still not a go person though, so definitely tap someone else for a more detailed review of best practices and design patterns!

incubating/cache_response.go
```diff
@@ -1,13 +1,18 @@
...
+// TopicMessage Base type for possible messages received on a topic
+type TopicMessage interface {
+	isTopicMsg()
+}
 
-type TopicMessageReceiveResponse struct {
-	// TODO talk about user experience for bytes/strings
-	value string
+type TopicMessageBytes struct {
+	Value []byte
```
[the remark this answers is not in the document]
pgautier404: It would be more consistent with the data operation responses if we exposed `ValueString()` and `ValueBytes()` methods instead of defining type-specific structs here.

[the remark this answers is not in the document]
kvcache: So a user would have to unwrap both a topic message's value type and the typed value? Can you clarify what utility you're looking for by double-wrapping value types?

Pubsub allows people to send typed messages. String versus bytes messages is a salient discrimination. I think you could make a reasonable case for helper methods if you really wanted to? But I'm not so sure it's genuinely helpful. If they published a BytesValue they probably want a BytesValue from the stream. Do you have other use cases in mind?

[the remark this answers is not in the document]
pgautier404: I'm referring to, for example, a `CacheGetHitResponse` response type in this SDK, which exposes both `ValueString()` and `ValueBytes()` methods. We don't use `CacheGetHitResponseString` and `CacheGetHitResponseBytes` objects with `Value()` methods. That lack of consistency is all I'm asking about, but if this pattern is a natural fit for pubsub then that's fine.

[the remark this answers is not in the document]
kvcache: Oh, got you. The older unary cache get/set methods are bytes-only. If we had them to over we might choose a different model. The Stringification methods are quality of life things.

Pubsub is different. We actually reify String and byte[] all the way through. It's a feature! =)

momento/simple_cache_client.go
eaddingtonwhite: made this return error for now if we get back unexpected CacheGetResponse type other then Miss or Hit.

internal/services/pubsub_data_service.go
```diff
@@ -66,16 +66,35 @@ func (client *PubSubClient) Subscribe(ctx context.Context, request *models.Topic
...
 func (client *PubSubClient) Publish(ctx context.Context, request *models.TopicPublishRequest) error {
-	_, err := client.unaryGrpcClient.Publish(ctx, &pb.XPublishRequest{
-		CacheName: request.CacheName,
-		Topic:     request.TopicName,
-		Value: &pb.XTopicValue{
-			Kind: &pb.XTopicValue_Text{
-				Text: request.Value,
+	switch request.Value.(type) {
+	case string:
+		_, err := client.unaryGrpcClient.Publish(ctx, &pb.XPublishRequest{
```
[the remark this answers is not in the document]
eaddingtonwhite: Yeah I agree. I need to look at how were generating the proto stubs again. I think maybe we can rework this to stop IDE from wanting us to do import alias like this. I can play with it some more and do follow up pr.

cprice404:
Sorry @eaddingtonwhite I haven't had a chance to look at this earlier.  Today I reviewed it in the github UI but still haven't had time to pull it down and play with it in an IDE.

So far I am on board with all of @kvcache 's suggestions.  Do we have a path forward that we're all on the same page about or are there any particular things you need more 👀 on?

internal/models/requests.go
```diff
@@ -24,6 +24,57 @@ type CacheGetRequest struct {
 	Key interface{}
 }
 
+type CacheSetRequest struct {
+	CacheName  string
+	Key        interface{}
+	Value      interface{}
```
kvcache, line 30: These seem not great, and like a footgun to the user. Any reason not to use `type momento.Bytes interface { isBytes() }` and `struct RawBytes` + `struct StringBytes`?

internal/models/responses.go
```diff
@@ -37,41 +25,22 @@ func NewCacheInfo(cache *pb.XCache) CacheInfo {
...
-}
+func (_ CacheGetMiss) isCacheGetResponse() {}
 
-type TopicSubscribeRequest struct {
-	CacheName string
-	TopicName string
+// CacheGetHit Hit Response to a cache Get api request.
+type CacheGetHit struct {
+	Value []byte
 }
```
kvcache, line 40: this is nice

kvcache (commented):
This is looking a lot better. I have 1 remaining UX question

internal/models/requests.go
eaddingtonwhite: Yes totally. was thinking about this as well after this pr and iteration on types. I was going to add to this PR last night but this is getting so big already I was thinking could make this change in a follow up PR. Probably ok for this PR since that functionality is not changing today from where it was. That work for you? can tag u in follow up pr also.

eaddingtonwhite:
No I think this is gtg @cprice404 im in alignment w/ @kvcache also I think have all changes we want in for now. I want to make the one more change to CacheGet Key and Body that we were talking in follow up pr though since this is getting so large. So if I can get final 👍 from either you or @kvcache will merge this one now I think should be gtg.

internal/models/requests.go
kvcache: A follow up for this sounds great!

kvcache (approved):
Topic types look good to me. I like the new safe types direction and am looking forward to seeing the request types for plain scalars become explicit about their supported types and intents too.

PR #78, opened 2023-02-01
feat: adds types for keys and values

eaddingtonwhite:


momento/cache_requests.go
```diff
@@ -1,5 +1,22 @@
 package momento
 
+// Bytes generic interface to help users deal with passing us bytes
+type Bytes interface{ isBytes() }
```
kvcache, line 4: What if you put a function like `AsByteArray() []byte` on this interface? That would let you trivially implement the byte array conversion for each of the supported types (and any future supported type) as well as _separate concerns_ between data type conversion and validation.

Below, conflating validation with type conversion makes the method Is*Valid families confusing, and I think the encode methods are probably doing too much work?

momento/simple_cache_client.go
```diff
@@ -187,3 +216,63 @@ func convertCacheInfo(i []models.CacheInfo) []CacheInfo {
 	}
 	return convertedList
 }
+
+func encodeKey(key Bytes) ([]byte, momentoerrors.MomentoSvcErr) {
+	switch k := key.(type) {
+	case StringBytes:
+		return isKeyValid([]byte(k.Text))
+	case *StringBytes:
+		return isKeyValid([]byte(k.Text))
```
kvcache, line 225: How could there be a `*StringBytes` when key is `Bytes` not `*Bytes`?

momento/simple_cache_client.go
```diff
@@ -187,3 +216,63 @@ func convertCacheInfo(i []models.CacheInfo) []CacheInfo {
...
+		// If target is not string or byte[] then throw error for now. In future should do marshaling here.
+		return nil, momentoerrors.NewMomentoSvcErr(
+			momentoerrors.InvalidArgumentError,
+			"error encoding cache key only support momento.StringBytes or momento.RawBytes currently",
+			nil,
+		)
+	}
+}
+
+func encodeValue(value Bytes) ([]byte, momentoerrors.MomentoSvcErr) {
```
kvcache, line 240: this is a copypaste with only a string message varying. I think you want to DRY up this stuff a tad

momento/simple_cache_client_test.go
```diff
@@ -464,20 +464,32 @@ func TestSetGet(t *testing.T) {
...
 				}
 			} else {
 				// set string key/value with different ttl
-				err := client.Set(ctx, &CacheSetRequest{CacheName: testCacheName, Key: tt.key, Value: tt.value, TTLSeconds: TTL(tt.ttl)})
+				err := client.Set(ctx, &CacheSetRequest{
+					CacheName:  testCacheName,
+					Key:        &StringBytes{Text: tt.key},
+					Value:      &StringBytes{Text: tt.value},
+					TTLSeconds: TTL(tt.ttl),
+				})
```
kvcache, line 482: I think this reads really nicely

kvcache (commented):
this will make regular scalar apis look much nicer. It will be easier for users to use the sdk by moving errors closer to the compiler. Thanks for circling back to fix it up!

momento/simple_cache_client.go
eaddingtonwhite: I had broke out this way because wanted to come back and add support for handling complex objects was original thought process on why breaking out more even though is bit redundant. Let me circle back though I like idea of attaching AsBytes() to interface let me try that and circle back work to dry up more.

momento/simple_cache_client_test.go
eaddingtonwhite: yes agree feels safer as well

momento/simple_cache_client.go
eaddingtonwhite: `key` is `Bytes` interface but this allows user to be able to pass either `StringBytes{Text: "test"}` or `&StringBytes{Text: "test"}` which in go are both valid to adhearing to our `Bytes` interface but would get error when passing later before if didnt handle pointer type.

eaddingtonwhite: this resolves w/ helper funcs as well which nice can just scrap having to do this.

eaddingtonwhite:
@kvcache good call on the asBytes() function on Bytes interface 👍

kvcache (approved):
works for me

PR #82, opened 2023-02-02
chore: add check for context cancel or timeout

pgautier404:


incubating/simple_cache_client_test.go
```diff
@@ -33,121 +87,141 @@ func TestBasicHappyPathLocalPubSub(t *testing.T) {
...
+		publishTopic(client, i, ctx)
+		// Call the cancel function here and make sure it stops the messages.
+		if i == cancelAtNumber {
+			cancelFunction()
 		}
 		time.Sleep(time.Second)
 	}
+
+	// Ensure cancelFunction is called to quiet a context leak warning
+	cancelFunction()
```
eaddingtonwhite, line 117: nit I think this is ok to leave as is. One thing you could do is right when cancelFunction is defined could call 
`defer cancelFunction()` to ensure this is called when this overall function exits

eaddingtonwhite (approved):
Nice @pgautier404 Thanks for jumping on this!

incubating/simple_cache_client_test.go
pgautier404: Ah, yes, gotcha. I'll make that change in a followup PR, thanks!

eaddingtonwhite: yeah sounds good!

PR #85, opened 2023-02-06
chore: update `README` template

### LINKED ISSUE #70: Clean up main `README`
Update `README` to align with other SDK's `README`.
Example from our .NET SDK: https://github.com/momentohq/client-sdk-dotnet/blob/main/README.template.md

poppoerika:
Updated `README` template to provide Go SDK Error handling. 
Also cleaned up the requirements section.

Closes #70

### DIFF
--- a/README.template.md
+++ b/README.template.md
@@ -35,6 +35,7 @@ This makes them more visible, and allows your IDE to be more helpful in ensuring
 
 The preferred way of interpreting the return values from `ScsClient` methods is using a `switch` statement to match and handle the specific response type. 
 Here's a quick example:
+
 ```go
 switch r := resp.(type) {
 case *momento.CacheGetHit:
@@ -50,6 +51,7 @@ Using this approach, you get a type-safe `CacheGetHit` object in the case of a c
 But if the cache read results in a Miss, you'll also get a type-safe object that you can use to get more info about what happened.
 
 In cases where you get an error response, it can be treated as `momentoErr` using `As` method and it always include an `momentoErr.Code` that you can use to check the error type:
+
 ```go
 err = client.CreateCache(ctx, &momento.CreateCacheRequest{
     CacheName: cacheName,
--- a/README.template.md
+++ b/README.template.md
@@ -39,9 +39,9 @@ Here's a quick example:
 switch r := resp.(type) {
 case *momento.CacheGetHit:
     log.Printf("Lookup resulted in cahce HIT. value=%s\n", r.ValueString())
-default:
-	// you can handle other cases via pattern matching in other `switch case`, or a default case
-    // via the `default` block.  For each return value your IDE should be able to give you code
+default: 
+    // you can handle other cases via pattern matching in other `switch case`, or a default case
+    // via the `default` block.  For each return value your IDE should be able to give you code 
     // completion indicating the other possible "case"; in this case, `*momento.CacheGetMiss`.
 }
 ```
--- a/README.template.md
+++ b/README.template.md
@@ -30,9 +30,6 @@ Here is a quickstart you can use in your own project:
 
 ### Error Handling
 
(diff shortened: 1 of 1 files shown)


README.template.md
```diff
@@ -7,15 +7,10 @@
 - [Go version 1.18.\*](https://go.dev/dl/)
 - A Momento Auth Token is required, you can generate one using
   the [Momento CLI](https://github.com/momentohq/momento-cli)
-- golang
-  - `brew install go`
```
poppoerika, line 11: During the bug bash, people had an issue with this brew command so I removed it for now.
People can still reference the donwload link in the Go website on the top.

README.template.md
```diff
@@ -7,15 +7,10 @@
...
-- golint
-  - `go get -u golang.org/x/lint/golint`
```
poppoerika, line 13: I removed this linter since it's deprecated. https://github.com/golang/lint
We should try alternatives to linter if we want to recommend users to install. Or, we could suggest users to run `go vet` after installing this linter but I don't want to suggest users to install a deprecated package..

README.template.md
cprice404: +1 installing things like go via brew is bad juju

README.template.md
```diff
@@ -35,7 +30,41 @@ Here is a quickstart you can use in your own project:
 
 ### Error Handling
 
-Coming soon...
+Errors that occur in calls to `ScsClient` methods are surfaced to developers as part of the return values of the calls, as opposed to by throwing exceptions. 
+This makes them more visible, and allows your IDE to be more helpful in ensuring that you've handled the ones you care about. (For more on our philosophy about this, see our blog post on why [Exceptions are bugs](https://www.gomomento.com/blog/exceptions-are-bugs). And send us any feedback you have!)
```
cprice404, line 34: These two sentences might not be necessary in go since that is already the paradigm in this language.  @eaddingtonwhite ?

README.template.md
```diff
@@ -35,7 +30,41 @@ Here is a quickstart you can use in your own project:
...
+}
+```
+
+Using this approach, you get a type-safe `CacheGetHit` object in the case of a cache hit. 
+But if the cache read results in a Miss, you'll also get a type-safe object that you can use to get more info about what happened.
+
+In cases where you get an error response, it can be treated as `momentoErr` using `As` method and it always include an `momentoErr.Code` that you can use to check the error type:
+
+```go
+err = client.CreateCache(ctx, &momento.CreateCacheRequest{
```
cprice404, line 56: i'm a little out of date on this SDK but I would have expected this to be `resp, err = ...`

README.template.md
```diff
@@ -35,7 +30,41 @@ Here is a quickstart you can use in your own project:
...
+    CacheName: cacheName,
+})
+if err != nil {
+    var momentoErr momento.MomentoError
+    if errors.As(err, &momentoErr) {
+        if momentoErr.Code() != momento.AlreadyExistsError {
```
cprice404, line 62: @eaddingtonwhite are we going to hoist this AlreadyExists up to not be an error?  That i what we have done in the other SDKs.

schwern (commented):
I agree with Chris' comments.

README.template.md
eaddingtonwhite: good call can make follow up ticket for that missed the create cache response types.

README.template.md
eaddingtonwhite: For api calls where we were just only returning a empty success response seemed bit strange in go to just have a response type with nothing on it. If were going to add a CacheAlreadyExists response type in for CreateCache() function then I think makes more sense to not just return error.

README.template.md
poppoerika: I can update the example above to use `TIMEOUT_ERROR`

cprice404 (approved):

PR #88, opened 2023-02-07
chore: add single msg recv

pgautier404:


incubating/subscription.go
```diff
@@ -10,14 +10,42 @@ import (
...
+		}
+		return nil, err
+	}
+
+	var msgBytes []byte
+	switch typedMsg := rawMsg.Kind.(type) {
+	case *pb.XSubscriptionItem_Discontinuity:
+		// Don't pass discontinuity messages back to user for now
+		// TODO decide how want to notify client
+		return nil, nil
```
eaddingtonwhite, line 36: hmm this seems bit odd to me... maybe we should call RecvMsg again until they get non `Discontinuity` type message since were not exposing to user yet? Just seems weird to just return nil to user here to me.

@kvcache - interested ur thoughts on short term user experience here. Can sync in slack also if easier.

incubating/subscription.go
```diff
@@ -10,14 +10,42 @@ import (
 )
 
 type SubscriptionIFace interface {
-	Recv(ctx context.Context, f func(ctx context.Context, m TopicValue)) error
+	Consume(ctx context.Context, f func(ctx context.Context, m TopicValue)) error
+	Recv() ([]byte, error)
```
eaddingtonwhite, line 14: Maybe we return our Bytes wrapper interface instead?
https://github.com/momentohq/client-sdk-go/blob/main/momento/simple_cache_client_test.go#L55-L56

so like
```go
Recv() (momento.Bytes, error)
```

then could swith on `*momento.StringBytes` or `*momento.RawBytes` ?

incubating/subscription.go
kvcache: Since in Go we are skipping the discontinuity yes we should skip this value.

incubating/subscription.go
kvcache: This should not be returning a raw []byte, that's the wrong type as noted with the typecast on line 40 below.

This should not be `Bytes` either though, it needs to be `TopicValue`.

kvcache, line 14: Since you're redefining what this function is and does, I think you should make the name reflect what it is and does - how about `Item()`, `ReceiveOne()`, `Next()`, `Pop()` or `Poll()` off the top of my head? Something else that reflects what this function does would be fine too.

eaddingtonwhite: I think main thing im thinking about is I think Consume and Recv should have similar types returned to user. Feels bit weird to me to have Recv return `TopicValueBytes` or `TopicValueString` and this return just `[]byte`.

incubating/subscription.go
```diff
@@ -10,14 +10,42 @@ import (
 )
 
 type SubscriptionIFace interface {
-	Recv(ctx context.Context, f func(ctx context.Context, m TopicValue)) error
+	Consume(ctx context.Context, f func(ctx context.Context, m TopicValue)) error
```
kvcache, line 13: If we have `Next()` below, why are we also eagerly reimplementing a concurrency & structure construct? Shouldn't we only expose 1 way to use the service, and make that be the simplest way possible?

It might be helpful to review how I handled this  in rust: It would be good to review how I did this in Rust: https://github.com/momentohq/client-sdk-rust/blob/main/src/preview/topics.rs#L167-L180 A `Subscription` has 1 public function, `item()` which returns you the next SubscriptionItem (or None if it's all over, per the rust stream idea)

incubating/subscription.go
eaddingtonwhite: sorry @kvcache reloaded saw ur comment now. Yes agree w/ you I think.

incubating/subscription.go
eaddingtonwhite: I personally do like having simple helper method for blocking I think is common pattern makes for minimal simple code for end user.  Vs making them loop on Recv. I can get behind just having less to support though and wait for user feedback if others feel strongly we should not have this for now. Just still seems pretty thin to me were not getting to involved with how messages are handled just invoking a call back on each new message for the user and then they can decide if they want to dump on a channel or how to process.

@pgautier404 wdut?

kvcache: I don't think it is hard for a user who knows go to write
```
for {
    item := subscription.Next()
    useThe(item)
}
```
It is a plain method call.

Contrast that with a method that [1] is a higher order method (more complex) and [2] requires a `Context` (I don't even know what that is tbh). 0 parameters versus 2 parameters 😄. One of them is hard to use incorrectly while the other, I'm not sure what you can do with a Context.

That said, if the higher order function were implemented as above instead of as a copy paste of `Next()` then it seems maybe fine? But I'm looking at line 60 and  wouldn't that thing never terminate if another message didn't come? Is it even using Context correctly? (I have no idea, it just looks fishy)

eaddingtonwhite: ok ack. I think convincing me. agree lets keep it thinner for now to start just expose single way to start.

pgautier404: I think the single item `Next()` style is simpler and more flexible for users. If there's a need for the complexity of the callback style, we can add it later.

incubating/subscription.go
eaddingtonwhite: Hmm so I think on first glance Item() or Next() would probably be my preference. If we already did Item() in rust SDK maybe we stick to that so is consistent?

poppoerika (commented):
Are we implementing this single message recv in addition to `consume` to provide options for users? 
Is this how most subscriber APIs provide how to receive/consume items from the topic? (one at a time vs keep consuming till EOF?)

pgautier404:
> Sorry if I missed any contexts to this PR but I have a few questions for my own knowledge. Are we implementing this single message recv in addition to `consume` to provide options for users? Is this how most subscriber APIs provide how to receive/consume items from the topic? (one at a time vs keep consuming till EOF?)

@poppoerika We decided to remove the `Consume()` function and go with the simpler `Item()` function that simply returns one message per call. It's easy enough in golang for users to implement `Consume` style message handling by just looping over calls to `Item()` as Kenny pointed out [above](https://github.com/momentohq/client-sdk-go/pull/88#discussion_r1099074874).

eaddingtonwhite (approved):
This is looking good to me @pgautier404 I like the new interface it looks nice.

PR #90, opened 2023-02-09
feat: adds sorted set initial impl in incubating

eaddingtonwhite:
Opening so can sync w/ @pgautier404 and discuss more will merge this to main after pr from his branch is merged.


incubating/simple_cache_client.go
```diff
@@ -206,6 +214,147 @@ func (c *DefaultScsClient) ListPushBack(ctx context.Context, request *ListPushBa
...
+		NumberOfResults: convertSortedSetFetchNumResultsRequest(request.NumberOfResults),
+	})
+	if err != nil {
+		return nil, convertMomentoSvcErrorToCustomerError(err)
+	}
+	switch r := rsp.(type) {
+	case *models.SortedSetFetchMissing:
+		return SortedSetFetchMissing{}, nil
+	case *models.SortedSetFetchFound:
+		return &SortedSetFetchFound{
```
pgautier404, line 254: Can we stick with `Hit` and `Miss` for these instead of `Found` and `Missing`?

incubating/simple_cache_client.go
```diff
@@ -206,6 +214,147 @@ func (c *DefaultScsClient) ListPushBack(ctx context.Context, request *ListPushBa
...
+		ElementNames: momentoBytesListToPrimitiveByteList(request.ElementNames),
+	})
+	if err != nil {
+		return nil, convertMomentoSvcErrorToCustomerError(err)
+	}
+	switch r := rsp.(type) {
+	case *models.SortedSetGetScoreMissing:
+		return &SortedSetGetScoreMissing{}, nil
+	case *models.SortedSetGetScoreFound:
+		return &SortedSetGetScoreFound{
```
pgautier404, line 284: Same re hit/miss vs found/missing.

incubating/simple_cache_client.go
```diff
@@ -206,6 +214,147 @@ func (c *DefaultScsClient) ListPushBack(ctx context.Context, request *ListPushBa
...
+		return &SortedSetGetRankMissing{}, nil
+	case *models.SortedSetGetRankFound:
+		if r.Status == models.Hit {
+			return &SortedSetGetRankFound{
+				Element: &SortedSetRankHit{
+					Rank: r.Rank,
+				},
+			}, nil
+		} else if r.Status == models.Miss {
+			return &SortedSetGetRankFound{
```
pgautier404, line 339: This should probably be `&SortedSetGetRankMissing`?

eaddingtonwhite: So this is bit weird I agree. It when set is found but then element you requested is missing... Let me push up change to make Hit/Miss on overall set quick so is consistent and then lets look at this again.

incubating/sortedset_requests.go
```diff
@@ -0,0 +1,90 @@
...
+	Score float64
+}
+
+type SortedSetRemoveRequestElement struct {
+	Name momento.Bytes
+}
+
+type SortedSetPutRequest struct {
+	CacheName     string
+	SetName       momento.Bytes
```
pgautier404, line 26: I thought we decided to go with strings for collection names and stick with bytes for the element names?

eaddingtonwhite: O yes missed this. good catch.

internal/models/sortedset_response.go
```diff
@@ -0,0 +1,62 @@
+package models
+
+type CacheResult int32
+
+const (
+	Hit  CacheResult = 2
+	Miss CacheResult = 3
```
pgautier404, line 7: Any particular reason to skip 1? Just curious.

incubating/sortedset_requests.go
```diff
@@ -0,0 +1,90 @@
...
+	Elements      []*SortedSetScoreRequestElement
+	CollectionTTL incubating.CollectionTtl
```
pgautier404: This is now `utils.CollectionTtl`.

internal/models/sortedset_response.go
eaddingtonwhite: These map directly to grpc enums there is also `invalid` and `OK` but we dont seem to use anywhere?
```
enum ECacheResult {
  Invalid = 0;
  Ok = 1;
  Hit = 2;
  Miss = 3;

  reserved 4 to 6;
}
```

makes it easy to go from GRPC type to internal type was reasoning.

momento/cache_requests.go
```diff
@@ -3,14 +3,14 @@ package momento
 import "time"
 
 // Bytes generic interface to help users deal with passing us bytes
-type Bytes interface{ asBytes() []byte }
+type Bytes interface{ AsBytes() []byte }
```
pgautier404, line 6: Good catch on these.

pgautier404 (approved):
LGTM!

Issue #101, opened 2023-02-10
feat!: Every SimpleCacheClient returns a response and error.


schwern:
Right now, some methods only return an error. That means sometimes its `resp, err = client.op(...)` and sometimes its `err = client.op(...)`. This is an extra thing to check per method, and potentially forward incompatible if we decide we do need to return a response.

- [ ] All SimpleCacheClient methods return a response, even if just Success.
  - Users can use the normal Go idiom for ignoring the response `_, err := client.op(...);`

PR #104, opened 2023-02-11
feat!: Consolidate method implementations

### LINKED ISSUE #97: Rearchitect scalar ops
Set a new faster, slimmer pattern for developing methods.

- Implementing an op should be four pieces of code
  1. [ ] SimpleCacheClient wrapper
  2. [ ] Initialize the gRPC request (get.Request.init_request())
  3. [ ] Interpret the response (get.ResponseBuilder)
  4. [ ] Response methods (get.Hit.value, etc)
- Extract request and response scaffolding into common code.
  - [ ] Logging and tracing
  - [ ] Common validation
  - [ ] Exception wrapping
  - [ ] Common argument conversions (TTL, for example)
- [ ] All the code for one op (get, set, delete, etc...) goes in one file.
- [ ] An op's types all live together: get.Hit, get.Request, get.Miss, get.Error.

schwern:
The code for each method has been consolidated into one file. Boilerplate code is now handled in scsDataClient.

* Cache name validation is extracted
* Intermediate models are gone, they were a lot of extra work.
* ScsDataClient interface has been removed. It's costly and unnecessary.
* ScsDataClient methods now return a value, not a pointer. This may be reversed.
* List methods are not implemented. @pgautier404 will follow up.
  * The list example has been retained; it's currently broken.
* Follow ups #113, #114, #111, #110, #101, #115
  
Closes #97


momento/delete.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+
+	r.grpcRequest = &client_sdk_go.XDeleteRequest{CacheKey: key}
+
+	return nil
+}
+
+func (r *DeleteRequest) makeGrpcRequest(client scsDataClient, metadata context.Context) (grpcResponse, error) {
+	resp, err := client.grpcClient.Delete(metadata, r.grpcRequest)
```
pgautier404, line 52: Calling the context variable `metadata` here made me double take. This is conventionally passed as `ctx` and I think it would reduce confusion to follow that convention here.

momento/scs_data_client.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	if err := r.initGrpcRequest(client); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, client.requestTimeout)
+	defer cancel()
+
+	metadata := metadata.NewOutgoingContext(
```
pgautier404, line 67: My IDE is flagging this:

image

momento/delete.go
```diff
@@ -0,0 +1,65 @@
+package momento
+
+import (
+	"context"
+
+	client_sdk_go "github.com/momentohq/client-sdk-go/internal/protos"
```
pgautier404, line 6: `client_sdk_go` is really not an accurate name for the protos package. Can we go back to using `pb` or pick a different alias?

momento/get.go
```diff
@@ -0,0 +1,94 @@
+package momento
+
+import (
+	"context"
+
+	client_sdk_go "github.com/momentohq/client-sdk-go/internal/protos"
```
pgautier404, line 6: Same comment about protos alias.

momento/get.go
```diff
@@ -0,0 +1,94 @@
...
+	resp := r.grpcResponse
+
+	if resp.Result == client_sdk_go.ECacheResult_Hit {
+		r.response = GetHit{value: resp.CacheBody}
+		return nil
+	} else if resp.Result == client_sdk_go.ECacheResult_Miss {
+		r.response = GetMiss{}
+		return nil
+	} else {
+		return errUnexpectedGrpcResponse
```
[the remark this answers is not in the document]
schwern: I agree. That will take a little thought about doing the equivalent of exception classes in Go. I put it in as #114.

examples/pubsub-example/main.go
```diff
@@ -37,30 +36,30 @@ func main() {
...
 		item, err := sub.Item()
 		if err != nil {
 			panic(err)
 		}
 		switch msg := item.(type) {
-		case *incubating.TopicValueString:
+		case *momento.TopicValueString:
 			fmt.Printf("received message: '%s'\n", msg.Text)
-		case *incubating.TopicValueBytes:
+		case *momento.TopicValueBytes:
```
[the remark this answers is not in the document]
schwern: Those are response methods. The responses haven't changed.

Maybe that's a good argument for returning pointer responses, make all the type switches consistent.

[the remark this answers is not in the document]
eaddingtonwhite: I agree with consistency part so I think if we do for other methods and dont return pointers then we should come over here and do for these response types also. 

Arent these just response types as well? Not totally following "response methods"?

momento/sorted_set_remove.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	grpcReq := &pb.XSortedSetRemoveRequest{
+		SetName: []byte(r.SetName),
+	}
+
+	switch r := r.ElementsToRemove.(type) {
+	case *RemoveAllElements:
+		grpcReq.RemoveElements = &pb.XSortedSetRemoveRequest_All{}
+	case *RemoveSomeElements:
```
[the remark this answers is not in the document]
schwern: For now I've left it as it was originally. I think we need to have a more coordinated discussion about pointers vs values.

I've added a default.

momento/delete.go
schwern: The metadata is more than just the context. It is also the cache name. Passing it in avoids having to repeat...

```
	defer cancel()

	metadata := metadata.NewOutgoingContext(
		ctx, client.CreateNewMetadata(r.cacheName()),
	)
```

pgautier404: Ah, ok, I didn't realize the `NewOutgoingContext` call had already been applied here.

examples/pubsub-example/main.go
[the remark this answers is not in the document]
schwern: You're right, they're response types.

I'm coming around to just returning pointers for everything.

pgautier404 (approved):

PR #121, opened 2023-02-14
feat: Change back to returning responses as pointers

schwern:
Closes #!15


pgautier404 (dismissed):

schwern:
I found one more reference in the README and fixed it.

PR #122, opened 2023-02-14
fix: refactor list operations

pgautier404:


momento/list_concatenate_front.go
```diff
@@ -0,0 +1,92 @@
...
+		Values:             values,
+		TtlMilliseconds:    ttl,
+		RefreshTtl:         r.CollectionTTL.RefreshTtl,
+		TruncateBackToSize: r.TruncateBackToSize,
+	}
+
+	return nil
+}
+
+func (r *ListConcatenateFrontRequest) makeGrpcRequest(client scsDataClient, ctx context.Context) (grpcResponse, error) {
```
eaddingtonwhite, line 77: ctx should always be first arg to func in golang. I think maybe govet should fail on this?

momento/list_remove_value.go
```diff
@@ -0,0 +1,71 @@
...
+
+	r.grpcRequest = &pb.XListRemoveRequest{
+		ListName: []byte(r.ListName),
+		Remove:   &pb.XListRemoveRequest_AllElementsWithValue{AllElementsWithValue: value},
+	}
+
+	return nil
+}
+
+func (r *ListRemoveValueRequest) makeGrpcRequest(client scsDataClient, ctx context.Context) (grpcResponse, error) {
```
eaddingtonwhite, line 57: same here should context be first?

momento/list_push_front.go
```diff
@@ -0,0 +1,92 @@
...
+		Value:              value,
+		TtlMilliseconds:    ttl,
+		RefreshTtl:         r.CollectionTTL.RefreshTtl,
+		TruncateBackToSize: r.TruncateBackToSize,
+	}
+
+	return nil
+}
+
+func (r *ListPushFrontRequest) makeGrpcRequest(client scsDataClient, ctx context.Context) (grpcResponse, error) {
```
eaddingtonwhite, line 78: ctx first arg?

eaddingtonwhite (approved):
one nit on how were passing context in internal method not as first arg I think we should fix. I think that pattern was set before this PR though so im ok merging this we can iterate on future pr on that.

momento/list_concatenate_front.go
pgautier404: Ok, yeah, this argument order is currently baked into all of the request classes. I'll fix it in a quick followup.

PR #129, opened 2023-02-15
feat!: Ensure all methods return a response.

### LINKED ISSUE #101: feat!: Every SimpleCacheClient returns a response and error.

schwern:
The one oddball is ListCaches(). It seems like it should only ever return one response (as Go is not returning error responses) so making the user do the type switch seems unnecessary. But that's the interface pattern. 🤷 And it would make it the oddball that doesn't return an interface.

Closes #101

### DIFF
--- a/examples/list-example/main.go
+++ b/examples/list-example/main.go
@@ -252,7 +252,7 @@ func main() {
 	printList()
 
 	// Delete the cache
-	if err = client.DeleteCache(ctx, &momento.DeleteCacheRequest{CacheName: cacheName}); err != nil {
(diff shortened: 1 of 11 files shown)


eaddingtonwhite (approved):
This looks good to me @schwern agree w/ what you did for now on list cache. Id rather keep like this for now and be consistent and then can do follow up pr to go to pattern like you did w/ ruby sdk and do a higher level iterator.

PR #135, opened 2023-02-15
feat!: Simplify passing values.

schwern:
momento.Bytes is named from the POV of what we need internally. Instead, name them from the POV of what the user is passing in.

* Bytes -> Value
* RawBytes -> Bytes
* StringBytes -> String

Simplify the types to be a single value, not a struct, making it a thin wrapper around the base value.

Before:

```
Value: momento.StringBytes{ Text: "abc" }
Value: momento.RawBytes{ Bytes: []byte("abc") }
```

After:

```
Value: momento.String("abc")
Value: momento.Bytes([]byte("abc"))
```

This simplifies passing in values, and *greatly* simplifies passing in multiple values.

TopicValue is left alone, pubsub is still developing.


momento/pub_sub_test.go
```diff
@@ -53,12 +53,12 @@ func setup() {
 }
 
 func publishTopic(ctx context.Context, pubClient SimpleCacheClient, i int) {
-	var topicVal TopicValue
+	var topicVal Value
```
kvcache, line 56: TopicValue is not the same thing as Value: You'll need to maintain the distinction.

eaddingtonwhite (approved):
Nice I like this approach more @schwern good idea. Would love have @pgautier404 take pass as well but this looks good to me. Happy with how it looks.

momento/set.go
```diff
@@ -23,9 +23,9 @@ type SetRequest struct {
 	// Name of the cache to store the item in.
 	CacheName string
 	// string or byte key to be used to store item.
-	Key Bytes
+	Key Value
 	// string ot byte value to be stored.
-	Value Bytes
+	Value Value
```
kvcache, line 28: This seems like a poor name choice compared to `Bytes`. What alternatives did you consider that don't conflate struct member names with types?

kvcache (commented):
TopicValue needs to stay separate from whatever else. Happy to chat with you on slack about it if you prefer

momento/pub_sub_test.go
schwern: How are they distinct?

momento/set.go
schwern: I thought about StringOrBytes, but that's a bit mechanistic and limits our future options.

There's no conflating issue because it will be used as `Value: momento.Value`. It's almost universally thing we pass in as Value, so Value.

I could make a Key type alias.

momento/pub_sub_test.go
kvcache: `Bytes` is something that is represented as bytes.

`TopicValue` is something that can be published to a topic and received on a Subscription.

If you conflate them then you will need to put it back in the near/medium-term future when we add more publishable `TopicValue` types.

schwern: What types do you anticipate adding to pubsub that won't be added to everything else?

kvcache (commented):
seems good to me

eaddingtonwhite (approved):

PR #141, opened 2023-02-16
feat: Recognize pubsub heartbeat

schwern:
This allows a subscription to stay alive even if there's nothing coming through the pipe.

This required updated the proto files.


momento/topic_subscription.go
```diff
@@ -45,6 +46,11 @@ func (s topicSubscription) Item() (TopicValue, error) {
 					Bytes: subscriptionItem.Binary,
 				}, nil
 			}
+		case *pb.XSubscriptionItem_Heartbeat:
+			// Doesn't count against our retries.
+			continue
+		default:
+			return nil, fmt.Errorf("unexpected gRPC response when polling a subscription: %T %v", typedMsg, typedMsg)
```
eaddingtonwhite, line 53: I don't think we want to fail and return error here. In rust SDK were just logging an warning when we receive an unkown topic message type. Will make it harder to add new serverside features if we hard fail here.
https://github.com/momentohq/client-sdk-rust/commit/41108e9e5121428c3438cb4ed38907a2e8657e6b

Makefile
```diff
@@ -26,6 +26,17 @@ staticcheck:
...
+.PHONY: install-protos-devtools
+install-protos-devtools:
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+	go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+
+.PHONY: protos
+protos:
+	@echo "Run `make install-protos-devtools` if you're missing the Go grpc tools."
+	@echo "Did you copy momentohq/client_protos/protos/*.proto to internal/protos?"
+	protoc -I=internal --go_out=internal --go_opt=paths=source_relative --go-grpc_out=internal --go-grpc_opt=paths=source_relative internal/protos/*.proto
```
eaddingtonwhite, line 38: thanks for adding these 👍

cprice404 (commented):
lgtm other than what ellery called out

momento/topic_subscription.go
schwern: Just for this op so it continues to poll? Leave the other ops so they will error if they get an unknown response?

Should it count against the retries?

eaddingtonwhite: I would think for pubsub we would not want to retry just wait for another message and move on. For other parts in SDK is good call out. I think we should handle consistently in all our SDK's. Would you make a follow up issue for that and lets make sure were aligned internal on behavior and then can make sure we follow that approach throughout.

schwern: > I would think for pubsub we would not want to retry just wait for another message and move on.

`continue` same as a heartbeat response?

eaddingtonwhite: maybe log a warning like we are doing in rust sdk? Im fine though just continuing for now in this PR. We are working very closely with anyone using this feature right now so will be working w/ them on upgrades.

schwern: I think logging a warning and then continue. The other option is to exit the loop which might as well be an error.

Except we don't have logging. 🏌️

eaddingtonwhite: yup lets just `continue` for now agree.

internal/protos/cacheclient.proto
```diff
@@ -455,18 +455,45 @@ message _SortedSetFetchRequest {
...
   bytes set_name = 1;
   Order order = 2;
   oneof num_results {
     _All all = 3;
     _Limit limit = 4;
   }
+  oneof range {
+    _ByIndex by_index = 5;
+    _ByScore by_score = 6;
+  }
```
cprice404, line 496: i presume the above is just an artifact of copying the latest proto down?

cprice404 (approved):
lgtm assuming the answer to my question about the proto changes is "yes", and pending 👍 from @eaddingtonwhite

internal/protos/cacheclient.proto
eaddingtonwhite: yup these are new sorted set pieces for doing range queries on a sorted set. We havent added to go sdk yet but its in protos already.

eaddingtonwhite (approved):

PR #146, opened 2023-02-16
chore: add dictionary operations

pgautier404:


momento/dictionary_fetch.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (resp DictionaryFetchHit) ValueMapStringString() map[string]string {
+	if resp.itemsStringString == nil {
+		resp.itemsStringString = make(map[string]string)
+		for k, v := range resp.items {
+			resp.itemsStringString[k] = string(v)
+		}
+	}
+	return resp.itemsStringString
+}
```
schwern, line 34: We've gone back and forth about whether we should cache the string versions. My concern is it potentially eats a lot of memory, depends on whether `string(v)` copies the memory or not.

momento/dictionary_fetch.go
```diff
@@ -0,0 +1,94 @@
...
+// DictionaryFetchResponse
+
+type DictionaryFetchResponse interface {
+	isDictionaryFetchResponse()
+}
+
+type DictionaryFetchHit struct {
+	items             map[string][]byte
+	itemsStringString map[string]string
+}
```
schwern, line 18: What about byte fields?

momento/dictionary_get_field.go
```diff
@@ -0,0 +1,107 @@
...
+
+// DictionaryGetFieldResponse
+
+type DictionaryGetFieldResponse interface {
+	isDictionaryGetFieldResponse()
+}
+
+type DictionaryGetFieldHit struct {
+	field []byte
+}
```
schwern, line 17: Where's the value?

momento/dictionary_get_field.go
```diff
@@ -0,0 +1,107 @@
...
+	var fields [][]byte
+	fields = append(fields, field)
+
+	r.grpcRequest = &pb.XDictionaryGetRequest{
+		DictionaryName: []byte(r.DictionaryName),
+		Fields:         fields,
+	}
+
+	return nil
+}
```
schwern, line 82: DictionaryGetField doesn't need its own implementation. It can be a thin wrapper around DictionaryGetFields. DictionaryGetField just needs its request and responses and translates.

momento/dictionary_get_fields.go
```diff
@@ -0,0 +1,128 @@
...
+			} else if val.Result == pb.ECacheResult_Miss {
+				field = r.Fields[idx].asBytes()
+				responses = append(responses, &DictionaryGetFieldMiss{field: field})
+			} else {
+				field = r.Fields[idx].asBytes()
+				responses = append(responses, nil)
+			}
+			fields = append(fields, field)
+		}
+		r.response = &DictionaryGetFieldsHit{fields: fields, items: rtype.Found.Items, responses: responses}
```
schwern, line 123: Rather than storing the fields and responses and then building a map in the value methods, could we just build and store a map right here?

The fields and responses are presumably stored so a user can check which fields are a hit and which are a miss, but it's simpler to check if the field exists in the map.

momento/dictionary_increment.go
```diff
@@ -0,0 +1,100 @@
...
+
+type DictionaryIncrementSuccess struct {
+	value int64
+}
+
+func (DictionaryIncrementSuccess) isDictionaryIncrementResponse() {}
+
+func (resp DictionaryIncrementSuccess) ValueUint32() int64 {
+	return resp.value
+}
```
schwern, line 28: Type mismatch. ValueUint32 returns an int64.

The user can see what it returns from the return type, can it be Value()?

momento/dictionary_increment.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var field []byte
+	if field, err = prepareField(r); err != nil {
+		return err
+	}
+
+	var ttl uint64
+	if ttl, err = prepareTTL(r, client.defaultTtl); err != nil {
+		return err
+	}
```
schwern, line 67: Use `ttlMills, refreshTTL := prepareCollectionTtl(r.CollectionTTL, client.defaultTtl)`

momento/dictionary_remove_field.go
```diff
@@ -0,0 +1,70 @@
...
+		return err
+	}
+	fields = append(fields, field)
+	r.grpcRequest = &pb.XDictionaryDeleteRequest{
+		DictionaryName: []byte(r.DictionaryName),
+		Delete:         &pb.XDictionaryDeleteRequest_Some_{Some: &pb.XDictionaryDeleteRequest_Some{Fields: fields}},
+	}
+
+	return nil
+}
```
schwern, line 56: Can be a thin wrapper around DictionaryRemoveFields.

momento/dictionary_set_field.go
```diff
@@ -0,0 +1,96 @@
...
+	var ttl uint64
+	if ttl, err = prepareTTL(r, client.defaultTtl); err != nil {
+		return err
+	}
+
+	var items []*pb.XDictionaryFieldValuePair
+	items = append(items, &pb.XDictionaryFieldValuePair{
+		Field: field,
+		Value: value,
+	})
```
schwern, line 72: Can be a thin wrapper around DictionarySetFields.

momento/dictionary_set_fields.go
```diff
@@ -0,0 +1,90 @@
...
+		pbItems = append(pbItems, &pb.XDictionaryFieldValuePair{
+			Field: []byte(k),
+			Value: v,
+		})
+	}
+
+	var ttl uint64
+	if ttl, err = prepareTTL(r, client.defaultTtl); err != nil {
+		return err
+	}
```
schwern, line 66: Use prepareCollectionTTL

momento/requester.go
```diff
@@ -81,6 +93,28 @@ func prepareKey(r hasKey) ([]byte, error) {
...
 
+func prepareField(r hasField) ([]byte, error) {
+	field := r.field().asBytes()
+
+	if len(field) == 0 {
+		err := momentoerrors.NewMomentoSvcErr(momentoerrors.InvalidArgumentError, "field cannot be empty", nil)
+		return nil, convertMomentoSvcErrorToCustomerError(err)
+	}
+	return field, nil
+}
```
schwern, line 104: The check can be extracted into a function that takes Value and returns []byte, error. Something like...

```
func innerPrepareField(f Value) ([]byte, error) {
        field := f.asBytes()

	if len(field) == 0 {
		err := momentoerrors.NewMomentoSvcErr(momentoerrors.InvalidArgumentError, "field cannot be empty", nil)
		return nil, convertMomentoSvcErrorToCustomerError(err)
	}
	return field, nil
}

func prepareField(r hasField) ([]byte, error) {	
	return prepareField(r.field())
}

func prepareFields(r hasFields) ([][]byte, error) {
	var fields [][]byte
	for _, f := range r.fields() {
	    field, err := innerPrepareField(f)
	    if err != nil { return err }
	    fields = append(fields, field)
	}
	return fields, nil
}
```

momento/requester.go
```diff
@@ -104,6 +138,14 @@ func prepareValues(r hasValues) ([][]byte, momentoerrors.MomentoSvcErr) {
 	return values, nil
 }
 
+func prepareItems(r hasItems) (map[string][]byte, momentoerrors.MomentoSvcErr) {
```
schwern, line 141: The function never returns an error. 🤔  I guess that's ok?

schwern (changes requested):

momento/dictionary_fetch.go
pgautier404: Go doesn't allow byte array map keys.

momento/dictionary_get_field.go
pgautier404: Yeah, thanks, not sure what was going on here. Looks like I meant to come back to this but never did. It's fixed/finished in 2439adce.

momento/dictionary_fetch.go
pgautier404: Who's we and did you capture the discussion anywhere? I get the memory concerns but also have concerns about performance if these methods are called multiple times. I'm happy to remove the caching code if that's the consensus but would appreciate more context on the discussion.

momento/dictionary_get_fields.go
pgautier404: Responses are not part of a map. They're exposed as their own list as was done in the node SDK. I didn't see that they're used anywhere besides testing but have them here for completeness and consistency.

momento/dictionary_increment.go
pgautier404: Exposing `CollectionTtl.ttl` using a `ttl()` method in the requester satisfies `hasScalarTtl()` (which I think could just be called `hasTtl()`). How I have this set up is functionally equivalent to `prepareCollectionTtl` without the need for adding the new method. I think this is a cleaner and DRYer approach and would suggest that we remove `prepareCollectionTtl` and adopt this approach for collection types instead. If we ever need extra specific processing for collection ttls we can add it then. What do you think?

momento/dictionary_set_fields.go
pgautier404: See above.

momento/dictionary_fetch.go
schwern: Wat.

momento/dictionary_get_fields.go
schwern: tl;dr: I'd say remove responses and fields, the user can do everything with the map value. Once we publish those fields we can't take it back, but we can always add them back later.

I think the responses are an implementation detail that leaked out. It's not in [the SDK docs](https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083?pvs=4#74c7425802484341aeca59958f68cb22). Nothing else exposes intermediate responses like that. It doesn't have a purpose that's not fulfilled by the map.

If we make the responses public we have to store all the data twice, once in the responses, once in the items, and a third time for the fields. That's an interface promise, we can't rescind it.

momento/dictionary_increment.go
schwern: We don't do that when there's only one type. See [SDK docs](https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083?pvs=4#cbbaaddadb0541bd8a91bbbc4ab313f8).

Also it's warmed over [Hungarian Notation](https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083?pvs=4#cbbaaddadb0541bd8a91bbbc4ab313f8). 🤢

momento/dictionary_increment.go
schwern: tl;dr It's an internal thing, do what you think is best as long as its consistent.

`prepareCollectionTtl` future proofs us against changes to how CollectionTTL is processed, and it reminds us to set the refreshTTL. We can add a hasCollectionTTL.

I'd keep `hasScalarTtl` to disambiguate  between time.Duration TTLs and CollectionTTLs and CollectionTTL.ttl; they're all referred to as "TTL". Maybe `hasDurationTtl`?

momento/dictionary_increment.go
```diff
@@ -0,0 +1,100 @@
...
+
+type DictionaryIncrementSuccess struct {
+	value int64
+}
+
+func (DictionaryIncrementSuccess) isDictionaryIncrementResponse() {}
+
+func (resp DictionaryIncrementSuccess) ValueUint64() int64 {
+	return resp.value
+}
```
schwern, line 28: Method name and return type still out of alignment. uint64 vs int64.

It's like those redundant comments that are lies.

```
// Add the thing to the thing.
removeThing(thing)
```

momento/dictionary_get_fields.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// DictionaryGetFieldsResponse
+
+type DictionaryGetFieldsResponse interface {
+	isDictionaryGetFieldsResponse()
+}
+
+type DictionaryGetFieldsHit struct {
+	items     []*pb.XDictionaryGetResponse_XDictionaryGetResponsePart
```
schwern, line 16: I don't see where `items` is being set.

schwern (changes requested):

momento/dictionary_fetch.go
schwern: Just discussions over the course of implementing this in other SDKs. There is no consensus and we haven't done profiling to see if this is a performance concern.

Its a internal issue, so we can change it later. Do what you think is best.

momento/dictionary_get_fields.go
pgautier404: It's in the Node SDK. We already can't rescind it. I'd rather ask the folks who worked on that SDK whether or not there's a reason it's there than just assume it's a mistake. I think there may be value in allowing users to verify that fields missing from the dictionaries returned to them were actually processed.

momento/dictionary_increment.go
pgautier404: We absolutely do do that in the Node SDK in DictionaryIncrement.

momento/dictionary_get_fields.go
schwern: I worked on the Node SDK and I don't know how that slipped in. We can always add it later. We don't need method-for-method parity between SDKs, our customers are unlikely to be using Momento in multiple languages. But we can discuss it in channel.

> I think there may be value in allowing users to verify that fields missing from the dictionaries returned to them were actually processed.

If you're saying they'll want to check if a field was a miss, `_, ok := valueMap[field]`. Simpler and faster and less memory than looping through a list of responses.

momento/dictionary_increment.go
pgautier404: I'd rather add special processing for collection ttls when we need to instead of introducing redundant code in anticipation (and all of the list ops already work like this so it's consistent with the approach I was taking before you added the method). I'm just going to continue to reuse `hasScalarTtl()` so I don't have to add a new functionally equivalent interface method to all the existing types. We can easily add a new interface if we ever need to specialize more.

I also added a validation check to `prepareTTL` to make sure we weren't returning values <= 0.

momento/requester.go
pgautier404: Good callout. I refactored it a bit differently in 7f96990d (added a generalized validator instead of another specialized `prepare` method) and got it DRYed out a bit.

momento/dictionary_increment.go
pgautier404: Fixed in 7f96990d.

momento/dictionary_get_fields.go
pgautier404: https://github.com/momentohq/client-sdk-go/blob/2439adced0d353f78620d129242bbaa7866aa56c/momento/dictionary_get_fields.go#L120

momento/dictionary_get_fields.go
pgautier404: A miss will not be in valueMap.

momento/requester.go
pgautier404: Good catch, that should have been validating the keys. Fixed.

momento/dictionary_get_fields.go
schwern: Do what you think is best.

Chris and Landis and I discussed it in a deveco Slack thread. We didn't come to a consensus.

momento/requester.go
```diff
@@ -104,21 +135,30 @@ func prepareValues(r hasValues) ([][]byte, momentoerrors.MomentoSvcErr) {
...
 	}
-	return uint64(ttl.Milliseconds()), nil
+	return retMap, nil
 }
 
-func prepareCollectionTtl(ttl utils.CollectionTTL, defaultTtl time.Duration) (uint64, bool) {
-	ttlDuration := defaultTtl
-	if ttl.Ttl != time.Duration(0) {
-		ttlDuration = ttl.Ttl
+func prepareTTL(r hasScalarTTL, defaultTtl time.Duration) (uint64, error) {
```
schwern, line 149: Should it be `hasTTL` now?

momento/dictionary_increment.go
pgautier404: We definitely are copying what the 1.0 SDKs do now, but it's also `Value` in .NET so I'll go along.

schwern (approved):
lgtm. Thanks for fielding all my feedback.

pgautier404:
> lgtm. Thanks for fielding all my feedback.

Totally, thanks for providing it.

PR #147, opened 2023-02-16
fix: Rewrite the tests using Ginkgo and Gomega, fix some bugs.

schwern:
The march of spec testing continues.

This uses [Ginkgo](https://onsi.github.io/ginkgo/) and [Gomega](https://onsi.github.io/gomega/) for spec style test organization and matchers.

Bug fixes:
* CreateCache should respond AlreadyExists when the cache already exists, not an error.
* DeleteCache should succeed when the cache does not exist.
* Values are allowed to be blank.
* CredentialProvider should return InvalidArgument error when the token cannot be parsed, not a generic ClientSdkError.

Closes #98

PR #149, opened 2023-02-17
fix: Update golang.org/x/net to 0.7.0 for minor security vulnerability.

schwern:
https://github.com/advisories/GHSA-vvpx-j8f3-3w6h

#147 tripped this vulnerability.


eaddingtonwhite (approved):
@schwern LGTM thanks for jumping on this. I would make this `fix` for these not `chore` IMO so we get a new minor version on sdk and can do release

PR #151, opened 2023-02-17
feat: handle h2 go-away from server for topic stream

### LINKED ISSUE #74: Handle Idle Connection timeout and End of Stream errors as happy path for streams
Currently when we receive either of the following errors our topic stream gets ended. We should try to re-establish subscription when we encounter these specific errors and consider them part of happy path operations.

On idle stream connection termination by envoy:
`rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: NO_ERROR`

eaddingtonwhite:
Closes #74


momento/topic_subscription.go
```diff
@@ -1,57 +1,92 @@
...
 }
+
+func (s *topicSubscription) handleStreamError(ctx context.Context, err error) error {
+	var returnErr error
+	if err == io.EOF {
+		returnErr = s.reInitStream(ctx)
+	}
+	if grpcStatusErr, ok := status.FromError(err); ok {
+		if grpcStatusErr.Code() == codes.Internal &&
+			grpcStatusErr.Message() == "stream terminated by RST_STREAM with error code: NO_ERROR" {
```
schwern, line 68: Error message strings are prone to change. Is there a way to check without string matching the message? Can we do a substring match on NO_ERROR?

momento/topic_subscription.go
```diff
@@ -1,57 +1,92 @@
...
 		}
 	}
 }
+
+func (s *topicSubscription) handleStreamError(ctx context.Context, err error) error {
+	var returnErr error
+	if err == io.EOF {
+		returnErr = s.reInitStream(ctx)
+	}
+	if grpcStatusErr, ok := status.FromError(err); ok {
```
schwern, line 66: Should this be an `else if`?

momento/topic_subscription.go
eaddingtonwhite: good call. Im playing w/ http2 package in go now. Might be able to do.
```go
	if e, ok := err.(http2.GoAwayError); ok {
		if e.ErrCode == http2.ErrCodeNo {
			fmt.Println(fmt.Sprintf("got goaway error %+v", e))
		}
	}
```

eaddingtonwhite: ^not working w/ h2 package for sure.. I think problem is because this is wrapped in GRPC error we need to handle as grpc error like I am now. I'm not seeing much in http2 package for matching on grpc error description that will work. I'm little hesitant to just match on `NO_ERROR` this was working well with testing but maybe will defer to @kvcache on this one as he has gone pretty deep w/ h2 spec here as well.

momento/topic_subscription.go
```diff
@@ -1,57 +1,91 @@
...
+					return &TopicValueString{
+						Text: subscriptionItem.Text,
+					}, nil
+				case *pb.XTopicValue_Binary:
+					return &TopicValueBytes{
+						Bytes: subscriptionItem.Binary,
+					}, nil
+				}
+			case *pb.XSubscriptionItem_Heartbeat:
+				// Doesn't count against our retries.
```
pgautier404, line 50: Retries have been removed, so the comment should probably go too.

momento/topic_subscription.go
eaddingtonwhite: I agree makes sense doesnt type cast. 

So raw error were getting if just dump it out is RPC Status error from go grpc package
```go
rpc error: 
  code = Internal
  desc = stream terminated by RST_STREAM with error code: NO_ERROR
```
Can see here I think they are wrapping all errors back from Recv call with this error type
https://github.com/grpc/grpc-go/blob/12db695f1648ce84ad74c5240730fb9401f67844/stream.go#L986-L1053

Here is snip it from go GRPC stream client recv message docs as well seems to be what they are saying
```
	// RecvMsg blocks until it receives a message into m or the stream is
	// done. It returns io.EOF when the stream completes successfully. On
	// any other error, the stream is aborted and the error contains the RPC
	// status.
```

Im not seeing a clean way to hook right into network errors receved w/ go grpc library like you can do w/ tonic in rust SDK @kvcache.

Im thinking maybe we could define our own `DialOption` when initlizing grpc client and then hook in at lower level. Im looking [here](https://github.com/grpc/grpc-go/blob/master/dialoptions.go#L69). I see the underlying default http2 client that gets created has a goaway channel that can be hooked into. [Here](https://github.com/grpc/grpc-go/blob/55dfae6e5becc928314500087c01690708deef72/internal/transport/http2_client.go#L82) 

Id probably want to make a follow up ticket though to explore this further im not finding good references online of someone doing this in clean way. I agree though its weird that this is getting passed back up as internal error and not handled more gracefully by go sdk in error passed back.

momento/topic_subscription.go
```diff
@@ -1,57 +1,91 @@
...
+					}, nil
+				case *pb.XTopicValue_Binary:
+					return &TopicValueBytes{
+						Bytes: subscriptionItem.Binary,
+					}, nil
+				}
+			case *pb.XSubscriptionItem_Heartbeat:
+				continue
+			default:
+				// Ignore unknown responses, so we don't stop polling if we add a new message.
```
cprice404, line 52: do we have a ticket to log something at debug for these two cases

momento/topic_subscription.go
cprice404: if this makes a big improvement in the UX for 1.0 then we should do it, but the exact string match for the message does indeed seem sketchy and fragile.  if we are going to consider merging this as-is, we at least need a ticket for follow-up work and we need to link that ticket here.

It also might be helpful to share a branch of how the example code looks before/after this change so folks will have a better understanding of the UX impact.

eaddingtonwhite: In terms of client experience user would currently be forced to handle this error them selves to keep the stream alive. Also were not handling this nicely at the moment and returning a interrupted exception or something similar so requires user to parse error themselves which not ideal(seperate but related issue).

before
```go
	// Instantiate subscriber
	sub, err := c.MomentoClient.SubscribeTopic(r.Context(), &incubating.TopicSubscribeRequest{
		CacheName: serviceconfig.CacheName,
		TopicName: chatRoomName,
	})
	if err != nil {
		// fatal error
	}
	for {
		item, err := sub.Item()
		if err != nil {
			if s, ok := status.FromError(err); ok {
				if s.Code() == codes.Internal &&
					s.Message() == "stream terminated by RST_STREAM with error code: NO_ERROR" {

					// Re-establish connection
					sub, err = c.MomentoClient.SubscribeTopic(r.Context(), &incubating.TopicSubscribeRequest{
						CacheName: "mycache",
						TopicName: "mytopic",
					})
					if err != nil {
                                                //fatal error
					}

				} else {
                                       //fatal error
				}
			} else {
                                //fatal error
			}
		}
		// Handle message
	}
```
vs after
```go
	// Instantiate subscriber
	sub, err := c.MomentoClient.SubscribeTopic(r.Context(), &incubating.TopicSubscribeRequest{
		CacheName: serviceconfig.CacheName,
		TopicName: chatRoomName,
	})
	if err != nil {
             // fatal error
	}
	for {
		item, err := sub.Item()
		if err != nil {
			// fatal error
		}
		// Handle message
	}
```

I made this follow up ticket to investigate handling at a lower level I agree not ideal to try and parse the grpc wrapped error returned.

https://github.com/momentohq/client-sdk-go/issues/156

momento/topic_subscription.go
eaddingtonwhite: I just made this ticket https://github.com/momentohq/client-sdk-go/issues/157

I agree should at least log warning

momento/topic_subscription.go
cprice404: that UX looks much much better.

Would you mind putting a link to the ticket in a comment in the code?  then i'm 👍

kvcache: Yeah, hooking the goaway channel seems like the right thing to do; but it also seems fine to defer it with a ticket!

kvcache: good find; this is a good step toward hiding a goaway reconnect seamlessly behind Item()

eaddingtonwhite: @cprice404 added some follow up FIXME comments 👍

cprice404 (approved):

PR #159, opened 2023-02-22
fix: add testing for sets and some type fixes

pgautier404:


momento/set_test.go
```diff
@@ -0,0 +1,396 @@
...
+	It("errors when the cache is missing", func() {
+		cacheName := uuid.NewString()
+		setName := uuid.NewString()
+
+		fetchResp, err := client.SetFetch(ctx, &SetFetchRequest{
+			CacheName: cacheName,
+			SetName:   setName,
+		})
+		Expect(fetchResp).To(BeNil())
+		Expect(err).To(HaveMomentoErrorCode(NotFoundError))
```
schwern, line 92: I checked and it's safe to write these like so. It will fail if the response is non-nil.

```suggestion
		Expect(
			client.SetFetch(ctx, &SetFetchRequest{
				CacheName: cacheName,
				SetName:   setName,
			}),
		).Error().To(HaveMomentoErrorCode(NotFoundError))
```

pgautier404: Cool, I'll switch over to the less verbose form for these.

momento/simple_cache_client.go
```diff
@@ -278,7 +278,7 @@ func (c defaultScsClient) SetAddElement(ctx context.Context, r *SetAddElementReq
 	if err := c.dataClient.makeRequest(ctx, newRequest); err != nil {
 		return nil, err
 	}
-	return SetAddElementSuccess{}, nil
+	return &SetAddElementSuccess{}, nil
```
eaddingtonwhite, line 281: good catch on these

eaddingtonwhite (approved):

PR #161, opened 2023-02-23
chore: add examples for dictionaries and sets

pgautier404:


examples/dictionary-example/main.go
```diff
@@ -0,0 +1,255 @@
...
+	client, err = momento.NewSimpleCacheClient(&momento.SimpleCacheClientProps{
+		Configuration:      config.LatestLaptopConfig(),
+		CredentialProvider: credentialProvider,
+		DefaultTTL:         itemDefaultTTLSeconds * time.Second,
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// Create Cache and check if CacheName exists
```
eaddingtonwhite, line 43: do we still need to "check if CacheName exists" I think since thats response type now not error maybe not correct comment anymore?

pgautier404: Good catch. Fixed in b204931.

examples/set-example/main.go
```diff
@@ -0,0 +1,145 @@
...
+	client, err = momento.NewSimpleCacheClient(&momento.SimpleCacheClientProps{
+		Configuration:      config.LatestLaptopConfig(),
+		CredentialProvider: credentialProvider,
+		DefaultTTL:         itemDefaultTTLSeconds * time.Second,
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// Create Cache and check if CacheName exists
```
eaddingtonwhite, line 43: same here on comment sorry just noticed this one also

eaddingtonwhite (dismissed):

examples/set-example/main.go
pgautier404: Fixed in e358576.

eaddingtonwhite (approved):

PR #164, opened 2023-02-24
fix: inspect single field response to determine hit or miss

pgautier404:


momento/simple_cache_client.go
```diff
@@ -412,10 +412,17 @@ func (c defaultScsClient) DictionaryGetField(ctx context.Context, r *DictionaryG
 	case *DictionaryGetFieldsMiss:
 		return &DictionaryGetFieldMiss{}, nil
 	case *DictionaryGetFieldsHit:
-		return &DictionaryGetFieldHit{
-			field: rtype.fields[0],
-			body:  rtype.items[0].CacheBody,
-		}, nil
+		switch rtype.responses[0].(type) {
```
eaddingtonwhite, line 415: should we check length make sure we have 1 field returned here before trying to grab 0 index?

pgautier404: The request is for exactly one field, so in this context it's safe as we know we got a hit. It's the same logic that is pulling the field and body properties out to populate the hit response.

eaddingtonwhite: Yeah was just thinking be extra defensive not trust the API but I agree should not happen.

eaddingtonwhite (approved):

Issue #178, opened 2023-02-24
Return `IllegalArgument` error when no elements to remove/add/update passed to Collection apis


eaddingtonwhite:
Currently we return a server error if you call `DictionaryGetFields()`  with no elements to fetch. We should return an IllegalArgument error in this case. We should also audit other collection apis and make sure we fail in consistent way. 
```go
	_, err = client.DictionaryGetFields(ctx, &momento.DictionaryGetFieldsRequest{
		CacheName:      "foo",
		DictionaryName: "test",
	})
```

PR #194, opened 2023-02-25
chore: add dictionary testing

pgautier404:


momento/dictionary_test.go
```diff
@@ -0,0 +1,676 @@
...
+			).Error().To(BeNil())
+			getFieldResp, err := sharedContext.Client.DictionaryGetField(sharedContext.Ctx, &DictionaryGetFieldRequest{
+				CacheName:      sharedContext.CacheName,
+				DictionaryName: sharedContext.CollectionName,
+				Field:          field,
+			})
+			Expect(err).To(BeNil())
+			Expect(getFieldResp).To(BeAssignableToTypeOf(&DictionaryGetFieldHit{}))
+			switch result := getFieldResp.(type) {
+			case *DictionaryGetFieldHit:
```
eaddingtonwhite, line 119: should we add default on this switch w/ an explicit fail if hits?

momento/dictionary_test.go
```diff
@@ -0,0 +1,676 @@
...
+
+	It("returns an error for set field when field is empty", func() {
+		Expect(
+			sharedContext.Client.DictionarySetField(sharedContext.Ctx, &DictionarySetFieldRequest{
+				CacheName:      sharedContext.CacheName,
+				DictionaryName: sharedContext.CollectionName,
+				Field:          String(""),
+				Value:          String("myValue"),
+			}),
+		).Error().To(HaveMomentoErrorCode(InvalidArgumentError))
```
eaddingtonwhite, line 140: maybe add case for nil also? I had made this [ticket](https://github.com/momentohq/client-sdk-go/issues/178) for nil `Field` or `Value` also so can wait till we do that if easier come back in one pass.

momento/dictionary_test.go
```diff
@@ -0,0 +1,676 @@
...
+
+	It("returns an error if an item field is empty", func() {
+		Expect(
+			sharedContext.Client.DictionarySetFields(sharedContext.Ctx, &DictionarySetFieldsRequest{
+				CacheName:      sharedContext.CacheName,
+				DictionaryName: sharedContext.CollectionName,
+				Items:          map[string]Value{"myField": String("myValue"), "": String("myOtherValue")},
+			}),
+		).Error().To(HaveMomentoErrorCode(InvalidArgumentError))
+	})
```
eaddingtonwhite, line 194: same here for nil check. Maybe just add TODO comment in test for now?

eaddingtonwhite (approved):
Awesome job on these @pgautier404 !! Loving the new test lib were using looks great.

momento/dictionary_test.go
pgautier404: I've already asserted that it is assignable to that type above. The only purpose for the switch is to get the cast.

momento/dictionary_test.go
pgautier404: Good call, I'll add that in a followup.

momento/dictionary_test.go
pgautier404: Same, will follow up on this in my list testing PR.

PR #196, opened 2023-02-25
test: SortedSet

### LINKED ISSUE #116: Collections default ttl is broken (sorted sets)
When incrementing an element in a sorted set, if you don't pass in a ttl, we should be using the default ttl. For unary operations, we check to make sure that the request ttl isnt 0, and if it isnt we use it, else we use the default client configured ttl

```
func (c *DefaultScsClient) Set(ctx context.Context, request *CacheSetRequest) error {
	if err := isCacheNameValid(request.CacheName); err != nil {
		return err
	}
	ttlToUse := c.defaultTTL
	if request.TTL != time.Duration(0) {
		ttlToUse = request.TTL
	}
```

We need to perform a similar check for sorted sets ttls. These ttls are a bit different since they are more than just a duration, they also have a boolean flag to determine if we should refresh the ttl of the item or not `CollectionTTL: utils.CollectionTTL{RefreshTtl: true, Ttl: 60 * time.Second},`.

This code below should be respecting the default 60 second ttl that was set for the cache client, however items are getting evicted immediately.
```
client, err := incubating.NewScsClient(&momento.SimpleCacheClientProps{
	Configuration:      config.LatestLaptopConfig(),
	CredentialProvider: credProvider,
	DefaultTTL:         60 * time.Second,
})
if err != nil {
	panic(err)
}
	
response, err := client.SortedSetIncrement(ctx, &incubating.SortedSetIncrementRequest{
		CacheName:   cacheName,
		SetName:     "game1",
		ElementName: momento.StringBytes{Text: "Allen"},
		Amount:      2,
	})

	if err != nil {
		panic(err)
	} else {
		switch r := response.(type) {
		case *incubating.SortedSetIncrementResponseSuccess:
			fmt.Println(fmt.Sprintf("new value %f", r.Value))
		}
	}
```

As a part of this ticket, we should also audit other collections apis to make sure they also respect the default ttl

### LINKED ISSUE #195: SortedSetPutRequest takes a SortedSetScoreRequestElement
```
type SortedSetPutRequest struct {
	CacheName     string
	SetName       string
	Elements      []*SortedSetScoreRequestElement
	CollectionTTL utils.CollectionTTL

	grpcRequest  *pb.XSortedSetPutRequest
	grpcResponse *pb.XSortedSetPutResponse
	response     SortedSetPutResponse
}
```

schwern:
Complete tests for sorted sets.

* Verifies that collectionTTLs work, Closes #116.
* SortedSetFetch took a surprising SortedSetScoreRequestElement. Closes #195

### DIFF
--- a/momento/sorted_set_test.go
+++ b/momento/sorted_set_test.go
@@ -18,7 +18,7 @@ var _ = Describe("SortedSet", func() {
 	})
 
 	// A convenience for adding elements to a sorted set.
-	var putElements = func(elements []*SortedSetScoreRequestElement) {
+	putElements := func(elements []*SortedSetPutElement) {
 		Expect(
 			sharedContext.Client.SortedSetPut(
 				sharedContext.Ctx,
@@ -84,7 +84,7 @@ var _ = Describe("SortedSet", func() {
 				}),
 			).Error().To(HaveMomentoErrorCode(InvalidArgumentError))
 
-			putElements := []*SortedSetScoreRequestElement{{
+			putElements := []*SortedSetPutElement{{
 				Value: element,
 				Score: float64(1),
 			}}
@@ -129,7 +129,7 @@ var _ = Describe("SortedSet", func() {
 
 		It(`Fetches`, func() {
 			putElements(
-				[]*SortedSetScoreRequestElement{
+				[]*SortedSetPutElement{
 					{Value: String("first"), Score: 9999},
 					{Value: String("last"), Score: -9999},
 					{Value: String("middle"), Score: 50},
@@ -210,7 +210,7 @@ var _ = Describe("SortedSet", func() {
 
 		It(`Gets the rank`, func() {
 			putElements(
-				[]*SortedSetScoreRequestElement{
+				[]*SortedSetPutElement{
 					{Value: String("first"), Score: 9999},
 					{Value: String("last"), Score: -9999},
 					{Value: String("middle"), Score: 50},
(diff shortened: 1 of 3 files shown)


momento/sorted_set_test.go
```diff
@@ -0,0 +1,533 @@
...
+			Expect(
+				client.SortedSetFetch(ctx, &SortedSetFetchRequest{
+					CacheName: badName, SetName: collectionName,
+				}),
+			).Error().To(HaveMomentoErrorCode(InvalidArgumentError))
+			Expect(
+				client.SortedSetFetch(ctx, &SortedSetFetchRequest{
+					CacheName: cacheName, SetName: badName,
+				}),
+			).Error().To(HaveMomentoErrorCode(InvalidArgumentError))
```
eaddingtonwhite, line 65: I feel like was bit easier to read the way @pgautier404 did [here](https://github.com/momentohq/client-sdk-go/pull/194/files#diff-e5b04cc1cb89537fc2404c0ad0b013fc3954df1218ba70e47b3dfacf47bd41cdR27-R98) with `expectedErrorCode` bassed in w/ table test entry. Dry'd up test a bit more?

momento/sorted_set_test.go
```diff
@@ -0,0 +1,533 @@
...
+					CacheName:   sharedContext.CacheName,
+					SetName:     sharedContext.CollectionName,
+					ElementName: element.Value,
+					Amount:      element.Score,
+				}
+				if ttl != nil {
+					request.CollectionTTL = *ttl
+				}
+
+				sharedContext.Client.SortedSetIncrementScore(sharedContext.Ctx, request)
```
eaddingtonwhite, line 206: hmm the golint error here seems right? Should we be expecting something?

momento/sorted_set_test.go
schwern: Yeah, I saw that as well.

eaddingtonwhite (approved):

PR #201, opened 2023-02-27
fix: Topic name and cache validation

### LINKED ISSUE #198: Test topics
Before we make a whole bunch of changes, get basic tests in place.

schwern:
Testing topic name validation revealed some problems.

* TopicSubscribe and TopicPublish were not validating names.
* TopicSubscribe did not error when the cache did not exist.
* TopicPublish was not wrapping its error.

Closes #198

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 .idea
 .DS_Store
 vendor
+.env
+.vscode
--- a/momento/simple_cache_client.go
+++ b/momento/simple_cache_client.go
@@ -3,6 +3,7 @@ package momento
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -10,6 +11,7 @@ import (
 	"github.com/momentohq/client-sdk-go/config"
 	"github.com/momentohq/client-sdk-go/internal/models"
 	"github.com/momentohq/client-sdk-go/internal/momentoerrors"
+	pb "github.com/momentohq/client-sdk-go/internal/protos"
 	"github.com/momentohq/client-sdk-go/internal/services"
 )
 
@@ -190,6 +192,14 @@ func (c defaultScsClient) Delete(ctx context.Context, r *DeleteRequest) (DeleteR
 }
 
 func (c defaultScsClient) TopicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (TopicSubscription, error) {
+	if err := isCacheNameValid(request.CacheName); err != nil {
+		return nil, err
+	}
+
+	if _, err := prepareName(request.TopicName, "Topic name"); err != nil {
+		return nil, err
+	}
+
 	clientStream, err := c.pubSubClient.TopicSubscribe(ctx, &TopicSubscribeRequest{
 		CacheName: request.CacheName,
 		TopicName: request.TopicName,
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -1,3 +1,5 @@
 .idea
 .DS_Store
-vendor
\ No newline at end of file
+vendor
+.env
+.vscode
```
poppoerika, line 5: I take this as you enjoying VSCode? 😄

schwern: It's growing on me. Now that I have the debugger configured it's very useful.

momento/simple_cache_client.go
```diff
@@ -190,30 +192,68 @@ func (c defaultScsClient) Delete(ctx context.Context, r *DeleteRequest) (DeleteR
...
+	err = clientStream.RecvMsg(rawMsg)
+	if err != nil {
+		return nil, momentoerrors.NewMomentoSvcErr(
+			momentoerrors.NotFoundError,
+			fmt.Sprintf("Did not get a heartbeat from topic %v in cache %v", request.TopicName, request.CacheName),
+			err,
+		)
+	}
+	switch rawMsg.Kind.(type) {
+	case *pb.XSubscriptionItem_Heartbeat:
```
eaddingtonwhite, line 222: Did we check w/ @kvcache will always be a heart beat first? There no chance of `XSubscriptionItem_Discontinuity` during deployment or error scenario re-connecting? Will always be heart beat first?

eaddingtonwhite (approved):
Looks ok to me. Just want to double check and make sure @kvcache is ok w/ the initial message handling on stream. I just worry about server side changing down line this is bit of unwritten contract that heartbeat must come first.

momento/simple_cache_client.go
schwern: Yes, Kenny said that's the case. He suggested the default to raise an error should there be a server side bug. I'll add him as a reviewer.

kvcache: It is the express intent now and forever that every subscription must receive a distinct, immediate Heartbeat message upon subscription before any other message in the subscription stream. That special heartbeat message is called "The Hello Heartbeat." It is specifically intended as an affordance by the server for SDK authors to provide better experiences to SDK users.

It was originally added to move the GRPC response state into 1 particular known state where we can optionally gracefully recover from certain classes of disconnection events. However it is not prescriptive about limitations to its utility. For example, herewith it is going to move the stack at which a user encounters "cache not found" to a more reasonable / obvious / intuitive frame.

If that subscription stream behavior changes, it is a service bug and should be met with the appropriate rollback and general triage + fix process.

momento/simple_cache_client.go
```diff
@@ -190,30 +192,68 @@ func (c defaultScsClient) Delete(ctx context.Context, r *DeleteRequest) (DeleteR
...
-	}, err
+	}, nil
 }
 
 func (c defaultScsClient) TopicPublish(ctx context.Context, request *TopicPublishRequest) (TopicPublishResponse, error) {
+	if err := isCacheNameValid(request.CacheName); err != nil {
+		return nil, err
+	}
+
+	if _, err := prepareName(request.TopicName, "Topic name"); err != nil {
```
kvcache, line 245: ` ` is a perfectly valid topic name. Any particular reason to relabel it as a synthetic client-side, go-only `momentoerrors.InvalidArgumentError`?

momento/simple_cache_client.go
```diff
@@ -190,30 +192,68 @@ func (c defaultScsClient) Delete(ctx context.Context, r *DeleteRequest) (DeleteR
...
+	// Ping the stream to provide a nice error message if the cache does not exist.
+	rawMsg := new(pb.XSubscriptionItem)
+	err = clientStream.RecvMsg(rawMsg)
+	if err != nil {
+		return nil, momentoerrors.NewMomentoSvcErr(
+			momentoerrors.NotFoundError,
+			fmt.Sprintf("Did not get a heartbeat from topic %v in cache %v", request.TopicName, request.CacheName),
+			err,
+		)
+	}
```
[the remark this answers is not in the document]
schwern: You tell me, what other errors might I expect?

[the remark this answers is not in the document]
kvcache: Here are a couple links from a brief google search for how to use grpc:
* Here is the go tutorial about grpc (header link to particularly relevant docs for your convenience): https://grpc.io/docs/languages/go/basics/#streaming-rpcs
* And here are the definitions for grpc Status codes: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

We don't define any additional codes, so that list is comprehensive.

momento/simple_cache_client.go
schwern: To be clear, this PR is just adding the [existing name validation](https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083?pvs=4#cff6d0d8adc84c51a053247c04533529) to topics.

I have a slight preference for relaxing the restriction. Fortunately we can relax the restriction later.

momento/simple_cache_client.go
schwern: I think that's "yes".

PR #202, opened 2023-02-27
chore: add list testing

pgautier404:
More testing is on the way for all collection data types regarding struct defaults, nil requests, and related issues. This commit brings the list type testing to parity with the other types to this point.

### DIFF
--- a/momento/list_test.go
+++ b/momento/list_test.go
@@ -1,0 +1,126 @@
+package momento_test
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+	. "github.com/momentohq/client-sdk-go/momento"
+	. "github.com/momentohq/client-sdk-go/momento/test_helpers"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func getValueAndExpectedValueLists() ([]Value, []string) {
+	var values []Value
+	var expected []string
+	for i := 0; i < 10; i++ {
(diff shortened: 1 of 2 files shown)


momento/list_test.go
```diff
@@ -0,0 +1,537 @@
...
+	values, expected := getValueAndExpectedValueLists(numItems)
+	for _, value := range values {
+		Expect(
+			sharedContext.Client.ListPushBack(sharedContext.Ctx, &ListPushBackRequest{
+				CacheName: sharedContext.CacheName,
+				ListName:  sharedContext.CollectionName,
+				Value:     value,
+			}),
+		).To(BeAssignableToTypeOf(&ListPushBackSuccess{}))
+	}
```
schwern, line 40: Can this be replaced with ListConcatenateBack?

momento/list_test.go
```diff
@@ -0,0 +1,537 @@
...
+
+	BeforeEach(func() {
+		sharedContext = NewSharedContext()
+		sharedContext.CreateDefaultCache()
+		DeferCleanup(func() {
+			sharedContext.Close()
+		})
+	})
+
+	DescribeTable("try using invalid cache and list names",
```
schwern, line 55: This is missing most of the list methods.

pgautier404: Good catch, thanks.

momento/list_test.go
pgautier404: Why yes, yes it can.

schwern (approved):

PR #205, opened 2023-02-28
feat!: Make TopicValue easier to work with.

### LINKED ISSUE #171: Re-Visit Topics Request values and response types
We had spent time improving the user experience for the scalar APIs  with `momento.String()` and `momento.Bytes()` but did not have chance to come back and try improve topics. We should see if there some improvements to user experience we can make here.

schwern:
We need TopicValue because in the future topics may accept more value types than normal methods. But there's no need for different string and bytes types.

* Replace TopicValueString and TopicValueBytes with the existing String and Bytes.
* String and Bypes implement TopicValue.

Closes #171

### DIFF
--- a/momento/topic_value.go
+++ b/momento/topic_value.go
@@ -1,0 +1,16 @@
+package momento
+
+// For now, topics take the same types as normal methods.
+// In the future this might not be so, topics might take
+// different types than methods.
+//
+// The TopicValue interface alias future proofs us. Topics
+// still take the normal String and Bytes, but in the future
+// we can add types that other methods might not take.
+type TopicValue interface {
+	isTopicValue()
+}
+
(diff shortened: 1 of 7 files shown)


momento/topic_value.go
```diff
@@ -0,0 +1,16 @@
...
+// The TopicValue interface alias future proofs us. Topics
+// still take the normal String and Bytes, but in the future
+// we can add types that other methods might not take.
+type TopicValue interface {
+	isTopicValue()
+}
+
+func (String) isTopicValue() {}
+
+func (Bytes) isTopicValue() {}
```
kvcache, line 16: this is better.

pgautier404 (approved):
LGTM!

eaddingtonwhite (approved):

Issue #192, opened 2023-02-24
Audit request structs to make sure we wont get nil pointer errors or bad defaults


eaddingtonwhite:
We should take one more pass on request structs and make sure we have proper error checking or defaults on all functions request objects prior to 1.0

PR #206, opened 2023-02-28
chore: audit and test for nils

pgautier404:
This commit adds validation to prevent nil values in request structs and adds testing to verify the added validation across request types.

Addresses #192 and #178

### DIFF
--- a/momento/requester.go
+++ b/momento/requester.go
@@ -106,15 +106,13 @@ func prepareCacheName(r hasCacheName) (string, error) {
 }
 
 func prepareKey(r hasKey) ([]byte, error) {
-	err := buildError(momentoerrors.InvalidArgumentError, "key cannot be nil or empty", nil)
-
 	if r.key() == nil {
-		return nil, err
+		return nil, buildError(momentoerrors.InvalidArgumentError, "key cannot be nil or empty", nil)
 	}
 
 	key := r.key().asBytes()
 	if len(key) == 0 {
-		return nil, err
+		return nil, buildError(momentoerrors.InvalidArgumentError, "key cannot be nil or empty", nil)
 	}
 	return key, nil
 }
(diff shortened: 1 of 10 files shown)


momento/requester.go
```diff
@@ -70,29 +70,61 @@ type hasTTL interface {
...
+	return nameBytes, nil
+}
+
 func prepareCacheName(r hasCacheName) (string, error) {
 	return prepareName(r.cacheName(), "Cache name")
 }
 
 func prepareKey(r hasKey) ([]byte, error) {
-	key := r.key().asBytes()
+	err := buildError(momentoerrors.InvalidArgumentError, "key cannot be nil or empty", nil)
```
eaddingtonwhite, line 109: nit: but seems slightly off to build up error we might return at top like this especially in high traffic place like `prepareKey` maybe better just create inline for each time returned on line 112 and 117 so doesnt get allocated every time this function called? Similar to what your doing in `prepareField`.

momento/requester.go
```diff
@@ -101,8 +133,15 @@ func prepareField(r hasField) ([]byte, error) {
 }
 
 func prepareFields(r hasFields) ([][]byte, error) {
+	err := buildError(momentoerrors.InvalidArgumentError, "fields cannot be nil or empty", nil)
```
eaddingtonwhite, line 136: same thought here on error being inline on 138 and 143

momento/requester.go
```diff
@@ -155,24 +190,22 @@ func prepareTTL(r hasTTL, defaultTtl time.Duration) (uint64, error) {
...
-			nil,
+		return 0, buildError(
+			momentoerrors.InvalidArgumentError, "ttl must be a non-zero positive value", nil,
 		)
 	}
 	return uint64(ttl.Milliseconds()), nil
 }
 
 func momentoValuesToPrimitiveByteList(i []Value) ([][]byte, momentoerrors.MomentoSvcErr) {
+	err := buildError(momentoerrors.InvalidArgumentError, "values may not be nil", nil)
```
eaddingtonwhite, line 201: also here

momento/requester.go
pgautier404: That's fine. I figured it was DRYer to do it this way and relatively low impact, but I am fine with the tiny but of duplication.

pgautier404: Fixed in 0c32ba1

eaddingtonwhite (approved):
Nice work! Tests Looking great! 🧼

PR #210, opened 2023-03-01
fix: Properly interpret the error from pinging the new topic stream.

schwern:
Previously it assumed any error was a Not Found Error.


eaddingtonwhite:
@schwern im pulling topic client out into standalone client in another PR right now. I'm going to close this PR and will just make this change over there will tag you on it.

PR #212, opened 2023-03-01
chore: make `CollectionTtl` pointer and assign default value to  `Ttl` and `RefreshTtl`

### LINKED ISSUE #174: Make CollectionTTL a pointer and try to make behave like our other SDKS when no value passed

poppoerika:
Update the collection operations to accept a `CollectionTtl` pointer as a param.
If no value is passed for `Ttl`, it will default to default ttl that user specified when creating a new cache client.. 
If no value is passed for `RefreshTtl`, it will default to `true`. 

Also updated the new param structure for `NewCacheClient` in the examples.

Closes #174


poppoerika:
I made both `Ttl` and `RefreshTtl` pointers so that we receive `nil` if users don't pass any values.

momento/sorted_set_put.go
```diff
@@ -29,7 +29,7 @@ type SortedSetPutRequest struct {
 	CacheName string
 	SetName   string
 	Elements  []*SortedSetPutElement
-	Ttl       utils.CollectionTtl
+	Ttl       *utils.CollectionTtl
```
pgautier404, line 32: I think this is the only place we need to convert anything to a pointer. Then, when it's time to use the CollectionTtl, we can check if it is nil or not. The rest of this should stay as it was.

poppoerika: I thought the same thing first but I realized that I cannot determine when `Ttl.refreshTtl` is originally `false` or a user passed `false`.  The same thing for `Ttl.ttl`. Unless a user passes something to `Ttl.ttl` I don't know how we can determine no value was passed to that property. Hence the pointer. If nothing is passed, the "zero" value for a pointer is `nil`. 
I thought if nothing at all is provided for `Ttl`, then `*utils.CollectionTtl` is sufficient but if a user misses one of the properties, we need those to be pointers as well?
Maybe I'm misunderstanding something here. 🤔

pgautier404: If the CollectionTtl object is `nil` then the user hasn't passed anything at all, so there's no `ttl` or `refreshTtl` to look at. That's the case we're trying to catch so we can provide the appropriate defaults. The `ttl` and `refreshTtl` properties should stay as values so that users cannot provide nil values for them.

poppoerika: Somehow I thought users can pass one of the struct properties 🤷 But got it. Thank you.

momento/dictionary_test.go
```diff
@@ -682,7 +682,7 @@ var _ = Describe("Dictionary methods", func() {
 						DictionaryName: sharedContext.CollectionName,
 						Field:          String("foo"),
 						Value:          String("bar"),
-						Ttl:            utils.CollectionTtl{},
+						Ttl:            &utils.CollectionTtl{},
```
eaddingtonwhite, line 685: maybe just remove this fields from these tests now since we don't want defined for this test case. I think will just read a little more clear and also makes sure we handle not passed value ok. 

Also I think we should add a couple tests to just make sure error gets thrown when we expect when passing a bad value to this.

poppoerika: Yup, makes sense to me!
I will add some tests.

momento/dictionary_increment.go
```diff
@@ -61,8 +63,9 @@ func (r *DictionaryIncrementRequest) initGrpcRequest(client scsDataClient) error
 		return err
 	}
 
+	collectionTtl := prepareCollectionTtl(r, client.defaultTtl)
 	var ttl uint64
-	if ttl, err = prepareTtl(r, client.defaultTtl); err != nil {
+	if ttl, err = prepareCollectionTtlTtl(collectionTtl.Ttl, client.defaultTtl); err != nil {
```
schwern, line 68: Previously we had one function that returned the TTL  and refreshTTL values to pass to gRPC. I think that would be simpler here.

momento/dictionary_test.go
```diff
@@ -148,6 +146,18 @@ var _ = Describe("Dictionary methods", func() {
 		Entry("both nil", nil, nil),
 	)
 
+	It("try using negative ttl for set", func() {
```
schwern, line 149: It should read like a sentence describing the expected behavior.

```suggestion
	It("errors with a negative ttl", func() {
```

momento/simple_cache_client.go
```diff
@@ -239,11 +240,15 @@ func (c defaultScsClient) SetAddElements(ctx context.Context, r *SetAddElementsR
...
+	ttl := r.Ttl
+	if r.Ttl == nil {
+		ttl = assignDefaultCollectionTtl(c.dataClient.defaultTtl)
+	}
 	newRequest := &SetAddElementsRequest{
 		CacheName: r.CacheName,
 		SetName:   r.SetName,
 		Elements:  []Value{r.Element},
-		Ttl:       r.Ttl,
+		Ttl:       ttl,
```
schwern, line 251: Is it necessary to handle a nil TTL? That should be handled by making the SetAddElementsRequest.

I think this can be simplified by calling the client method instead of making a request.

```
_, err := c.SetAddElements(ctx, &SetAddElementsRequest{...})
if err != nil {
  return err
}
return &SetAddElementSuccess
```

momento/simple_cache_client.go
```diff
@@ -348,13 +353,17 @@ func (c defaultScsClient) DictionarySetField(ctx context.Context, r *DictionaryS
 			),
 		)
 	}
+	ttl := r.Ttl
+	if r.Ttl == nil {
+		ttl = assignDefaultCollectionTtl(c.dataClient.defaultTtl)
+	}
```
schwern, line 359: This should be handled by making the DictionarySetFieldsRequest?

momento/dictionary_test.go
```diff
@@ -672,32 +682,6 @@ var _ = Describe("Dictionary methods", func() {
 			).Error().To(BeNil())
 		})
 
-		When("collection TTL is empty", func() {
```
schwern, line 675: We still need this test, but the CollectionTTL is null.

momento/dictionary_increment.go
schwern: PS I understand if you don't want to redo all that work.

momento/dictionary_test.go
poppoerika: Ah sorry about that! Thank you for catching this!

eaddingtonwhite (approved):

schwern (approved):
LGTM

PR #213, opened 2023-03-01
feat: breaks topic client into standalone client

eaddingtonwhite:


momento/topic_client.go
```diff
@@ -0,0 +1,190 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// Ping the stream to provide a nice error message if the cache does not exist.
+	rawMsg := new(pb.XSubscriptionItem)
+	err = clientStream.RecvMsg(rawMsg)
+	if err != nil {
+		return nil, momentoerrors.ConvertSvcErr(err)
```
eaddingtonwhite, line 137: @schwern JFYI pulled this change over

eaddingtonwhite:
@pgautier404 Lets get https://github.com/momentohq/client-sdk-go/pull/209 merged first for sure and then I can circle back and update this to match pattern.

internal/momentoerrors/scs_errors.go
```diff
@@ -36,38 +37,38 @@ const (
 
 func ConvertSvcErr(err error) MomentoSvcErr {
 	if grpcStatus, ok := status.FromError(err); ok {
-		switch grpcStatus.Code().String() {
-		case "InvalidArgument":
+		switch grpcStatus.Code() {
+		case codes.InvalidArgument:
```
kvcache, line 41: thanks

momento/pubsub_test.go
```diff
@@ -55,7 +55,7 @@ var _ = Describe("Pubsub", func() {
 			Bytes([]byte{1, 2, 3}),
 		}
 
-		sub, err := sharedContext.Client.TopicSubscribe(sharedContext.Ctx, &TopicSubscribeRequest{
+		sub, err := sharedContext.TopicClient.TopicSubscribe(sharedContext.Ctx, &TopicSubscribeRequest{
```
kvcache, line 58: I don't know if this has come up in conversation, but maybe it'd be worth considering stripping the "topic" prefix from topic methods, given that now they're now relevantly type-spaced? It would look like `sharedContext.TopicClient.Subscribe(` here instead. I don't think the topic prefix meaningfully disambiguates anything here - what do you think?

momento/topic_client.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/momentohq/client-sdk-go/internal/services"
+
+	"github.com/momentohq/client-sdk-go/auth"
+	"github.com/momentohq/client-sdk-go/config"
+)
+
+type TopicClient interface {
+	CreateCache(ctx context.Context, request *CreateCacheRequest) (CreateCacheResponse, error)
+	DeleteCache(ctx context.Context, request *DeleteCacheRequest) (DeleteCacheResponse, error)
+	ListCaches(ctx context.Context, request *ListCachesRequest) (ListCachesResponse, error)
```
kvcache, line 20: Hmm, maybe we can talk about this? I realize an "it can do everything" client is nice in some respects, but in other respects - this is the job of a CacheClient not a TopicClient. It's confusing to not know which client to use for cache DDL actions, and to have the cognitive burden of worrying whether each does something different.
To be clear, I think these 3 DDL methods don't belong on this client. (Whether they even belong on the Cache DML client is a different conversation that I'd be willing to have sometime too ;-))

kvcache (commented):
This looks much improved, thanks! I have a couple questions

momento/pubsub_test.go
eaddingtonwhite: thats good call. I think I agree. I can make that change. 👍

momento/topic_client.go
eaddingtonwhite: I had brainstormed about this w/ @cprice404  a bit after bug bash this is where we had landed. I just thought it was a bit of not ideal user experience to have to create both a cache client and topic client to use topics so had wanted to just port control plane actions over for cache into topic client. Im not 100% sold on this though can be convinced to just remove from topic client for now. Let me start slack convo quick on this can close out on it and then will update this PR.

internal/momentoerrors/scs_errors.go
eaddingtonwhite: yeah the strings were bugging me O_O :)

momento/topic_client.go
eaddingtonwhite: removed for now.

kvcache (dismissed):
l's gtm

momento/topic_client.go
```diff
@@ -0,0 +1,123 @@
...
+	Close()
+}
+
+// defaultScsClient represents all information needed for momento client to enable cache control and data operations.
+type defaultTopicClient struct {
+	credentialProvider auth.CredentialProvider
+	pubSubClient       *pubSubClient
+}
+
+// NewTopicClient returns a new TopicClient with provided authToken, DefaultTTLSeconds, and opts arguments.
```
pgautier404, line 28:
```suggestion
// NewTopicClient returns a new TopicClient with provided configuration and credential provider arguments.
```

eaddingtonwhite: good catch accepted suggestion

pgautier404 (approved):

Issue #187, opened 2023-02-24
Make SDK tests simpler to new devs


eaddingtonwhite:
- [x] `TEST_AUTH_TOKEN` not being set resulted in nil pointer error
- [x] `TEST_CACHE_NAME` not being set worked fine but confused new dev coming on to repo if needed

schwern:
We can't use a single cache name for testing, it will conflict when tests are run simultaneously on Github or if we want to test in parallel, so TEST_CACHE_NAME is gone.

I'll remove it from the CONTRIBUTING doc.

pgautier404:
Tests were swallowing the credential provider error message informing the user that their environment variable was not set or was empty. This was fixed in 17f677ee.

Issue #176, opened 2023-02-24
Make `auth` `config` packages more navigable


eaddingtonwhite:
- [ ] should only have what user directly needs exported
- [ ] auth Provider interface threw off few users need to look into helper functions in `nodejs` and `.net` sdks (from_env_var etc)
- [ ] Was hard to discover right config objects to use they should be grouped by the type rather than the version.  so it should be more like `LaptopLatest` and `LaptopV1` instead of `LatestLaptop` and `LatestInRegion`

PR #217, opened 2023-03-01
fix: navigable auth and config packages

pgautier404:
This commit reduces the set of exported symbols from the `auth` and `config` packages down to only those that users need to interact with. It also panics on an error that it was previously swallowing when constructing the CredentialProvider for the tests so users will now see a clear explanation of the failure. 

Addresses #176 and #187

### DIFF
--- a/auth/credential_provider.go
+++ b/auth/credential_provider.go
@@ -57,10 +57,15 @@ func FromString(authToken string) (CredentialProvider, error) {
 }
 
 // WithEndpoints overrides the cache and control endpoint URIs with those provided by the supplied Endpoints struct
-// and returns a CredentialProvider with the new endpoint values
+// and returns a CredentialProvider with the new endpoint values. An endpoint supplied as an empty string is ignored
+// and the existing value for that endpoint is retained.
 func (credentialProvider defaultCredentialProvider) WithEndpoints(endpoints Endpoints) (CredentialProvider, error) {
-	credentialProvider.cacheEndpoint = endpoints.CacheEndpoint
-	credentialProvider.controlEndpoint = endpoints.ControlEndpoint
+	if credentialProvider.cacheEndpoint != "" {
+		credentialProvider.cacheEndpoint = endpoints.CacheEndpoint
+	}
+	if credentialProvider.controlEndpoint != "" {
+		credentialProvider.controlEndpoint = endpoints.ControlEndpoint
+	}
 	return credentialProvider, nil
 }
 
--- a/auth/credential_provider.go
+++ b/auth/credential_provider.go
@@ -56,6 +56,8 @@ func FromString(authToken string) (CredentialProvider, error) {
 	return credentialProvider, nil
 }
 
+// WithEndpoints overrides the cache and control endpoint URIs with those provided by the supplied Endpoints struct
+// and returns a CredentialProvider with the new endpoint values
 func (credentialProvider defaultCredentialProvider) WithEndpoints(endpoints *Endpoints) (CredentialProvider, error) {
 	if endpoints == nil {
 		return nil, momentoerrors.NewMomentoSvcErr(
--- a/auth/credential_provider.go
+++ b/auth/credential_provider.go
@@ -10,11 +10,6 @@ import (
 	"github.com/momentohq/client-sdk-go/internal/momentoerrors"
 )
 
-type ResolveRequest struct {
-	AuthToken        string
-	EndpointOverride string
-}
-
 type Endpoints struct {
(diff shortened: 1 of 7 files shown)


auth/credential_provider.go
```diff
@@ -24,28 +19,58 @@ type CredentialProvider interface {
...
-// TODO: add overrides for endpoints
+func FromEnvironmentVariable(envVar string) (CredentialProvider, error) {
+	credentialProvider, err := NewEnvMomentoTokenProvider(envVar)
+	if err != nil {
+		return nil, err
+	}
+	return credentialProvider, nil
+}
+
+func FromString(authToken string) (CredentialProvider, error) {
```
poppoerika, line 51: Yay, more option!

auth/credential_provider.go
```diff
@@ -24,28 +19,58 @@ type CredentialProvider interface {
...
+	credentialProvider, err := NewStringMomentoTokenProvider(authToken)
+	if err != nil {
+		return nil, err
+	}
+	return credentialProvider, nil
+}
+
+func (credentialProvider defaultCredentialProvider) WithEndpoints(endpoints *Endpoints) (CredentialProvider, error) {
```
poppoerika, line 59: What is the purpose of this function?

config/config.go
```diff
@@ -30,39 +30,39 @@ type Configuration interface {
 	WithClientTimeout(clientTimeout time.Duration) Configuration
 }
 
-type SimpleCacheConfiguration struct {
+type simpleCacheConfiguration struct {
```
poppoerika, line 33: Should we update this to `cacheConfiguration` to align with the recent rename of `CacheClient`?

poppoerika: The same suggestions for the below code.

auth/credential_provider.go
pgautier404, line 59: @cprice404 This is a departure from the other SDKs but is a much more natural fit for golang which doesn't allow optional or nullable non-pointer method params. Usage looks like:

```
credentialProvider, err = credentialProvider.WithEndpoints(
	&auth.Endpoints{
		ControlEndpoint: controlEndpoint,
		CacheEndpoint:   cacheEndpoint,
	},
)
```

Let me know if that looks ok to you.

pgautier404: That is to allow users to override their cache and/or control endpoint URLs.

config/config.go
pgautier404: That's in progress! I'm not quite done with this PR, and that change is on it's way :-)

auth/credential_provider.go
eaddingtonwhite: nit: maybe just add function docs on this while in here so is clear to users also

auth/credential_provider_test.go
```diff
@@ -21,4 +23,52 @@ var _ = Describe("CredentialProvider", func() {
...
+	It("returns a credential provider from a string via method", func() {
+		credentialProvider, err := auth.FromString(os.Getenv("TEST_AUTH_TOKEN"))
+		Expect(err).To(BeNil())
+		Expect(credentialProvider.GetAuthToken()).To(Equal(os.Getenv("TEST_AUTH_TOKEN")))
+	})
+
+	It("overrides endpoints", func() {
+		var err error
+		var credentialProvider auth.CredentialProvider
+		credentialProvider, err = auth.NewEnvMomentoTokenProvider("TEST_AUTH_TOKEN")
```
eaddingtonwhite, line 54: nit: usually would just write with wallrus operator for this in go and ditch the var declarations above on line 52 and 53. 
```go
credentialProvider, err := auth.NewEnvMomentoTokenProvider("TEST_AUTH_TOKEN")
```

auth/credential_provider.go
cprice404: seems fine but is there a reason anyone would ever want to call this and pass nil?  like, idk if i understand why the arg to this needs to be a pointer instead of a value but i am not opposed to it

cprice404 (approved):

auth/credential_provider.go
pgautier404: Added in 180e887

auth/credential_provider_test.go
pgautier404: Updated in 180e887

auth/credential_provider.go
pgautier404: Excellent point. Fixed in 9597f39

schwern (commented):
This is a major improvement. Looks good, just those two things.

PR #219, opened 2023-03-02
chore: revert Lint the example code

eaddingtonwhite:
This reverts commit f2c2397be875cde9a4018f0bf76ee62a90ecf3d9.

@schwern going to revert this for time being causing some churn on prs. Im going to come back will try re-add back in tomorrow and come up w/ better flow for updating examples and running linting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,3 @@
-LINTER_ARGS=./... ./examples/...
-
 .PHONY: install-devtools
 install-devtools:
 	go install golang.org/x/tools/cmd/goimports@latest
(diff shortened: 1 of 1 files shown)


schwern:
What sort of churn? It looks like its doing its job.
https://github.com/momentohq/client-sdk-go/actions/runs/4309710153/jobs/7517407697

eaddingtonwhite:
@schwern we need to be updating examples  with tagged releases and way we have at the moment forces us to make updates to examples as we make changes to sdk. Want to make this lint based off version of sdk in go.mod file in examples folder for which version of sdk it should be linting against.

schwern:
@eaddingtonwhite Doesn't it make more sense for the examples to run off the code in the source directory? The user won't have the examples without the source code.

eaddingtonwhite:
We need to update to use release versions and have Dependabot bump version like we do on other sdk repos. Allows us to iterate a bit in main prior to release and keep examples stable. Also just to be consistent with how we're doing in other sdk repos.

Issue #180, opened 2023-02-24
Spike on byte keys for maps do we need to change up how we do `momento.Bytes()`


eaddingtonwhite:
Currently our dictionaries can only support `map[string]momento.Value` we should spike to see what we can do if there is way to make more flexible so can support byte keys in the map in clean way for users.

Do we need to make any changes to `momento.Bytes()` prior to 1.0 release?

pgautier404:
Addressed in ba3fe975 with the inclusion of an `Element` type that contains `ElemField` and `ElemValue`, both with the `Value` type. Also included are convenience methods to convert from maps with string, []byte, or Value values to `[]Element`.

PR #221, opened 2023-03-02
chore: spike on map field accepted type

poppoerika:
Started with what @pgautier404 had in his experiment branch, `value-map-key` ([commit](https://github.com/momentohq/client-sdk-go/commit/7915ad7ade96638272b460a504feee5dbbc7bc0c)), I extended the spike on the map field param type.
I created a type alias called `Field` which is an alias for our existing `Value` interface type.
For this spike, I attempted to keep the map field to accept `Field` type so that it's clearer for users that they can use both `string` and `byte[]` types.
But it turns out that Go cannot hash interface, which made the following caused panic: 
```
elements := make(map[Field]Value)
elements[r.Field] = r.Value
```

To solve this, I made `Field` property to be a pointer. I'm not sure how I feel about this since one of the few reasons for this to be a pointer is so that it can be used as a map field. 
Another reason is so that users can see what types of fields are availabe.

I'm curious to see what feedback/advice @pgautier404 @schwern @eaddingtonwhite have!

https://github.com/momentohq/client-sdk-go/issues/180

### DIFF
--- a/momento/dictionary_set_fields.go
+++ b/momento/dictionary_set_fields.go
@@ -24,7 +24,7 @@ func (DictionarySetFieldsSuccess) isDictionarySetFieldsResponse() {}
 type DictionarySetFieldsRequest struct {
 	CacheName      string
 	DictionaryName string
-	Elements       map[string]Value
+	Elements       map[Value]Value
 	Ttl            utils.CollectionTtl
 
 	grpcRequest  *pb.XDictionarySetRequest
@@ -34,7 +34,7 @@ type DictionarySetFieldsRequest struct {
 
 func (r *DictionarySetFieldsRequest) cacheName() string { return r.CacheName }
 
-func (r *DictionarySetFieldsRequest) elements() map[string]Value { return r.Elements }
+func (r *DictionarySetFieldsRequest) elements() map[Value]Value { return r.Elements }
 
 func (r *DictionarySetFieldsRequest) ttl() time.Duration { return r.Ttl.Ttl }
 
--- a/momento/dictionary_set_fields.go
+++ b/momento/dictionary_set_fields.go
@@ -24,7 +24,7 @@ func (DictionarySetFieldsSuccess) isDictionarySetFieldsResponse() {}
 type DictionarySetFieldsRequest struct {
 	CacheName      string
 	DictionaryName string
-	Elements       map[Value]Value
+	Elements       map[*Field]Value
 	Ttl            utils.CollectionTtl
 
 	grpcRequest  *pb.XDictionarySetRequest
@@ -34,7 +34,7 @@ type DictionarySetFieldsRequest struct {
 
 func (r *DictionarySetFieldsRequest) cacheName() string { return r.CacheName }
 
-func (r *DictionarySetFieldsRequest) elements() map[Value]Value { return r.Elements }
+func (r *DictionarySetFieldsRequest) elements() map[*Field]Value { return r.Elements }
 
 func (r *DictionarySetFieldsRequest) ttl() time.Duration { return r.Ttl.Ttl }
 
--- a/momento/dictionary_set_field.go
+++ b/momento/dictionary_set_field.go
@@ -19,7 +19,7 @@ func (DictionarySetFieldSuccess) isDictionarySetFieldResponse() {}
 type DictionarySetFieldRequest struct {
 	CacheName      string
 	DictionaryName string
-	Field          Value
+	Field          *Field
 	Value          Value
 	Ttl            utils.CollectionTtl
 }
--- a/momento/dictionary_test.go
+++ b/momento/dictionary_test.go
@@ -77,11 +77,13 @@ var _ = Describe("Dictionary methods", func() {
 				}),
 			).Error().To(HaveMomentoErrorCode(expectedErrorCode))
 
+			var f Field
+			f = String("hi")
 			Expect(
 				sharedContext.Client.DictionarySetField(sharedContext.Ctx, &DictionarySetFieldRequest{
 					CacheName:      cacheName,
 					DictionaryName: dictionaryName,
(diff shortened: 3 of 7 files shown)


momento/dictionary_set_fields.go
```diff
@@ -24,7 +24,7 @@ func (DictionarySetFieldsSuccess) isDictionarySetFieldsResponse() {}
 type DictionarySetFieldsRequest struct {
 	CacheName      string
 	DictionaryName string
-	Elements       map[string]Value
+	Elements       map[*Field]Value
```
pgautier404, line 27: I honestly don't know if it's safe to use pointers as map keys, but it makes me nervous. It also seems like making users pass in this specific type as a pointer where everything else uses values would be inconvenient and potentially confusing. @eaddingtonwhite @schwern wdut?

poppoerika: +1 on pointers for map field seems unsafe. 
It looks to me that Go does not have a concept of accepting multiple types in a clear way. 
What I saw on the Internet was to use `interface{}` as a param but then in our case, it's not clear what types (e.g. `string` ? `byte[]`?) of map fields users can use.

schwern: It makes me nervous, too. And now Fields and Values are passed differently.

1. Would it be a breaking change to do this later?
2. Is there a better way to accomplish this?

For #2 I'll ask on Stack Overflow.

schwern: > you can define Hash() string method and use a map[string]whatever.

🤔 

https://stackoverflow.com/a/45870654/14660

schwern: I can't get it to work. :(

schwern:
Pointers don't really work. The map is doing pointer equality. A copy of the struct will not match.

[Demonstration](https://play.golang.com/p/g_wtzPBJG2h).

momento/dictionary_set_fields.go
schwern: tl;dr [You can't do it](https://stackoverflow.com/questions/29662003/go-map-with-user-defined-key-with-user-defined-equality).

poppoerika:
@schwern @pgautier404 Thank you for the feedback.
As suspected, this proposal won't work... So we know that we cannot do `map[Field]Value` or `map[*Field]Value`.
Would `interface{}` be acceptable to be a map field param?
or how about create a separate function to take byte map field? Would that be more confusing?
cc @eaddingtonwhite

momento/dictionary_set_fields.go
schwern: Longer answers...

[Effective Go: Map Types](https://go.dev/ref/spec#Map_types)
> The [comparison operators](https://go.dev/ref/spec#Comparison_operators) == and != must be fully defined for operands of the key type; thus the key type must not be a function, map, or slice. If the key type is an interface type, these comparison operators must be defined for the dynamic key values; failure will cause a [run-time panic](https://go.dev/ref/spec#Run_time_panics).

Since Byte has a `byte[]` (a slice) we can't use our type in the map. Even if we redefined it as a struct like `type Byte struct { value byte[] }`. Go's idea of equality is very strict and does not allow you to define equality for your own types. 😞 

* [Why are slices in go unhashable?](https://www.reddit.com/r/golang/comments/ecqgha/comment/fbedqp8/?context=3) (byte[] is a slice)
* [Go map with user-defined key with user-defined equality?](https://stackoverflow.com/questions/29662003/go-map-with-user-defined-key-with-user-defined-equality)

schwern:
@poppoerika 

> Would `interface{}` be acceptable to be a map field param?

[`map[interface{}]int` has the same problem](https://play.golang.com/p/ARMI5ADfMsR).

schwern:
If the language doesn't support bytes as map keys, do we need to go crazy to support that?

Do we have a strong use-case for bytes as keys?

poppoerika:
> @poppoerika
> 
> > Would `interface{}` be acceptable to be a map field param?
> 
> [`map[interface{}]string` has the same problem](https://play.golang.com/p/hNWNzY44k8D).

I meant after accepting `interface{}` and if its type is byte we convert it to string before assigning it as a map key.
I [modified your example](https://play.golang.com/p/jZm3lP5rGJ5) and it does what I just mentioned 👍 
But this means that users need to pass string casted byte value to a map key, which is a bit confusing since that's is not the same way as they 
But again as I mentioned on Slack, I think we need a bit more time to come up with a agreed solution and some time to experiment, so we think it's better to not rush this to be included in the 1.0 release.

poppoerika:
@eaddingtonwhite @schwern @pgautier404 @cprice404 
My findings based on the investigation on making map key type be more flexible:

1. Method chaining to make byte available for map key is one option. But I don't believe this is backward compatible, since the return type of our dictionary API needs to be a struct which contains the response and error. This also means that error handling is not straight forward (I think usually Go returns something, err separately not in one struct and depend on users to spread to handle the error).
2. Accepting `interface{}` or `Value {AsString(), AsBytes{}}` as `Elements` or `Field` props for the dictionary APIs' request type is not possible since Go does not allow `interface{}` or `[]byte` as a map key.
3. We can update `Elements` and `Field` props for the dictionary APIs' request type to accept `interface{}` instead of `string` or `Value` and add elaborate documentation to let users know that both `string` and `[]bytes` are accepted to be passed to a map key. And if they pass anything else, they will get `InvalidArgumentError`. As mentioned in [this comment](https://github.com/momentohq/client-sdk-go/pull/221#issuecomment-1453046186), just convert a passed value to string when its type is `[]byte`.

If I missed/misunderstood anything, please let me know!

poppoerika:
Closing this PR since @pgautier404 opened a new PR to address this issue.
https://github.com/momentohq/client-sdk-go/pull/224

PR #222, opened 2023-03-02
chore: move responses into their own package

pgautier404:


responses/list_caches.go
```diff
@@ -0,0 +1,56 @@
...
+
+//func convertCacheInfo(i []CacheInfo) []CacheInfo {
+//	var convertedList []CacheInfo
+//	for _, c := range i {
+//		convertedList = append(convertedList, CacheInfo{
+//			name: c.Name,
+//		})
+//	}
+//	return convertedList
+//}
```
eaddingtonwhite, line 56: @pgautier404 mean to leave this here?

eaddingtonwhite (approved):

responses/list_caches.go
pgautier404: I did not. I'll strip that out in a quick followup.

PR #224, opened 2023-03-03
fix: use alternative data structure for dict ops

pgautier404:
This commit introduces an `Element` type that contains an `ElemField` and `ElemValue`, both of which use our `Value` type. This type is for use with `DictionarySetFields` and allows users to provide byte and string fields and values, as opposed to go's built-in `map` which disallows byte slice keys. It also includes helper functions to convert from a few different types of maps to `[]Element`. Finally, the shared test context code has been modified to prepend test caches with `golang-` for easy identification of the source of stranded caches.

### DIFF
--- a/momento/requester.go
+++ b/momento/requester.go
@@ -65,7 +65,7 @@ type hasFields interface {
 }
 
 type hasElements interface {
-	elements() map[string]Value
+	elements() []Element
 }
 
 type hasTtl interface {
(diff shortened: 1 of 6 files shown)


cprice404 (dismissed):
i think i like this direction, personally!  we could also provide a utility function that converts from map[string]Value to []Element.

poppoerika (commented):
I like this approach as well! 
Also I don't think we need much education is need for Go users since I'd assume they're used to a struct param.

But since this is a breaking change, we need to include in 1.0 release though?

pgautier404:
> But since this is a breaking change, we need to include in 1.0 release though?

Yes, we're definitely going to want to get this in for 1.0.

momento/value.go
```diff
@@ -30,3 +30,7 @@ func (v String) asBytes() []byte {
 func (v String) asString() string {
 	return string(v)
 }
+
+type Element struct {
+	ElemField, ElemValue Value
+}
```
poppoerika, line 36: Maybe add doc here to describe what `Element` is?
Something like `Element` is a Momento custom map type which accepts flexible types for map key and value or something like that?

momento/requester.go
```diff
@@ -235,3 +233,36 @@ func validateNotEmpty(field []byte, label string) error {
 	}
 	return nil
 }
+
+func ElementsFromMapStringString(theMap map[string]string) []Element {
```
poppoerika, line 237: Are these functions customer facing so that users can easily create an `[]Element` value?

poppoerika: These means `ElementsFromMapStringString`, `ElementsFromMapStringBytes`, and `ElementsFromMapStringValue`

pgautier404: Yes, exactly.

momento/value.go
pgautier404: Good call, I'll add some docs.

pgautier404: Fixed in 036fedb, thanks for the suggestion!

momento/requester.go
poppoerika: Awesome. 🍾

momento/value.go
poppoerika: Thank you!

poppoerika (approved):
Looks good to me!
Thank you for doing this, Pete.

PR #228, opened 2023-03-03
fix: missed refactoring get responses somehow

pgautier404:
I somehow managed to miss the `get` operation 😅


poppoerika (approved):
I should have caught that and have noticed that when I didn't need to update the scaler example 😮‍💨 
Thank you for catching it.

PR #230, opened 2023-03-03
fix: fix misleading info in example

pgautier404:


poppoerika (approved):
Great catch! Thank you 🙇‍♀️

PR #232, opened 2023-03-06
chore: add `WARN`, `ERROR`  and `TRACE` log levels

### LINKED ISSUE #157: Add standard debug/warn/error logging for user
* Make sure to include warning logging when receiving unknown response type

poppoerika:
- Added `WARN`, `ERROR`  and `TRACE` log levels
- Updated `MomentoLoggerFactory`'s `GetLogger` to take `logLevel` to log correctly based on a requested log level
- Updated `MomentoLogger` functions to log based on a requested log level
- Added `TRACE` logging when detecting an unrecognized response besides `Discontinuity`, `Item`, and `Heartbeat` for topic subscription

Also
- Rename `simple_cache_client.go` to `cache_client.go`.

Closes #157


config/logger/builtin-logger.go
```diff
@@ -6,42 +6,77 @@ import (
...
 
 func (l *BuiltinMomentoLogger) Info(message string, args ...string) {
-	if len(args) != 0 {
-		log.Printf("[%s] %s (%s): %s, %s\n", time.RFC3339, l.level, l.loggerName, message, strings.Join(args, ", "))
-	} else {
-		log.Printf("[%s] %s (%s): %s\n", time.RFC3339, l.level, l.loggerName, message)
+	if l.level == INFO {
+		if len(args) != 0 {
+			logWithArgs(l.level, l.loggerName, message, args...)
+		} else {
```
pgautier404, line 18: You can save a lot of code by just using one `log` method and checking `len(args)` there. There's no need for `logWith` and `logWithout` because variadic args are not required (which is how they can have a zero length).

poppoerika: 💡

config/logger/builtin-logger.go
```diff
@@ -6,31 +6,46 @@ import (
...
+	if l.level == DEBUG {
+		momentoLog(l.level, l.loggerName, message, args...)
+	}
+}
+
+func (l *BuiltinMomentoLogger) Warn(message string, args ...string) {
+	if l.level == WARN {
+		momentoLog(l.level, l.loggerName, message, args...)
+	}
+}
```
schwern, line 30: This means a warning will not show up if the log level is set to a more detailed log level. For example, if the level is TRACE we'd log TRACE, INFO, WARN and ERROR.

This is easier if the log levels are defined as numbers in order of severity. Then it's `if l.level >= WARN`.

config/logger/builtin-logger.go
```diff
@@ -6,31 +6,46 @@ import (
...
+
+func momentoLog(level loggerLevel, loggerName string, message string, args ...string) {
+	if len(args) > 0 {
+		log.Printf("[%s] %s (%s): %s, %s\n", time.RFC3339, level, loggerName, message, strings.Join(args, ", "))
 	} else {
-		log.Printf("[%s] %s (%s): %s\n", time.RFC3339, l.level, l.loggerName, message)
+		log.Printf("[%s] %s (%s): %s\n", time.RFC3339, level, loggerName, message)
 	}
 }
```
schwern, line 50: Go does have a [log package](https://pkg.go.dev/log). There might be an advantage to build on top of that.

momento/topic_subscription.go
```diff
@@ -21,6 +23,7 @@ type topicSubscription struct {
 	momentoTopicClient *pubSubClient
 	cacheName          string
 	topicName          string
+	log                logger.MomentoLogger
```
schwern, line 26: logger?

schwern (changes requested):

schwern (approved):

PR #239, opened 2023-03-06
feat!: Sortedset inconsistencies

### LINKED ISSUE #167: Sorted set inconsistencies.

### LINKED ISSUE #168: SortedSet Experience Polish
Now that we have first pass done for sorted sets we should come back and take another pass as we write tests and see if any improvements we can make to user experience make sure no gaps in API prior to 1.0 release.

### LINKED ISSUE #204: SortedSetIncrement / DictionaryIncrement are inconsistent.
* DictionaryIncrementSuccess has a Value method.
* SortedSetIncrementSuccess has a Value member.

- SortedSetIncrementSuccess...
  - [x] Add a Score method.
  - [x] Make the Value member private.

schwern:
* Everything takes Value or Values
  * Not ElementValue, there's nothing special about sorted set values.
* SortedSetRemove only takes values.
  * The some/all flag was complex, use `client.delete`.
  * Use the usual value validation.
* SortedSetIncrementScoreSuccess
  * Add Score()
  * Remove public member.
  * Make it a type conversion.
* SortedSetGetRankHit
  * Add Rank()
  * Remove public member.
  * Make it a type conversion.

Closes #167 
Closes #204 
Closes #168

### DIFF
--- a/momento/requester.go
+++ b/momento/requester.go
@@ -93,7 +93,7 @@ func prepareName(name string, label string) (string, error) {
 func prepareElementValue(value Value) ([]byte, error) {
 	if value == nil {
 		return nil, buildError(
-			momentoerrors.InvalidArgumentError, "element value cannot be nil or empty", nil,
+			momentoerrors.InvalidArgumentError, "element value cannot be nil", nil,
 		)
 	}
 
--- a/momento/requester.go
+++ b/momento/requester.go
@@ -90,21 +90,20 @@ func prepareName(name string, label string) (string, error) {
 	return name, nil
 }
 
-func prepareElementValue(name Value) ([]byte, error) {
-	if name == nil {
+func prepareElementValue(value Value) ([]byte, error) {
+	if value == nil {
 		return nil, buildError(
 			momentoerrors.InvalidArgumentError, "element value cannot be nil or empty", nil,
 		)
 	}
 
 	// just validate not empty using prepareName
-	nameBytes := name.asBytes()
-	_, err := prepareName(string(nameBytes), "element value")
+	_, err := prepareName(value.asString(), "element value")
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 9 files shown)


poppoerika (dismissed):
Looks good to me!

momento/requester.go
```diff
@@ -90,21 +90,20 @@ func prepareName(name string, label string) (string, error) {
 	return name, nil
 }
 
-func prepareElementValue(name Value) ([]byte, error) {
-	if name == nil {
+func prepareElementValue(value Value) ([]byte, error) {
+	if value == nil {
```
pgautier404, line 94: I think the error below should say "element value cannot be nil" instead?

pgautier404:
LGTM apart from the question about the error message.

pgautier404 (approved):
LGTM!

PR #241, opened 2023-03-07
feat!: Complete SortedSetFetch

### LINKED ISSUE #124: Align default arguments for sorted set fetch
According to our api docs, `order` and `limit` should both be required arguments as part of the fetch request https://www.notion.so/momentohq/SDK-API-Methods-Spec-cf12e4db2535405ca2acc3608814b083#7e9ed5fca577491fb73e38161752e92b. However, the following go code compiles
```
client.SortedSetFetch(ctx, &momento.SortedSetFetchRequest{
        cacheName: "test",
        setName: "set",
})

client.SortedSetFetch(ctx, &momento.SortedSetFetchRequest{})
```

Its not clear what the default ordering is or what the default limit is in this code. We should change our function signatures to ensure that all required variables are set, else have it be a compile time failure.

### LINKED ISSUE #169: Coordinate Sorted Fetch Breaking Change release
We are making a breaking change to Sorted Set Fetch API prior to the 1.0 release
https://github.com/momentohq/client-protos/pull/125

This ticket is to track updating SDK to conform to this and to update examples and blogs referencing sorted sets if needed

schwern:
Splitting It into two methods lead to some pretty big design compromises. I decided for Go it's easier and idiomatic to use two different structs for arguments. It accomplishes the same thing; disambiguating the two types of fetching.

* SortedSetFetchHit
  * Add ValueStringElements and ValueByteElements
  * Struct members are private
* SortedSetFetch
  * Add ByIndex and ByScore parameters.
  * Pass by reference to support optional args.

Closes #124 
Closes #169

### DIFF
--- a/momento/cache_client.go
+++ b/momento/cache_client.go
@@ -25,6 +25,7 @@ type CacheClient interface {
 	Delete(ctx context.Context, r *DeleteRequest) (responses.DeleteResponse, error)
 
 	SortedSetFetchByIndex(ctx context.Context, r *SortedSetFetchByIndexRequest) (responses.SortedSetFetchResponse, error)
+	SortedSetFetchByScore(ctx context.Context, r *SortedSetFetchByScoreRequest) (responses.SortedSetFetchResponse, error)
 	SortedSetPut(ctx context.Context, r *SortedSetPutRequest) (responses.SortedSetPutResponse, error)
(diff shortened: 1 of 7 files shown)


pgautier404 (approved):
This looks good to me. It's a departure from some of the other APIs due to its complexity, but it looks to me like everything you've done here is necessary and correct. I would like to hear from @eaddingtonwhite and/or @cprice404 on the UX, but I'm 👍.

PR #243, opened 2023-03-07
feat: add key exists

pgautier404:


momento/requester.go
```diff
@@ -122,6 +125,17 @@ func prepareKey(r hasKey) ([]byte, error) {
 	return key, nil
 }
 
+func prepareKeys(r hasKeys) ([][]byte, error) {
+	var keys [][]byte
+	for _, key := range r.keys() {
+		if key == nil || len(key.asBytes()) == 0 {
+			return nil, buildError(momentoerrors.InvalidArgumentError, "key cannot be nil or empty", nil)
+		}
```
schwern, line 133:
```suggestion
		if err := validateNotEmpty(key, "key"); err != nil {
			return nil, err
		}
```

pgautier404: Good call, thanks. Updated in 974695c.

PR #244, opened 2023-03-07
fix: fix version requirements in docs and go.mod

pgautier404:
This commit updates documented minimum language requirements by removing the mention of version minimum in the main README and specifying the minimum as 1.18 in the CONTRIBUTING doc. For contributing developers, our use of go workspaces and the ginkgo test framework already have us dependent on 1.18+ for contributing and testing code. The minimum version declaration in the main SDK `go.mod` has been changed from 1.19 to 1.18 and the declaration has been removed altogether in the `examples/go.mod` file.

Note that the version declaration in the SDK `go.mod` does not prevent users from developing against the SDK in versions below the one we specify. For example, I was able to start a project from scratch, download the SDK, copy over the usage example, and execute the example code all using a 1.16 go binary.


eaddingtonwhite:
@pgautier404 do we need to update here also?
https://github.com/momentohq/client-sdk-go/blob/main/examples/README.md?plain=1#L5

schwern:
> This commit updates documented minimum language requirements by removing the mention of version minimum in the main README.

Why remove the min version from the README? Won't users will want to know that.

pgautier404:
> > This commit updates documented minimum language requirements by removing the mention of version minimum in the main README.
> 
> Why remove the min version from the README? Won't users want to know that?

Mostly because I don't know what the minimum version would be for just coding against the SDK module. I tried with 1.16 without a problem, but I don't want to declare that as the minimum as there may be previous versions that also work. That said, I doubt many folks would be using anything much earlier. All that considered, I don't really see a point in declaring a minimum version or spending lots more time trying old go versions to see if and where we bottom out.

eaddingtonwhite (approved):

PR #247, opened 2023-03-08
fix: All values can be blank.

### LINKED ISSUE #226: Some methods return a `""` value as nil.
`""` comes back as `[]byte("")` which is `[]uint8 len: 0, cap: 0, nil`. Some methods are turning this into `nil`. I haven't checked them all, but so far...

-DictionaryFetch

There's no reason to disallow an empty string value. I think it would be an easy fix.

Alternatively, we could temporarily disallow empty values.

schwern:
The original issue with DictionaryFetch returning `""` as `nil` seems to be gone. 🤔 

* Validate all values the same: can't be nil, can be empty.
* Add tests for empty values.

Closes #226


pgautier404 (approved):
LGTM!

PR #248, opened 2023-03-08
feat!: Normalize dictionaries.

schwern:
* Element -> DictionaryElement
  * Distinguish it from every other kind of "element".
* ElemField/ElemValue -> Field/Value
  * We no longer prefix struct members. A field is a field, a value is a value.

I think this is the last naming issue.

### DIFF
--- a/momento/dictionary_set_fields.go
+++ b/momento/dictionary_set_fields.go
@@ -13,7 +13,7 @@ import (
 type DictionarySetFieldsRequest struct {
 	CacheName      string
 	DictionaryName string
-	Elements       []Element
+	Elements       []DictionaryElement
 	Ttl            *utils.CollectionTtl
 
 	grpcRequest  *pb.XDictionarySetRequest
@@ -23,7 +23,7 @@ type DictionarySetFieldsRequest struct {
 
 func (r *DictionarySetFieldsRequest) cacheName() string { return r.CacheName }
 
-func (r *DictionarySetFieldsRequest) elements() []Element { return r.Elements }
+func (r *DictionarySetFieldsRequest) dictionaryElements() []DictionaryElement { return r.Elements }
 
 func (r *DictionarySetFieldsRequest) ttl() time.Duration { return r.Ttl.Ttl }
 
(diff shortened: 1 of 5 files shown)


momento/value.go
```diff
@@ -32,8 +32,8 @@ func (v String) asString() string {
 }
 
 // Element Type to hold field/value pairs for use with the DictionarySetFields operation.
```
eaddingtonwhite, line 34: We should update first word in func doc to match struct name. So like `// DictionaryElement Type to hold field/value pairs....`

eaddingtonwhite (approved):
This looks good to me. @pgautier404 @cprice404 take a look as well make sure good w/ naming here for `DcitionaryElement` makes sense to me though since we handle this differently the `Elements` in other data structures. I cant think of better name then `DictionaryElement` at the moment.

momento/requester.go
```diff
@@ -170,18 +170,18 @@ func prepareValues(r hasValues) ([][]byte, momentoerrors.MomentoSvcErr) {
 	return values, nil
 }
 
-func prepareElements(r hasElements) ([]Element, error) {
-	for _, v := range r.elements() {
-		if v.ElemValue == nil || v.ElemField == nil {
+func prepareDictinoaryElements(r hasDictionaryElements) ([]DictionaryElement, error) {
```
pgautier404, line 173:
```suggestion
func prepareDictionaryElements(r hasDictionaryElements) ([]DictionaryElement, error) {
```

momento/dictionary_set_fields.go
```diff
@@ -38,16 +38,16 @@ func (r *DictionarySetFieldsRequest) initGrpcRequest(client scsDataClient) error
 		return err
 	}
 
-	var elements []Element
-	if elements, err = prepareElements(r); err != nil {
+	var elements []DictionaryElement
+	if elements, err = prepareDictinoaryElements(r); err != nil {
```
pgautier404, line 42:
```suggestion
	if elements, err = prepareDictionaryElements(r); err != nil {
```

pgautier404:
I commented on a function name misspelling that needs to be fixed, but apart from that LGTM!

pgautier404 (approved):
LGTM!

PR #253, opened 2023-03-09
fix: export LogLevel

pgautier404:
This commit exports the `LogLevel` const from `momento-logger` so users can access it.


poppoerika (approved):
Thank you for catching it. Looks good to me!

cprice404 (approved):

PR #255, opened 2023-03-09
docs: macos contributing note about go path

malandis:
On MacOS, two developers had to add the mentioned path to the PATH
environment variable so go could find the dev tools.


cprice404 (approved):

poppoerika (approved):
Thank you!

PR #258, opened 2023-03-10
refactor: rename `SortedSetPut` to `SortedSetPutElements`

malandis:
Adjusts request and response types appropriately.


cprice404 (approved):
this matches node.js so it looks right to me, but will defer to @poppoerika and @poppoerika to make sure that's how we're naming stuff for list requests etc.

examples/sortedset-example/main.go
```diff
@@ -28,7 +28,7 @@ func main() {
 	// Put score for each element to set
 	// Using counter, element N has score N
 	for i := 1; i < 11; i++ {
-		_, err := client.SortedSetPut(ctx, &momento.SortedSetPutRequest{
+		_, err := client.SortedSetPut(ctx, &momento.SortedSetPutElementsRequest{
```
poppoerika, line 31: Do we want to call `client.SortedSetPutElements` here?

pgautier404:
> this matches node.js so it looks right to me, but will defer to @poppoerika and @poppoerika to make sure that's how we're naming stuff for list requests etc.

This is in line with the API spec and other SDKs. As far as lists go, the equivalent operations would be `ListPushFront/Back` so no mention of `Element`s. We do have `ListRemoveValue` as opposed to `SetRemoveElement`, but that is consistent with the spec and other SDKs. This LGTM.

pgautier404 (approved):

examples/sortedset-example/main.go
malandis: @poppoerika yes, but that will depend on the release version of the client, right?

malandis: Good eye. I didn't mean to update the examples.

pgautier404 (approved):

poppoerika (approved):

PR #263, opened 2023-03-10
fix: fixes to some minor misc issues

pgautier404:


momento/sorted_set_increment_score.go
```diff
@@ -27,7 +27,7 @@ type SortedSetIncrementScoreRequest struct {
 
 func (r *SortedSetIncrementScoreRequest) cacheName() string { return r.CacheName }
 
-func (r *SortedSetIncrementScoreRequest) requestName() string { return "Sorted set increment" }
+func (r *SortedSetIncrementScoreRequest) requestName() string { return "Sorted set increment score" }
```
cprice404, line 30: nit; some of the ones above use camel case here but these have spaces.  probably standardizing on camel is best.  not a blocker for 1.0.

cprice404 (approved):

momento/sorted_set_increment_score.go
pgautier404: Yeah, that's a sorted set thing. I looked into where these are being used and they're informational only, so fine to fix when we get to it.

PR #265, opened 2023-03-10
feat: add logrus logging example

cprice404:
This commit adds an example that shows how to implement the Momento logging interface using the logrus logging framework, and some example code illustrating that the Momento log messages are successfully routed to logrus.


cprice404:
This can't be merged until we have merged the corresponding logging changes from #264 , and done a release that this can point to.

examples/go.mod
```diff
@@ -1,17 +1,224 @@
...
 require (
 	github.com/google/uuid v1.3.0
-	github.com/momentohq/client-sdk-go v0.11.0
+	github.com/momentohq/client-sdk-go v0.13.0
+	// logrus is not required to use momento, but it is used in the logging-example
+	github.com/sirupsen/logrus v1.9.0
 )
 
 require (
+	cloud.google.com/go v0.105.0 // indirect
```
cprice404, line 15: all the additions in here seem really surprising and wrong to me.  I have no idea what caused them.  Before I merge this PR I will try to restore them back to the previous form by hand and see if everything still builds, but if any of you all have ideas on how exactly I might have screwed this up please lmk :)

examples/go.mod
```diff
@@ -1,17 +1,224 @@
 module github.com/momentohq/go-example
 
+go 1.19
+
+replace github.com/momento/client-sdk-go => /Users/cprice/git/momento/client-sdk-go
```
pgautier404, line 5: These shouldn't be committed.

cprice404: yup, good call, thanks

cprice404: should be fixed now

pgautier404 (approved):
LGTM!

PR #269, opened 2023-03-10
chore: add sorted set get score

pgautier404:


momento/cache_client.go
```diff
@@ -260,6 +260,26 @@ func (c defaultScsClient) SortedSetGetScores(ctx context.Context, r *SortedSetGe
 	return r.response, nil
 }
 
+func (c defaultScsClient) SortedSetGetScore(ctx context.Context, r *SortedSetGetScoreRequest) (responses.SortedSetGetScoreResponse, error) {
```
poppoerika, line 263: Do we not need to add this to `CacheClient` interface?

pgautier404: We sure do, thanks. Fixed in 2fa705f

poppoerika (approved):
Looks good to me!

malandis (approved):

PR #272, opened 2023-03-11
chore: add testing for newly added methods

pgautier404:


malandis (approved):
Let's make note to add tests for...

- put element: on an sorted set that doesn't exist, put overwrites, put bytes
- get score: on a sorted set that doesn't exist; on a value that doesn't exist
- same for get scores

pgautier404:
> Let's make note to add tests for...
> 
>     * put element: on an sorted set that doesn't exist, put overwrites, put bytes
> 
>     * get score: on a sorted set that doesn't exist; on a value that doesn't exist
> 
>     * same for get scores

Totally, I'll add a ticket, thanks!

PR #276, opened 2023-03-12
fix: simplify abstractions involved in sorted set get score

pgautier404:


responses/sorted_set_get_scores.go
```diff
@@ -12,37 +12,46 @@ func (SortedSetGetScoresMiss) isSortedSetGetScoresResponse() {}
...
+func (r SortedSetGetScoresHit) Responses() []SortedSetGetScoreResponse {
+	return r.responses
 }
 
-// SortedSetGetScore is the base response for individual scores.
-type SortedSetGetScore interface {
-	isSortedSetScoreElement()
+// ScoresFloat returns an array of float64 values that represent the hit responses. Misses
+// are not represented the array.
+func (r SortedSetGetScoresHit) ScoresFloat() []float64 {
```
cprice404, line 34: should this be called `ScoresArray`?  the scores are always floats, the difference between this accessor and the other one is the collection that you return them in.

cprice404 (approved):

PR #280, opened 2023-03-13
feat: Split sortedSetFetch up into ByRank and ByScore functions

cprice404:
As per the API design discussion and the pattern we established in the node.js SDK, this commit splits the SortedSetFetch into two functions, one for fetching by score and one for by rank. This prevents the user from needing to reason about which combinations of arguments are and are not legal together.


momento/cache_client.go
```diff
@@ -34,9 +34,11 @@ type CacheClient interface {
 	// KeysExist checks if provided keys exist in the cache.
 	KeysExist(ctx context.Context, r *KeysExistRequest) (responses.KeysExistResponse, error)
 
-	// SortedSetFetch fetches the elements in the given sorted set by index rank or by score.
-	SortedSetFetch(ctx context.Context, r *SortedSetFetchRequest) (responses.SortedSetFetchResponse, error)
-	// SortedSetElement adds an element to the given sorted set. If the element already exists,
+	// SortedSetFetchByRank fetches the elements in the given sorted set by rank.
+	SortedSetFetchByRank(ctx context.Context, r *SortedSetFetchByRankRequest) (responses.SortedSetFetchResponse, error)
+	// SortedSetFetchByScore fetches the elements in the given sorted set by rank.
```
poppoerika, line 39: `by rank` -> `by score`?

momento/sorted_set_fetch_by_rank.go
```diff
@@ -0,0 +1,100 @@
...
+
+	pb "github.com/momentohq/client-sdk-go/internal/protos"
+)
+
+type SortedSetFetchByRankRequest struct {
+	CacheName string
+	SetName   string
+	Order     SortedSetOrder
+	StartRank *int32
+	EndRank   *int32
```
poppoerika, line 16: Sorry maybe this has been discussed somewhere, but why do we use pointers for `StartRank` and `EndRank`?

momento/sorted_set_fetch_by_score.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type SortedSetFetchByScoreRequest struct {
+	CacheName string
+	SetName   string
+	Order     SortedSetOrder
+	MinScore  *float64
+	MaxScore  *float64
+	Offset    *uint32
+	Count     *uint32
```
poppoerika, line 18: Same comment as above.
Why do we use pointers here?

momento/sorted_set_fetch_by_rank.go
cprice404: I think it's because we want to be able to use `nil` to express that we aren't specifying them.

momento/cache_client.go
cprice404: good catch.

momento/sorted_set_fetch_by_rank.go
poppoerika: I know that that's how we did for `SortedSetFetchByRank` and `SortedSetFetchByScore` types, but I'm not sure why these were pointers.

momento/cache_client.go
cprice404, line 39:
```suggestion
	// SortedSetFetchByScore fetches the elements in the given sorted set by score.
```

cprice404: https://github.com/momentohq/client-sdk-go/pull/280/commits/a3967fc677e701a871f2f801d818a117f1fa005c

momento/sorted_set_fetch_by_rank.go
cprice404: if they were not pointers, how would I express that I want to use the defaults?

You may be right that there is another way to do it!  But that was my understanding.

poppoerika: Oh I see. It makes sense pointers were used so that they can be `nil` since these can be "optional."

I think it makes sense for those values to be optional since `ByScore` and `ByRank` were a part of the fetch API.
But now we're breaking into two separate APIs specifically named `ByScore` and `ByRank`, it feels a bit weird for those values to be optional?
What do you think? @cprice404

momento/cache_client.go
poppoerika: 🥳

momento/sorted_set_fetch_by_rank.go
poppoerika: But it looks like it's already decided in the API doc.

cprice404: They are optional because you could pass either one of them or the other, or both.

For example, you could pass only `EndRank` of `10` if you wanted to get the first 10.  Or you could pass only `StartRank` of 10 and then get all of the elements after the 10th one.

Also note that these numeric fields were pointers even inside of the ByRank/ByScore objects before.

poppoerika: It makes sense! Thank you for the clarification.

Yup, I saw that after leaving the above comment and figured that this is something decided with reasons.

poppoerika (approved):
Looks good to me!

PR #282, opened 2023-03-14
chore: update examples to v0.16.0

cprice404:


poppoerika (approved):
Looks good to me!

Two things:
- We need to make a follow up PR to update the SDK version in the examples to `1.0.0`?
- Do we want to add a step to run the examples in our GH workflow later to make sure these are properly updated?

cprice404:
> Looks good to me!
> 
> Two things:
> 
>     * We need to make a follow up PR to update the SDK version in the examples to `1.0.0`?

Yep!

> 
>     * Do we want to add a step to run the examples in our GH workflow later to make sure these are properly updated?

If we don't have that already then yes we should add that.  We are supposed to have a dependabot thing that will detect when new versions become available and then file an automatic PR to update the version used by the examples, and will detect if there is a build failure, but I haven't looked to see if we have that set up in this repo.  I'll create a ticket to capture that and you can scout it out.

PR #285, opened 2023-03-14
chore: Update main.go

NoSQLKnowHow:
Added comment before main section so people know why the cacheClient object is used.

### DIFF
--- a/examples/pubsub-example/main.go
+++ b/examples/pubsub-example/main.go
@@ -15,6 +15,9 @@ const (
 	topicName = "test-topic"
 )
 
+// If you were creating your own app, one wouldn't need the cacheClient object, setupCache function, or associated code if you were connecting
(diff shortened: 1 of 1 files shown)


examples/pubsub-example/main.go
```diff
@@ -15,6 +15,9 @@ const (
 	topicName = "test-topic"
 )
 
+// If you were creating your own app, one wouldn't need the cacheClient object, setupCache function, or associated code if you were connecting
+// to an existing cache in Momento Serverless Cache. For demonstration purposes, this example creates a cache for the new Topic to show the publish
+// and subscribe functionality.
```
pgautier404, line 20:
```suggestion
// You wouldn't need the cacheClient object, setupCache function, or associated code if you were connecting to an existing cache in Momento Serverless Cache.
//  For demonstration purposes, this example creates a cache for the new Topic to show the publish and subscribe functionality.
```

This feels a little more direct to me, but feel free to take it or leave it.

PR #288, opened 2023-03-15
chore: sorted set testing followup

pgautier404:
This commit adds testing for sorted set functionality as well as validation for sorted set "by rank" indexes. It also validates values in `SortedSetPutElements` and errors on nils.

### DIFF
--- a/momento/requester.go
+++ b/momento/requester.go
@@ -239,3 +239,17 @@ func validateNotNil(value Value, label string) error {
 
 	return nil
 }
+
+func validateSortedSetRanks(start int32, end int32) error {
+	if start >= 0 && end >= 0 && start >= end {
+		return buildError(
+			momentoerrors.InvalidArgumentError, "start rank must be less than end rank", nil,
+		)
+	}
+	if start < 0 && end < 0 && start >= end {
+		return buildError(
+			momentoerrors.InvalidArgumentError, "negative start rank must be less than negative end rank", nil,
+		)
+	}
+	return nil
+}
--- a/momento/sorted_set_test.go
+++ b/momento/sorted_set_test.go
@@ -406,23 +406,29 @@ var _ = Describe("SortedSet", func() {
 				))
 			})
 
-			It("returns an error for detectable invalid ranges", func() {
-				start := int32(10)
-				end := int32(3)
-				Expect(
-					sharedContext.Client.SortedSetFetchByRank(
-						sharedContext.Ctx,
-						&SortedSetFetchByRankRequest{
-							CacheName: sharedContext.CacheName,
-							SetName:   sharedContext.CollectionName,
-							StartRank: &start,
-							EndRank:   &end,
-						},
-					),
-				).Error().To(HaveMomentoErrorCode(InvalidArgumentError))
+			DescribeTable("we get an error for detectable invalid ranges",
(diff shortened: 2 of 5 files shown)


poppoerika:
No related to this specific PR but it looks like we get the `LimitExceededError`.
I thought we fixed how we ran the tests so that used caches are deleted at the end of the full test run...?

momento/requester.go
```diff
@@ -239,3 +239,17 @@ func validateNotNil(value Value, label string) error {
...
+		return buildError(
+			momentoerrors.InvalidArgumentError, "start rank must be less than end rank", nil,
+		)
+	}
+	if start < 0 && end < 0 && start >= end {
+		return buildError(
+			momentoerrors.InvalidArgumentError, "negative start rank must be less than negative end rank", nil,
+		)
+	}
+	return nil
```
cprice404, line 254: where did you land on this validation?  does it make sense to you or do you still have concerns about it being confusing?

momento/sorted_set_test.go
```diff
@@ -346,6 +364,27 @@ var _ = Describe("SortedSet", func() {
...
+						Order:     DESCENDING,
+						StartRank: &start,
+						EndRank:   &end,
+					},
+				)
+				Expect(err).To(BeNil())
+				switch fetchResp := fetchResp.(type) {
+				case *SortedSetFetchHit:
+					Expect(fetchResp.ValueBytesElements()).To(Equal([]SortedSetBytesElement{}))
+					Expect(fetchResp.ValueStringElements()).To(Equal([]SortedSetStringElement{}))
```
cprice404, line 384: you mentioned in slack that you were seeing some cases where the server or the SDK was returning `nil` vs empty list?  Here your test is asserting empty list, which is what I would expect, but if there are cases where we are returning `nil` I'd like to learn more about them.

cprice404 (commented):
thanks for doing this, just a few questions

momento/requester.go
pgautier404: It makes sense, but it seems like a half measure, like we're putting up handrails on one side of a staircase but not the other. That said, I think those handrails are better than no handrails, so I went ahead and opted for consistency with the nodejs sdk.

momento/sorted_set_test.go
pgautier404: The previous implementation was passing the `nil` we got from the server straight through to the user. I'm now checking for the `nil` from the server and converting it to an empty list in the response in the `NewSortedSetFetchHit` constructor instead.

pgautier404:
> No related to this specific PR but it looks like we get the `LimitExceededError`. I thought we fixed how we ran the tests so that used caches are deleted at the end of the full test run...?

Apparently there are still cases where the caches aren't getting deleted.

momento/sorted_set_test.go
cprice404: great, that seems better to me.

momento/requester.go
cprice404: okay.  that is basically the same conclusion that i came to.  I wouldn't have a problem changing this in both sdks in the next few days if we come up with a better idea, but I don't have one atm :)

cprice404 (approved):

PR #290, opened 2023-03-24
chore: version fix

pgautier404:


examples/go.mod
```diff
@@ -1,13 +1,14 @@
 module github.com/momentohq/go-example
 
-require (
-	github.com/google/uuid v1.3.0
-	github.com/momentohq/client-sdk-go v0.16.0
+go 1.20
```
poppoerika, line 3: It looks like we're using `go 1.18` in our main go.mod.
Do we want to align with that?

pgautier404: Good catch, that shouldn't be there at all for examples. I'll remove it.

poppoerika (approved):

PR #291, opened 2023-04-03
feat: add support for default cache name on client creation

pgautier404:


momento/cache_client.go
```diff
@@ -188,14 +184,43 @@ func NewCacheClient(configuration config.Configuration, credentialProvider auth.
...
+func NewCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration) (CacheClient, error) {
+	props := CacheClientProps{
+		Configuration:      configuration,
+		CredentialProvider: credentialProvider,
+		DefaultTtl:         defaultTtl,
+	}
+	return commonCacheClient(props)
+}
+
+func NewSingleCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration, cacheName string) (CacheClient, error) {
```
pgautier404, line 205: I am very open to naming suggestions for this constructor.

momento/cache_client.go
```diff
@@ -188,14 +184,43 @@ func NewCacheClient(configuration config.Configuration, credentialProvider auth.
...
+	props := CacheClientProps{
+		CacheName:          cacheName,
+		Configuration:      configuration,
+		CredentialProvider: credentialProvider,
+		DefaultTtl:         defaultTtl,
+	}
+	return commonCacheClient(props)
+}
+
+func (c defaultScsClient) setRequestCache(request hasCacheName) string {
```
cprice404, line 215: super nit - the name of this function implies that you are mutating something but you're not.  Maybe `getCacheNameForRequest`?

momento/delete_cache.go
```diff
@@ -4,3 +4,7 @@ type DeleteCacheRequest struct {
 	// string cache name to delete.
 	CacheName string
 }
+
+func (c DeleteCacheRequest) cacheName() string {
```
cprice404, line 8: why do you need these?  Can't you just reference the `CacheName` property directly?

momento/cache_client.go
cprice404: I'd probably go for `NewCacheClientWithDefaultCache` - @eaddingtonwhite ?

cprice404 (commented):
a couple minor questions / nits but otherwise lgtm

momento/cache_client.go
eaddingtonwhite: Yeah I think I agree especially since they could still overide cache name per request. So having "WithDefaultCache" in name I think makes more sense to me.

momento/delete_cache.go
pgautier404: These were added where necessary so the requests implemented the interface required for the `getRequestCache()` method.

momento/cache_client.go
pgautier404: Sounds good. Fixed in 1046215.

momento/cache_client.go
pgautier404: Good call. Fixed in 1046215.

pgautier404: *Actually* fixed in e1b74aa :-)

momento/cache_client.go
```diff
@@ -188,14 +184,43 @@ func NewCacheClient(configuration config.Configuration, credentialProvider auth.
...
+	props := CacheClientProps{
+		CacheName:          cacheName,
+		Configuration:      configuration,
+		CredentialProvider: credentialProvider,
+		DefaultTtl:         defaultTtl,
+	}
+	return commonCacheClient(props)
+}
+
+func (c defaultScsClient) getCacheNameForRequest(request hasCacheName) string {
```
poppoerika, line 215: When users call this function with `NewCacheClient`, it returns `defaultCache`, right? Where are we setting this default cache?

poppoerika: And for the Go SDK, this means that a cache is always created with both `NewCacheClient` and `NewCacheClientWithDefaultCache` even if users didn't provide a cache name when calling `CreateCache`?

momento/delete_cache.go
pgautier404: Yep.

momento/cache_client.go
pgautier404: The default cache is set in the `NewCacheClientWithDefaultCache` constructor, which returns a `CacheClient` just like the `NewCacheClient` constructor. If the client was created using `NewCacheClient`, it will not have a default cache set and the user has to provide a cache name as part of each request. If the client was created with `NewCacheClientWithDefaultCache`, it will have a default cache and the user can omit the cache name from the requests.

poppoerika (approved):
Looks good to me!

PR #293, opened 2023-04-04
feat: add batch delete util

pgautier404:


batchutils/batch_operations.go
```diff
@@ -0,0 +1,98 @@
...
+		myKey := <-keyChan
+		if myKey == nil {
+			return
+		}
+		_, err := client.Delete(ctx, &momento.DeleteRequest{
+			CacheName: cacheName,
+			Key:       myKey,
+		})
+		if err != nil {
+			errChan <- fmt.Sprintf("error deleting key %s: %s", myKey, err.Error())
```
cprice404, line 45: just out of curiosity - is there a reason you decided to just push strings onto this channel rather than the actual errors?

batchutils/batch_operations.go
```diff
@@ -0,0 +1,98 @@
...
+
+type BatchDeleteRequest struct {
+	Client               momento.CacheClient
+	CacheName            string
+	Keys                 []momento.Key
+	MaxConcurrentDeletes int
+}
+
+// BatchDelete deletes a slice of keys from the cache, returning an array containing messages from any delete errors
+func BatchDelete(ctx context.Context, props *BatchDeleteRequest) []string {
```
cprice404, line 60: if this function is potentially returning errors, then i think golang users will want to be able to do the whole `if err != nil` pattern on the line of code following their call to this.  Can you encapsulate the list of errors/error messages into a wrapper error class that would allow them to structure their code that way?

cprice404 (commented):
a couple of minor questions/nits but overall looks great!

batchutils/batch_operations.go
pgautier404: Mostly so they'd provide information about which keys failed.

batchutils/batch_operations.go
pgautier404: That's a great call, I'll take a look at that.

batchutils/batch_operations.go
pgautier404: Your comment got me thinking, and I switched this up to use a map from failing key to error message instead in a6353c2. It ended up being a little more complicated to consume, and if you think its *over*complicated let me know.

batchutils/batch_operations.go
```diff
@@ -56,8 +63,24 @@ type BatchDeleteRequest struct {
...
 }
 
-// BatchDelete deletes a slice of keys from the cache, returning an array containing messages from any delete errors
-func BatchDelete(ctx context.Context, props *BatchDeleteRequest) []string {
+// BatchDeleteError contains a map associating failing cache keys with their specific errors.
+// It may be necessary to use a type assertion to access the messages:
+//
+// messages := err.(*BatchDeleteError).Messages()
+type BatchDeleteError struct {
+	messages map[momento.Value]error
```
cprice404, line 71: should we name this property `errors` instead of `messages`?  I don't feel too strongly about it but they are actual error objects now, right?

cprice404 (dismissed):
one minor naming suggestion which you can take or leave

batchutils/batch_operations.go
pgautier404: That would be clearer, thanks. I'll make that change.

cprice404 (approved):

PR #295, opened 2023-04-05
Update README.md

NoSQLKnowHow:
Fixed spelling of cache, added more comments, and updated the CLI commands to create a module.


cprice404 (approved):
lgtm, ty

cprice404:
@NoSQLKnowHow - to fix the build failure you need to add `chore: ` as a prefix to your PR title / commit message.

We have a validator that tries to ensure we are using [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) standards in all of our commits, which helps with tooling that generates release notes and changelogs.

PR #308, opened 2023-05-22
chore: add `increment` API

### LINKED ISSUE #304: Add `Increment`

poppoerika:
Added `increment` API along with tests to check:
- if increments from 0 to expected amount with string field
- if increments from 0 to expected amount with bytes field
- if increments with setting and resetting field

Closes #304

### DIFF
--- a/momento/scalar_test.go
+++ b/momento/scalar_test.go
@@ -531,4 +531,167 @@ var _ = Describe("Scalar methods", func() {
 			).To(BeAssignableToTypeOf(&DecreaseTtlMiss{}))
 		})
 	})
+
+	Describe("Increment", func() {
+		It("Increments from 0 to expected amount with string field", func() {
+			field := String("field")
+
+			resp, err := sharedContext.Client.Increment(sharedContext.Ctx, &IncrementRequest{
+				CacheName: sharedContext.CacheName,
+				Field:     field,
+				Amount:    1,
+			})
+			Expect(
+				resp,
+			).To(BeAssignableToTypeOf(&IncrementSuccess{}))
+			Expect(err).To(BeNil())
+			switch result := resp.(type) {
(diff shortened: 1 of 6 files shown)


momento/scalar_test.go
```diff
@@ -531,4 +531,167 @@ var _ = Describe("Scalar methods", func() {
 			).To(BeAssignableToTypeOf(&DecreaseTtlMiss{}))
 		})
 	})
+
+	Describe("Increment", func() {
```
cprice404, line 535: you should add a test that covers a negative value for amount.

cprice404 (commented):
🙈 other than I would like a test that covers a negative amount.  Also, CI is red.

And I'm not really well versed with our go sdk so you should get Ellery and/or Pete to do final sign-off.

momento/scalar_test.go
poppoerika: I have it [here](https://github.com/momentohq/client-sdk-go/pull/308/files#diff-cc5e0552d15047a549c2801866c09b716f51430a815576d4299002a29fa8db7dR571-R586) for a string field and [here](https://github.com/momentohq/client-sdk-go/pull/308/files#diff-cc5e0552d15047a549c2801866c09b716f51430a815576d4299002a29fa8db7dR623-R638) for a bytes field.

Is it better if I take them out and make a separate test for a negative amount for easy readability?

pgautier404 (dismissed):
I'm happy with this when @cprice404 is, but we need to figure out these test failures. If we've switched to V1 tokens for testing, I would expect these to fail and the tests need to be modified. Using legacy tokens, the credential provider's `getAuthToken` method returns the auth token it was originally constructed with. For V1 tokens, however, `getAuthToken` returns an api token extracted from the original, so they are no longer expected to match as the tests assert.

PR #311, opened 2023-05-24
feat: add item get type and ttl

pgautier404:
This commit adds the `ItemGetType` and `ItemGetTtl` operations.


momento/scalar_test.go
```diff
@@ -694,4 +694,98 @@ var _ = Describe("Scalar methods", func() {
...
+			resp, err := sharedContext.Client.ItemGetTtl(sharedContext.Ctx, &ItemGetTtlRequest{
+				CacheName: sharedContext.CacheName,
+				Key:       String("hi"),
+			})
+			Expect(err).To(BeNil())
+			Expect(resp).To(BeAssignableToTypeOf(&ItemGetTtlHit{}))
+			switch result := resp.(type) {
+			case *ItemGetTtlHit:
+				Expect(ttl > result.RemainingTtl()).To(BeTrue())
+				Expect(result.RemainingTtl().Milliseconds() > (time.Second * 30).Milliseconds()).To(BeTrue())
```
poppoerika, line 785: These are checking if the remaining ttl is shorter than the set ttl in seconds and milliseconds to check if ttl is going down?

pgautier404: This is checking that it is between the original TTL (60 seconds) and 30 seconds.

poppoerika: Oh I meant [this part](https://github.com/momentohq/client-sdk-go/pull/311/files#diff-cc5e0552d15047a549c2801866c09b716f51430a815576d4299002a29fa8db7dR784) as well.
And why 30 seconds?

pgautier404: I just removed the `.Milliseconds()` calls from the second `Expect` in 4bfa3a7 as they were not necessary and were confusing. 

The first `Expect` makes sure the remaining ttl is shorter than the original ttl. The second `Expect` makes sure the remaining ttl is longer than 30 seconds. It's 30 seconds because I can be relatively sure that the remaining ttl should be higher (probably considerably) higher than that.

poppoerika: Understood! Thank you for the explanation.
I was thinking how we can test the exact remaining ttl but it looks like checking the remaining ttl is shorter would be a reasonable option.

poppoerika (approved):
Looks good to me 👍

PR #320, opened 2023-06-07
feat: add support to resubscribe to topics on disconnect

### LINKED ISSUE #319: topic logger timestamp
when logging data from the topic logger, I am seeing timestamps that are inaccurate, and don't change. Every log message is output with the following timestamp

```
[2006-01-02T15:04:05Z07:00] 20 (topic-client): failed to topic publish...
[2006-01-02T15:04:05Z07:00] 20 (topic-client): failed to reconnect to stream, will continue to try in 5s seconds
[2006-01-02T15:04:05Z07:00] 20 (topic-client): Attempting reconnecting to client stream
```

bruuuuuuuce:
- fixes logging format to correctly display timestamp closes #319 
- fixes logging to correctly log traces vs debug vs error ..., and log level string instead of number
- ~~adds stream retry interceptor to retry `subscribe` api call~~
- implements "indefinite retry" for subscriptions who are cancelled
- adds debug logging around stream reconnection

### DIFF
--- a/internal/grpcmanagers/data_manager.go
+++ b/internal/grpcmanagers/data_manager.go
@@ -45,7 +45,7 @@ func NewStreamDataGrpcManager(request *models.DataStreamGrpcManagerRequest) (*Da
 	conn, err := grpc.Dial(
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithChainStreamInterceptor(interceptor.StreamInterceptor(request.RetryStrategy), interceptor.AddStreamHeaderInterceptor(authToken)),
+		grpc.WithChainStreamInterceptor(interceptor.StreamRetryInterceptor(request.RetryStrategy), interceptor.AddStreamHeaderInterceptor(authToken)),
 	)
 
 	if err != nil {
--- a/internal/grpcmanagers/data_manager.go
+++ b/internal/grpcmanagers/data_manager.go
@@ -28,8 +28,7 @@ func NewUnaryDataGrpcManager(request *models.DataGrpcManagerRequest) (*DataGrpcM
 	conn, err := grpc.Dial(
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy)),
-		grpc.WithUnaryInterceptor(interceptor.AddHeadersInterceptor(authToken)),
+		grpc.WithChainUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy), interceptor.AddHeadersInterceptor(authToken)),
 	)
 	if err != nil {
 		return nil, momentoerrors.ConvertSvcErr(err)
@@ -46,9 +45,9 @@ func NewStreamDataGrpcManager(request *models.DataStreamGrpcManagerRequest) (*Da
 	conn, err := grpc.Dial(
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithDisableRetry(),
-		grpc.WithStreamInterceptor(interceptor.AddStreamHeaderInterceptor(authToken)),
+		grpc.WithChainStreamInterceptor(interceptor.StreamInterceptor(request.RetryStrategy), interceptor.AddStreamHeaderInterceptor(authToken)),
 	)
+
 	if err != nil {
 		return nil, momentoerrors.ConvertSvcErr(err)
 	}
@@ -61,6 +60,7 @@ func NewLocalDataGrpcManager(request *models.LocalDataGrpcManagerRequest) (*Data
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithDisableRetry(),
 	)
+	
 	if err != nil {
 		return nil, momentoerrors.ConvertSvcErr(err)
 	}
--- a/internal/grpcmanagers/data_manager.go
+++ b/internal/grpcmanagers/data_manager.go
@@ -60,7 +60,7 @@ func NewLocalDataGrpcManager(request *models.LocalDataGrpcManagerRequest) (*Data
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithDisableRetry(),
 	)
-	
+
 	if err != nil {
 		return nil, momentoerrors.ConvertSvcErr(err)
 	}
--- a/internal/grpcmanagers/data_manager.go
+++ b/internal/grpcmanagers/data_manager.go
@@ -45,7 +45,7 @@ func NewStreamDataGrpcManager(request *models.DataStreamGrpcManagerRequest) (*Da
 	conn, err := grpc.Dial(
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithChainStreamInterceptor(interceptor.StreamRetryInterceptor(request.RetryStrategy), interceptor.AddStreamHeaderInterceptor(authToken)),
+		grpc.WithChainStreamInterceptor(interceptor.AddStreamHeaderInterceptor(authToken)),
 	)
 
 	if err != nil {
--- a/internal/interceptor/stream_retry_interceptor.go
+++ b/internal/interceptor/stream_retry_interceptor.go
@@ -9,8 +9,8 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-// streamInterceptor is an example stream interceptor.
-func StreamInterceptor(retryStrat retry.Strategy) func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
+// StreamInterceptor is an interceptor that handles retries for streaming apis
+func StreamRetryInterceptor(retryStrat retry.Strategy) func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
 	return func(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
 		attempt := 1
 		for {
--- a/internal/interceptor/stream_retry_interceptor.go
+++ b/internal/interceptor/stream_retry_interceptor.go
@@ -16,11 +16,14 @@ func StreamRetryInterceptor(retryStrat retry.Strategy) func(ctx context.Context,
 		for {
 			s, lastErr := streamer(ctx, desc, cc, method, opts...)
 			if lastErr == nil {
-				// Success no error returned stop interceptor
+				// Successfully connected the stream. If we wanted to get fancy here, we could return 
+				// a struct that implements the Stream interface, and wraps the SendMsg and ReceiveMsg function
+				// calls. Since we only care about retrying on subscribe, we just return the stream here
+				// ex. https://github.com/grpc/grpc-go/blob/7a7caf363d9b33bfa5ddac83e7dab744f695fb5b/examples/features/interceptor/server/main.go#L106-L141
 				return s, nil
 			}
 
-			// Check retry eligibility based off last error received
+			// Unable to connect to the stream, checking retry eligibility based off last error received
 			retryBackoffTime := retryStrat.DetermineWhenToRetry(retry.StrategyProps{
 				GrpcStatusCode: status.Code(lastErr),
 				GrpcMethod:     method,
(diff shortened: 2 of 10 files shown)


momento/pubsub_client.go
```diff
@@ -22,6 +22,7 @@ type pubSubClient struct {
 func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, momentoerrors.MomentoSvcErr) {
 	streamDataManager, err := grpcmanagers.NewStreamDataGrpcManager(&models.DataStreamGrpcManagerRequest{
 		CredentialProvider: request.CredentialProvider,
+		RetryStrategy:      request.Configuration.GetRetryStrategy(),
```
cprice404, line 25: i'm not really sure we want to use the same retry strategy here that we use for normal unary requests. we can use the same interface/class, but, trying to think about the ux of this from a user perspective.

If I created a Configuration object and intended it to be shared between the CacheClient and the TopicClient in my app, it's very unlikely I'd want the same retry behavior for cache requests that I do for Topic requests; at least for the subscribe part.

That could be worked around by telling users to create a separate Configuration object for CacheClient vs. TopicClient, though it might be a little confusing. but even if we decide that's okay for UX, then the question is - as a consumer of TopicClient, would I want the same retry behavior for Publish and Subscribe? I *think* that personally I'd probably want a separate retry behavior for subscribe.

@bruuuuuuuce can you confirm that as far as 'subscribe' goes, this retry config only applies to the initial subscription?

@kvcache do you have thoughts on UX? Do you think users would expect to be able to set a different retry strategy for publish vs. initial subscribe call?

momento/pubsub_client.go
```diff
@@ -44,9 +45,9 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
 
 func (client *pubSubClient) TopicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (grpc.ClientStream, error) {
 	streamClient, err := client.streamGrpcClient.Subscribe(ctx, &pb.XSubscriptionRequest{
-		CacheName: request.CacheName,
-		Topic:     request.TopicName,
-		//ResumeAtTopicSequenceNumber: 0, TODO think about re-establish topic case
+		CacheName:                   request.CacheName,
+		Topic:                       request.TopicName,
+		ResumeAtTopicSequenceNumber: request.ResumeAtTopicSequenceNumber,
```
cprice404, line 50: what does this change do (vs. previously commented out code)?

momento/topic_subscription.go
```diff
@@ -3,91 +3,92 @@ package momento
...
+	topicName               string
+	log                     logger.MomentoLogger
+	lastKnownSequenceNumber uint64
+	mutex                   sync.Mutex
 }
 
 func (s *topicSubscription) Item(ctx context.Context) (TopicValue, error) {
 	for {
 		rawMsg := new(pb.XSubscriptionItem)
 		if err := s.grpcClient.RecvMsg(rawMsg); err != nil {
```
cprice404, line 33: so you seem to be retrying on all errors now, not just the ones that @eaddingtonwhite had filtered on before. Can you explain the rationale a bit? Not saying that's wrong, just need to understand the change.

momento/topic_subscription.go
```diff
@@ -3,91 +3,92 @@ package momento
...
-		returnErr = s.reInitStream(ctx)
-	} else if grpcStatusErr, ok := status.FromError(err); ok {
-		// FIXME ideally we could retry on raw h2 error not grpc error
-		// See this ticket for follow up https://github.com/momentohq/client-sdk-go/issues/156
-		if grpcStatusErr.Code() == codes.Internal &&
-			grpcStatusErr.Message() == "stream terminated by RST_STREAM with error code: NO_ERROR" {
-
-			returnErr = s.reInitStream(ctx)
+func (s *topicSubscription) lockingReconnect(ctx context.Context) {
+	s.mutex.Lock()
```
cprice404, line 64: this scares my pants off. why is a lock necessary here?

if we determine that it is indeed necessary it is vital for you to put comments here explaining why.

momento/topic_subscription.go
```diff
@@ -3,91 +3,92 @@ package momento
...
-			case *pb.XSubscriptionItem_Heartbeat:
-				continue
-			default:
-				s.log.Trace("Unrecognized response detected.",
-					"response", fmt.Sprint(typedMsg))
-				continue
+			s.log.Error("stream disconnected, attempting to reconnect err:", fmt.Sprint(err))
+			s.lockingReconnect(ctx)
+			// retry getting the latest item
+			continue
```
cprice404, line 37: ankita had written up a list of scenarios that we needed to test this in to observe the behavior. It would be helpful if you could write up some brief notes about what you observed in each of the scenarios, and include that in the PR description as well as add it to a comment on the ticket.

internal/grpcmanagers/data_manager.go
```diff
@@ -28,8 +28,7 @@ func NewUnaryDataGrpcManager(request *models.DataGrpcManagerRequest) (*DataGrpcM
 	conn, err := grpc.Dial(
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy)),
-		grpc.WithUnaryInterceptor(interceptor.AddHeadersInterceptor(authToken)),
+		grpc.WithChainUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy), interceptor.AddHeadersInterceptor(authToken)),
```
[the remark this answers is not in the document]
bruuuuuuuce: From my testing, it looked like before both interceptors were not getting applied. Only the last interceptor was getting applied. From the docs https://pkg.go.dev/google.golang.org/grpc#WithChainUnaryInterceptor, this seems to be how we are supposed to combine multiple interceptors

internal/interceptor/stream_retry_interceptor.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+	"time"
+
+	"github.com/momentohq/client-sdk-go/internal/retry"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/status"
+)
+
+// StreamInterceptor is an interceptor that handles retries for streaming apis
```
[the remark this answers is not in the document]
bruuuuuuuce: the way this interceptor is written, the retry only happens during the publish/subscribe methods. You can get fancy with the stream interceptors and modify the `RecvMsg` method, but for the retry interceptor I did not do that

momento/pubsub_client.go
bruuuuuuuce: From my understanding of our retry strategy, we have just implemented a super basic, no backoff retry strategy, knowing that we are going to add more advanced retry strategies later on.

I was initially confused when using the topic client because it takes in a configuration object with a retry strategy, but didn't use it anywhere.

momento/pubsub_client.go
bruuuuuuuce: previously, we would try and just start resubscribing to a topic. This makes a best effort to start streaming from where the client last left off

https://github.com/momentohq/client-protos/blob/main/proto/cachepubsub.proto#L105-L113

momento/topic_subscription.go
bruuuuuuuce: I was trying to make this impl on par with our rust sdk which also tries to reconnect as long as a user is awaiting an item from the stream

https://github.com/momentohq/client-sdk-rust/pull/144

momento/topic_subscription.go
bruuuuuuuce: the thought behind this was that if we have users who have multiple listeners to items, I didn't want them all to try and reconnect at the same time. This only allows a single listener (per client) to attempt reconnecting. 

```
client := new TopicClient()

sub1 := client.subscribe("topic1")
sub2 := client.subscribe("topic2")
...
subn := client.subscribe("topicn")
```

without the mutex lock, if a disconnect happens then we will have n subscribers try and reconnect to the stream at once


although now that I say this out loud I think I need to move the mutex up one level so all the `TopicSubscriptions` share the same mutex

momento/pubsub_client.go
bruuuuuuuce: > @bruuuuuuuce can you confirm that as far as 'subscribe' goes, this retry config only applies to the initial subscription?

yep this is correct for this impl of the retry strategy

kvcache: What if we did not include a retry strategy for topics clients?

Publish is a non-idempotent operation, and subscribe is something we expect to be long-lived: A failure to establish a subscription up front seems like a reasonable thing to tell a user about.

We can worry about retries later.

momento/topic_subscription.go
kvcache: you expect `func (s *topicSubscription) Item(ctx context.Context) (TopicValue, error) {` to be invoked concurrently, _in parallel_ for 1 Subscription?? If this is the case, let's all get on a zoom and talk about this. It does not match with any expected use case for this component in any language I've talked with anyone else about.

momento/pubsub_client.go
bruuuuuuuce: > What if we did not include a retry strategy for topics clients?

We don't have to but i found that it was confusing that the client already takes in a configuration object with a retry strategy, and our examples use a topic client with one.

momento/topic_subscription.go
bruuuuuuuce: I can remove it its only a few line change, but there is nothing preventing a user from doing this right?

bruuuuuuuce: I guess now that I think about it more, Im not sure the behavior that would occur if a user did this. Would the parallel tasks get alternating items? Or would one task just consume all the items and the other task would just be sitting waiting indefinitely

kvcache: Did you find out if `s.grpcClient.RecvMsg(rawMsg)` is a reentrant function? Pretty sure it's not. I'm not sure why we'd mix reentrancy semantics outside of our control within our client. Instead, I think thread semantics are better accounted for with something like
```
/**
  * Blah blah, comment about a cool method.
  *  
  * Thread safety: Single thread concurrent access only.
  */
```
and

```
/**
  * Blah blah, comment about a cool struct.
  *  
  * Thread safety: Single thread concurrent access only.
  */
```
given that this language does not have a way to express thread safety semantics at compile time.

bruuuuuuuce: > Did you find out if s.grpcClient.RecvMsg(rawMsg) is a reentrant function

correct it is not, it gets invoked outside of the interceptor see

https://github.com/grpc/grpc-go/blob/7a7caf363d9b33bfa5ddac83e7dab744f695fb5b/examples/features/interceptor/server/main.go#L106-L141

internal/interceptor/stream_retry_interceptor.go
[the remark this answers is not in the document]
cprice404: okay, i agree with that approach but you should explain it in comments

momento/pubsub_client.go
cprice404: yeah in retrospect we shouldn't have made the constructor for the topic client take a config object, but I didn't catch this in the early flurry of prs where this stuff came in.  @bruuuuuuuce  is this client using anything else from Config?

bruuuuuuuce: just the logger as of right now

momento/topic_subscription.go
cprice404: ok

momento/pubsub_client.go
cprice404: i know it would technically be a breaking change but I kind of think we should rip the band-aid off and make a new TopicConfiguration class/interface, with only the logger on it for now. To avoid the confusion you had, Matt, and set us up better for the future. wdyt?  cc: @kvcache

bruuuuuuuce: should that be done as a follow up pr or in this same pr

cprice404: either way is fine

internal/interceptor/stream_retry_interceptor.go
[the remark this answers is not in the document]
cprice404: i think it might be useful to keep this for the future even if we don't use it now but i would add a comment explaining that your current impl only hooks into the connect lifecycle, not the RecvMsg one

internal/models/requests.go
```diff
@@ -21,6 +21,7 @@ type DataGrpcManagerRequest struct {
 
 type DataStreamGrpcManagerRequest struct {
 	CredentialProvider auth.CredentialProvider
+	RetryStrategy      retry.Strategy
```
cprice404, line 24: you don't need this anymore based on latest convo right?

internal/interceptor/stream_retry_interceptor.go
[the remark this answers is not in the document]
bruuuuuuuce: 💯 I wasthinking the same thing. I added a comment here detailing the implementation 
https://github.com/momentohq/client-sdk-go/pull/320/files#diff-614e6a567658ca75e5a87273529bf3324e3c518bce2517ddaf0d73e6af8e9556R19-R22

internal/models/requests.go
bruuuuuuuce: correct ill remove it

PR #321, opened 2023-06-08
fix: add new TopicsConfiguration struct for topics client

bruuuuuuuce:
previously the topics client was taking in the same configuration client as the cache client. This configuration contained additional properties such as clientSideRetries and transportStrategies that the topics client was not using. This pr adds a new `TopicConfiguration` that contains only necessary properties for the topics client.


momento/pubsub_client.go
```diff
@@ -28,7 +29,7 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
 	}
 	unaryDataManager, err := grpcmanagers.NewUnaryDataGrpcManager(&models.DataGrpcManagerRequest{
 		CredentialProvider: request.CredentialProvider,
-		RetryStrategy:      request.Configuration.GetRetryStrategy(),
+		RetryStrategy:      retry.NewNoopRetryStrategy(),
```
cprice404, line 32: why do we have a retry strategy here at all? you are just leaving it as a placeholder in case we want to take advantage of your interceptor in the future?

internal/retry/retry.go
```diff
@@ -22,6 +22,17 @@ type Strategy interface {
 	DetermineWhenToRetry(props StrategyProps) *int
 }
 
+type noopRetryStrategy struct{}
```
cprice404, line 25: Can we call this `neverRetryStrategy`

cprice404 (commented):
couple nits

momento/pubsub_client.go
bruuuuuuuce: this is a `unaryDataGrpcManager`, for the publish api calls. These were never retried since they are not idempodent

cprice404: okay. still seems like we don't really have a need for a RetryStrategy property on this class at all right now so I would probably just remove it, but if you wanna keep it, and it's just hard-coded to `NeverRetryStrategy` that's fine too

bruuuuuuuce: I can create a new `grpcmanagers.NewUnaryStreamDataGrpcManager`, this code is just reused inside of the cache client as well. Maybe `grpcmanagers.NewPublishDataGrpcManager`?

pgautier404 (approved):
LGTM when Chris is good w/ it.

momento/pubsub_client.go
cprice404: ah okay i get it

 this is fine then

cprice404 (approved):

PR #331, opened 2023-07-03
feat: add request and response for set if not exists

pratik151192:
Added request and response for ```SetIfNotExists```. Next PR will update the cache client to have the method and the integration tests. I actually did wire the cache client and wrote a vanilla integ test for the new API and it succeeded. Will expand with wholistic tests and have it as a new PR.

### DIFF
--- a/responses/set_if_not_exists.go
+++ b/responses/set_if_not_exists.go
@@ -1,0 +1,44 @@
+package responses
+
+// SetIfNotExistsResponse is the base response type for a SetIfNotExists request
+type SetIfNotExistsResponse interface {
+	isSetIfNotExistsResponse()
+}
+
+// SetIfNotExistsNotStored indicates a successful set request where the value was already present.
+type SetIfNotExistsNotStored struct{}
+
+func (SetIfNotExistsNotStored) isSetIfNotExistsResponse() {}
+
+// SetIfNotExistsStored indicates a successful set request where the value was stored.
+type SetIfNotExistsStored struct {
+	key   []byte
+	value []byte
+}
+
+func (SetIfNotExistsStored) isSetIfNotExistsResponse() {}
+
+// ValueString returns value stored in the cache.
+func (resp SetIfNotExistsStored) ValueString() string {
+	return string(resp.value)
+}
+
+// ValueByte Returns value stored in the cache as bytes.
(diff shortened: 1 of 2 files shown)


responses/set_if_not_exists.go
```diff
@@ -0,0 +1,44 @@
...
+
+// KeyByte Returns key stored in the cache as bytes.
+func (resp SetIfNotExistsStored) KeyByte() []byte {
+	return resp.key
+}
+
+// NewSetIfNotExistsStored returns a new SetIfNotExistsStored containing the supplied key and value.
+func NewSetIfNotExistsStored(key, value []byte) *SetIfNotExistsStored {
+	return &SetIfNotExistsStored{key: key, value: value}
+}
```
pratik151192, line 44: okay this part is interesting because our Java SDK client provides these helpers and the key,value as a part of the scalar SetResponse; whereas JavaScript and Go SDK do not. I am not 100% on why we took the decision for Go and JS to not return these as a part of the response, but is it latency? If that's the case, then I can get rid of these in the response.

cprice404: I don't think this would be a latency issue because I don't think server actually returns the value, so we would just be populating it from our own variables in memory.

I'm not sure why we have the inconsistency but here are some rules of thumb for public APIs:
* Consistency within an SDK is the most important thing
* It's always easier to add things than to remove things, so err on the side of minimizing the footprint
* Our JS SDK is the closest thing we have to "canonical" so I would default to mirroring its behavior when there is a question. The Java one is one of the more recent and least used, so I would generally not use it as a model unless it is the only SDK that has a certain API implemented.

Given the above, my rec would be to remove these accessors for now.

cprice404 (commented):
looks reasonable to me! pending response to comment below, and Pete's review.

responses/set_if_not_exists.go
pratik151192: SGTM. Will remove it as I agree with 

> It's always easier to add things than to remove things, so err on the side of minimizing the footprint

cprice404 (approved):
+1 from me but @pgautier404 should sign off.

pgautier404 (approved):
LGTM!

PR #332, opened 2023-07-05
chore: copy protos for new APIs to go sdk and fix MakeFile

pratik151192:
Updated MakeFile to provide the protoc path as ```internal/protos``` rather than ```internal```. This fixed the problem of generating the proto files that was mentioned on the issue linked to this PR. However, there are certain changes I wanted to call out due to which one of our CIs is also failing:

- I'm not sure what was happening before and how was this succeeding, but seems like ```protoc``` wants the directory that contains all the ```.proto``` files and ```internal/protos``` is the correct one for both reading the input files and generating the output ones.
- Doing this, changed some variable and method names from ```protos_auth_proto``` to ```auth_proto``` whose format takes the directory structure. Since we told ```protoc``` to take files from internal/protos rather than internal only, all those source changes have been reflected either in ```source``` field or the variable/method names.
- Our CI for ```Verify no changes from goimports and go mod tidy``` fails because some of the imports were reordered as a part of this update to the input and output directories. Note that the protoc version was also bumped from ```3.20.3``` to ```4.23.3```

We should really invest some time in this https://github.com/momentohq/client-sdk-go/issues/333 as well and am going to do some digging once I have some time off my tasks. 

None of this seem breaking to me but I'm not 100% so we can discuss over a call as well.


cprice404 (commented):
I will hold off on weighing in on this until @eaddingtonwhite and @pgautier404 do, they will have more context than I do.

.github/workflows/test.yml
```diff
@@ -42,6 +42,15 @@ jobs:
...
 
+      - name: Build
+        run: make build
+
+      - name: Build examples
+        run: make build-examples
+
+      - name: Run test
+        run: make test
+
```
pgautier404, line 53: I would suggest you move these back below the linting just so the CI fails fast if there are any formatting issues.

pratik151192: Makes sense!

pgautier404 (approved):
LGTM, thanks!

PR #335, opened 2023-07-05
feat: add requests and responses to get lengths for sets, sorted sets and dictionary

pratik151192:
Added requests and responses to obtain lengths for set, sorted set, sorted set length by score, and dictionary.


momento/sorted_set_length_by_score.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/momentohq/client-sdk-go/responses"
+
+	pb "github.com/momentohq/client-sdk-go/internal/protos"
+)
+
+type SortedSetLengthByScoreRequest struct {
+	CacheName string
+	SetName   string
+	MinScore  *float64
+	MaxScore  *float64
```
cprice404, line 15: I presume this is the same way we were exposing optional params for existing APIs like FetchByScore?

cprice404 (dismissed):
lgtm but @pgautier404 should review.

one thing to keep in mind - if we merge this PR without the corresponding implementation, then we sort of can't do a release of the Go SDK off of the main branch until the implementation comes in. Just something to keep in mind in case an urgent bug fix elsewhere was needed. Not a big deal as long as the implementation PR is merged shortly after this one.

momento/sorted_set_length_by_score.go
pratik151192: that's correct.

pratik151192:
Note that this is blocked on https://github.com/momentohq/client-sdk-go/pull/332#pullrequestreview-1515153029 anyway

pratik151192:
This PR has become a bit clumsy due to a bunch of merges and rebases. I'm going to discard this and cut a new one.

PR #341, opened 2023-07-21
chore: add topics loadgen

pgautier404:
This commit adds a load generator for topics and updates the SDK dependency for the examples to the latest.


examples/loadgen-topics/main.go
```diff
@@ -0,0 +1,421 @@
...
+			publishErrorCounter.updateErrors(errCode)
+		default:
+			time.Sleep(time.Millisecond * 25)
+		}
+	}
+}
+
+func (r *loadGenerator) run(ctx context.Context, client momento.TopicClient) {
+	cancelContext, cancelFunction := context.WithTimeout(ctx, r.options.howLongToRun)
+	defer cancelFunction()
```
pratik151192, line 342: Does this automatically cancel the run after the method completes? Not sure what this does

examples/loadgen-topics/main.go
```diff
@@ -0,0 +1,421 @@
...
+		"max",
+		publishes.Max(),
+	))
+}
+
+func readablePercentage(numerator int64, denominator int64) int {
+	return int(math.Ceil(float64(numerator) / float64(denominator) * 100))
+}
+
+func timer(
```
pratik151192, line 295: If I understand this method correctly, it loops infinitely until the context signals Done. Is this signal triggered after ```howLongToRun``` via ```defer cancelFunction``` inside the run method?

examples/loadgen-topics/main.go
```diff
@@ -0,0 +1,421 @@
...
+
+		select {
+		case <-ctx.Done():
+			fmt.Println("\n=====> run complete <=====")
+			printStats(subscribeHistogram, publishHistogram, subscribeErrorCounter, publishErrorCounter, origStartTime)
+			return
+		case subscribeMessage := <-subscribeChan:
+			if err := subscribeHistogram.RecordValue(subscribeMessage); err != nil {
+				panic(err)
+			}
```
pratik151192, line 325: Here it seems like whenever there's some value on the subscriber channel, we will hit this case and record it in the histogram. Question: Since we are polling every 25 milliseconds, there could be multiple values on the channel right? Will this record all those values? If yes, is that one by one randomly or all together?

pratik151192 (commented):
Looks great. Just have a few ```go``` specific questions

pratik151192:
Hmm I ran the test by tweaking the numUsers, bytes etc to be super small but got an error 

```
panic: strings: negative Repeat count

goroutine 1 [running]:
strings.Repeat({0x1031efb60?, 0x0?}, 0x140001a1db8?)
        /opt/homebrew/opt/go/libexec/src/strings/strings.go:538 +0x394
main.newLoadGenerator({0x103336e20, 0x1400000c7e0}, {{0x1400003c740, 0x35}, 0xa, 0x12a05f200, 0xa, 0x1, 0x1, 0x1, ...})
        /Users/pratik/sandbox/client-sdk-go/examples/loadgen-topics/main.go:56 +0x108
main.main()
        /Users/pratik/sandbox/client-sdk-go/examples/loadgen-topics/main.go:412 +0x138

```


It's coming from

```
	messageValue := strings.Repeat("x", options.messageBytes-len(strconv.FormatInt(unixMilli, 10)))

```

cprice404 (approved):
lgtm once you have answered pratik's questions!

examples/loadgen-topics/main.go
pgautier404: Yes, `defer` ensures that the `cancelFunction` method is called at the end of the enclosing block. When that is called, it sends a signal on the `cancelContext`'s `Done()` channel that tells the `publishMessages`, `pollForMessages`, and `timer` routines to return.

examples/loadgen-topics/main.go
pgautier404: Yes.

examples/loadgen-topics/main.go
pgautier404: It will grab one message from a channel with a queued message per iteration, so if there are multiple messages on the channel it will grab the next available and then go again from the top of the loop.

pgautier404:
> Hmm I ran the test by tweaking the numUsers, bytes etc to be super small but got an error
> 
> ```
> panic: strings: negative Repeat count
> 
> goroutine 1 [running]:
> strings.Repeat({0x1031efb60?, 0x0?}, 0x140001a1db8?)
>         /opt/homebrew/opt/go/libexec/src/strings/strings.go:538 +0x394
> main.newLoadGenerator({0x103336e20, 0x1400000c7e0}, {{0x1400003c740, 0x35}, 0xa, 0x12a05f200, 0xa, 0x1, 0x1, 0x1, ...})
>         /Users/pratik/sandbox/client-sdk-go/examples/loadgen-topics/main.go:56 +0x108
> main.main()
>         /Users/pratik/sandbox/client-sdk-go/examples/loadgen-topics/main.go:412 +0x138
> ```
> 
> It's coming from
> 
> ```
> 	messageValue := strings.Repeat("x", options.messageBytes-len(strconv.FormatInt(unixMilli, 10)))
> ```

Good catch. I'm using the messages to hold a timestamp so we can calculate the latency from publication to receipt, so the `messageBytes` value must be at least that long to hold the timestamp. I added a quick validator and. `panic` with an explanatory message if the value is invalid in 7b9a700.

pratik151192 (approved):

Issue #342, opened 2023-08-17
Add eager connection config for golang; make this the default, and make a pre-built Lambda config


cprice404:
## Issue Type

- [ ] feat

## Description

Add a pre-built Lambda Configuration and support for eager connections in the golang SDK.

## Relevant Links and Information

- Nate's JS PR for adding eager connections: https://github.com/momentohq/client-sdk-javascript/pull/711
- Nate's JS PR for adding a pre-built config for Lambda: https://github.com/momentohq/client-sdk-javascript/pull/708
- Nate's JS PR for adding an example lambda: https://github.com/momentohq/client-sdk-javascript/pull/707
- Alpha module for deploying golang lambdas via CDK: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-lambda-go-alpha-readme.html

In the other SDKs I think we are making the eager connection be the default now, and I presume we will want to do that here as well. One thing to keep in mind about this is that there may be unit tests in the repo that test things like credentialproviders, which are instantiating a client and relying on it having lazy connection behavior (because we don't want to actually try to make any network requests as part of the constructor call for these tests). So it's important that there still be a way to construct a client with a lazy connection configuration, and update any such tests to use that mechanism for instantiating their client.

## Acceptance Criteria

- [x] There is a new pre-built Configuration for Lambda in the golang SDK
- [x] We've aligned on what settings it should have, especially w/rt connection counts
- [x] There is a new setting for configuring an eager connection timeout, but some way to explicitly choose lazy connection
- [x] There is an example go-based Lambda in the repo
- [x] We have some timing information to tell us what the latency is for the first request run in the lambda vs. subsequent requests. (Bonus points if we have before/after numbers for how the eager connection impacted this.)
- [x] If the new "best practice" code for instantiating a golang client has changed as a result of this work, we need to make sure we update all the example code in the repo and public dev docs.

PR #343, opened 2023-08-22
feat: lambda config, cdk and lambda example, and grpc eager connections

anitarua:
Addresses #342:
- Adds a prebuilt configuration for lambda environments (currently the same as the InRegion config object)
- Adds an example Lambda function written in Go and deployed with a CDK stack written in TypeScript
- Adds best effort eager connections and a new CacheClient prop field to specify this setting

Apparently in the golang grpc package, the [GetState](https://pkg.go.dev/google.golang.org/grpc#ClientConn.GetState) function doesn't have a way to specify a boolean for encouraging a connection, which seemed to be the main eager connection method in the Node, Python, and Java PRs. 

Instead, this PR uses the [round robin load balancing policy](https://github.com/grpc/grpc-go/blob/master/examples/features/load_balancing/client/main.go#L77) and [ClientConn.Connect](https://pkg.go.dev/google.golang.org/grpc#ClientConn.Connect) to establish eager connections as encouraged by the [grpc-go anti-patterns doc](https://github.com/ulascansenturk/grpc-go/blob/914b161b5b22a85fe5f19836ff9db1954a973d1a/Documentation/anti-patterns.md).

Edit: using Pratik's suggested Connect() function instead which does use a timeout and the connectivity state API for eager connections.


anitarua:
Did some testing on Lambda and there was quite a bit of variation in first request latencies, but the highest values I saw were 200ms for non-eager connections and 40ms for eager connections.

Before eager connections highest latency:
lambda before eager connections highest latency

After eager connections highest latency:
eager connections highest latency

examples/lambda-examples/simple-get/README.md
```diff
@@ -0,0 +1,67 @@
...
+[![project status](https://momentohq.github.io/standards-and-practices/badges/project-status-official.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+[![project stability](https://momentohq.github.io/standards-and-practices/badges/project-stability-stable.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+
+<br>
+
+## Simple Get Lambda
+
+This repo contains an example lambda, built using AWS CDK, that repeatedly calls get on a Momento cache.
+
+The primary use is to provide a base for testing Momento performance in a lambda environment. The lambda creates a Momento client, and then calls get on a hard-coded key 100 times, with a 100ms wait between calls. The metric logging middleware is enabled, so detailed information about each call is logged.
```
pratik151192, line 15: I personally changed this to a simple set/get instead. I thought it was unnecessary steps for a client to create a cache from console/CLI and set. And the lambda just doing a `get` but I'm not opposed to this. Just a thought

internal/grpcmanagers/data_manager.go
```diff
@@ -18,17 +18,30 @@ type DataGrpcManager struct {
...
-		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithChainUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy), interceptor.AddHeadersInterceptor(authToken)),
-	)
+
+	var conn *grpc.ClientConn
+	var err error
+	if eagerlyConnect {
+		conn, err = grpc.Dial(
+			endpoint,
+			grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`),
```
pratik151192, line 33: Okay this is really interesting if providing a load balancing config starts establishing eager connections. From my understanding, there are 2 ways the gRPC client will choose Momento's server. `round_robin` and `pick` (default) where the former picks servers to connect to in a round-robin fashion, and the latter picks whatever is available. From this code, it seems like we will override the load-balancing strategy to round-robin. I know you linked a doc, but can you point to the relevant section where it mentions that doing this also establishes an eager connection?

momento/cache_client.go
```diff
@@ -209,21 +213,23 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
 }
 
 // NewCacheClient returns a new CacheClient with provided configuration, credential provider, and default TTL seconds arguments.
-func NewCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration) (CacheClient, error) {
+func NewCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration, eagerlyConnect bool) (CacheClient, error) {
```
pratik151192, line 216: I believe we wanted to create a new factory method for this and leave the current signatures as is.

examples/lambda-examples/simple-get/README.md
pratik151192: The testing numbers look really good! I do notice that all attempts are Misses so am guessing the key wasn't written to the cache before the run. This is another reason i wanted the example to be basic set/get

internal/grpcmanagers/data_manager.go
anitarua: I was mostly referring to this part of the [anti-patterns doc](https://github.com/ulascansenturk/grpc-go/blob/914b161b5b22a85fe5f19836ff9db1954a973d1a/Documentation/anti-patterns.md):

> In fact, Dial does not always establish a connection to servers by default. The connection behavior is determined by the load balancing policy being used. For instance, an "active" load balancing policy such as Round Robin attempts to maintain a constant connection, while the default "pick first" policy delays connection until an RPC is executed. Instead of using the WithBlock option, which may not be recommended in some cases, you can call the [ClientConn.Connect](https://pkg.go.dev/google.golang.org/grpc#ClientConn.Connect) method to explicitly initiate a connection.

As well as this [comment on an issue](https://github.com/grpc/grpc-java/issues/2292#issuecomment-284486128) from the grpc-java repo: 

> Because round-robin eagerly creates connection, the requestConnection argument in getState() simply doesn't make any difference.

pratik151192: Okay since we will be creating a new factory method I'm not totally opposed to this as it won't change the existing client behaviors if at all. Fwiw I had played around these 2 LB policies in nodeJS and didn't find any significant differences so should be fine.

However, it does seem like we could do the eager connection setup using the `ClientConn.Connect` on our existing LB policy too? Just interpreting from the text you posted. 

@cprice404 thoughts?

momento/scs_data_client.go
```diff
@@ -79,3 +79,7 @@ func (client scsDataClient) makeRequest(ctx context.Context, r requester) error
 
 	return nil
 }
+
+func (client scsDataClient) Connect() {
+	client.grpcManager.Conn.Connect()
```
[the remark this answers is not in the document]
pratik151192: Oh damn I totally missed this part! sorry about that, thanks for pointing it. Will re-review based on this change

internal/grpcmanagers/data_manager.go
pratik151192: Okay scratch the above one as @anitarua posted the line where we are calling `Clientconn.connect` .. I just have one last follow up before I stop annoying you - does the `clientconn.connect` not take effect with the `pick` LB policy? The only reason I'm nudging on this is we might have done experiemnts/tests with the `pick` policy and might have chosen to keep that

momento/cache_client.go
```diff
@@ -187,10 +188,13 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
 		CredentialProvider: props.CredentialProvider,
 		Configuration:      props.Configuration,
 		DefaultTtl:         props.DefaultTtl,
-	})
+	}, props.EagerlyConnect)
 	if err != nil {
 		return nil, convertMomentoSvcErrorToCustomerError(momentoerrors.ConvertSvcErr(err))
 	}
+	if props.EagerlyConnect {
+		dataClient.Connect()
```
pratik151192, line 196: We might want to wrap a timeout for this with a default of 30 seconds and give clients ability to override it in the factory method. Otherwise it will block until the connection is established.

pratik151192:
LGTM overall!! Just a few minor things

internal/grpcmanagers/data_manager.go
anitarua: I wrap both `clientconn.connect` and round robin LB in `if` statements throughout the create cache process, so I think they should activated together and separately from the `pick` LB policy.

The `clientconn.connect` call will only occur if `cache_client.go/commonCacheClient()` receives `eagerlyConnect=true`:

> 
```
if props.EagerlyConnect {
     dataClient.Connect()
}
```

And the eagerlyConnect argument is piped through to `data_manager.go/NewUnaryDataGrpcManager()`

> 
```
if eagerlyConnect {
		conn, err = grpc.Dial(
			endpoint,
			grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`),
			grpc.WithChainUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy), interceptor.AddHeadersInterceptor(authToken)),
		)
	} else {
		conn, err = grpc.Dial(
			endpoint,
			grpc.WithChainUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy), interceptor.AddHeadersInterceptor(authToken)),
		)
	}
```

momento/cache_client.go
anitarua: I'm not entirely sure how blocking/nonblocking calls work in Go, but my impression from the [docs](https://pkg.go.dev/google.golang.org/grpc#ClientConn.Connect) was this was a non-blocking call: 

> Connect causes all subchannels in the ClientConn to attempt to connect if the channel is idle. Does not wait for the connection attempts to begin before returning.

pratik151192: Yeah I did the same in Python SDK, but Chris mentioned we wanted parity and this feature requires a blocking call until the timeout elapses but not a perennially blocking call. So essentially the factory method will connect and block until the timeout, and then give up.

anitarua: Okay, let me take a closer look at that thread and blocking calls in Go.

Thanks for the speedy review and good feedback so far!

internal/grpcmanagers/data_manager.go
anitarua: Okay, been doing some research, and I'm sure my understanding of Go is still quite incomplete, but this is my understanding of things so far:

Addressing the second point first: From the code for [`clientconn.connect`](https://github.com/grpc/grpc-go/blob/v1.57.0/clientconn.go#L710), I see that it calls [`exitIdleMode`](https://github.com/grpc/grpc-go/blob/v1.57.0/clientconn.go#L322) which apparently "moves the channel out of idle mode by recreating the name resolver and load balancer." Combined with the statement "The connection behavior is determined by the load balancing policy being use" from the [anti-patterns doc](https://github.com/ulascansenturk/grpc-go/blob/914b161b5b22a85fe5f19836ff9db1954a973d1a/Documentation/anti-patterns.md), I think this implies that the `round robin` LB policy is still needed to ensure eager connections.

As for the eager connection timeout, I'm not entirely sure of the best course of action. This [Go tutorial](https://grpc.io/docs/languages/go/basics/#calling-service-methods) says that:

> Note that in gRPC-Go, RPCs operate in a blocking/synchronous mode, which means that the RPC call waits for the server to respond, and will either return a response or an error.

I'm not sure if `clientconn.connect`](https://github.com/grpc/grpc-go/blob/v1.57.0/clientconn.go#L710) counts as an RPC call, but if it does, I think it's synchronous already. If it's not, well, it has no return value, so I'm not sure how I would set a timeout to wait until a promise is resolved (or the Go equivalent if there is an equivalent). I could perhaps put in a [`time.Sleep()`](https://pkg.go.dev/time#Sleep) call to sleep for the provided eagerConnectionTimeout duration to try to provide that extra bit of time for the connection to the server to get set up though.

What do you think @pratik151192 ?

Also, if you have a moment to spare, do you any insights from when you worked with Go, @pgautier404 ?

pratik151192: Here's a psuedo/non-production ready sample that I used to achieve the eager connection behavior (sleep() is usually not recommended)

```
func (client scsDataClient) Connect() error {
	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Second)
	defer cancel()

	err := client.grpcManager.Connect(ctx)
	return err
}
```

```
func (gm *DataGrpcManager) Connect(ctx context.Context) error {
	for {
		select {
		case <-ctx.Done():
			// Context timeout or cancellation occurred
			return ctx.Err()
		default:
			// Check current state
			state := gm.Conn.GetState()
			switch state {
			case connectivity.Ready:
				// Connection is ready, exit the method
				fmt.Println("connection is ready")
				return nil
			case connectivity.Idle, connectivity.Connecting:
				fmt.Println("connection is " + state.String())
				// If Idle or Connecting, wait for a state change
				if !gm.Conn.WaitForStateChange(ctx, state) {
					// Context was done while waiting
					return ctx.Err()
				}
			default:
				// For other states like TransientFailure, you might want to handle them differently
				return errors.New("connection is in an unexpected state")
			}
		}
	}
}

```

pratik151192: Some description about the code:

- `defer cancel` is just a best-practice in `go` to clean up resources explicitly when getting a context. The `defer` simply tells it to clean up after we are out of the `Connect()` block.
- `Connect` of the `DataGrpcManager` has a few elements that I ahve added some comments to explain a bit better; but let me know if it's not clear.

Honestly, GPT-4 has been helpful in guiding me with some of this stuff for new languages. I highly recommend taking it's help at times particularly for new languages and when we know the kind of problem we need to solve.

pratik151192: I ran a single test in `scalar_test` and saw those print statements

```
Scalar methods Gets, Sets, and Deletes when the key and value are strings
/Users/pratik/sandbox/client-sdk-go/momento/scalar_test.go:64
connection is CONNECTING
connection is ready
[{golang-default-35888518-9053-4d92-a482-5692c774589a}]
• [2.115 seconds]
```

I also tested with a 1 millisecond timeout value and saw

```
connection is CONNECTING
context deadline exceeded
```

anitarua:
@pratik151192 Thanks for the code snippets, I'll have to look into utilizing GPT-4 more 😄
Tested on Lambda and your suggestion works very well and definitely is more in line with the other SDKs. Thanks so much for your help!!!

With eagerConnectTimeout at 0ms:
Screenshot 2023-08-23 at 2 26 19 PM

With eagerConnectTimeout at 10ms:
Screenshot 2023-08-23 at 2 29 02 PM

Makefile
```diff
@@ -1,3 +1,5 @@
+GOFILES_NOT_NODE = $(shell find . -type f -name '*.go' -not -path "./examples/lambda-examples/simple-get/infrastructure/*")
```
anitarua, line 1: gofmt was somehow finding the Go template files in one of the examples/lambda-examples/simple-get-infrastructure/node_modules files and failing to lint

internal/grpcmanagers/data_manager.go
```diff
@@ -18,17 +22,30 @@ type DataGrpcManager struct {
...
-		grpc.WithTransportCredentials(credentials.NewTLS(config)),
-		grpc.WithChainUnaryInterceptor(interceptor.AddUnaryRetryInterceptor(request.RetryStrategy), interceptor.AddHeadersInterceptor(authToken)),
-	)
+
+	var conn *grpc.ClientConn
+	var err error
+	if eagerConnectTimeout > 0 {
+		conn, err = grpc.Dial(
+			endpoint,
+			grpc.WithDefaultServiceConfig(`{"loadBalancingConfig": [{"round_robin":{}}]}`),
```
pratik151192, line 37: I believe we don't need the round_robin policy override anymore since we are using a different/manual mechanism to connect.

momento/cache_client.go
```diff
@@ -187,11 +188,18 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
...
-	})
+	}, props.EagerConnectTimeout)
 	if err != nil {
 		return nil, convertMomentoSvcErrorToCustomerError(momentoerrors.ConvertSvcErr(err))
 	}
 
+	if props.EagerConnectTimeout > 0 {
+		err := dataClient.Connect()
+		if err != nil {
+			return nil, convertMomentoSvcErrorToCustomerError(momentoerrors.ConvertSvcErr(err))
```
pratik151192, line 199: We aren't doing this in other SDKs (i.e., returning an error if the eager connection attempt fails). Eager connection within timeout is a best-effort though it should almost always succeed. If it doesn't, we should just `log.debug` that we couldn't eagerly connect and return the client as is. When the client actually starts making data plane requests, a connection request will be automatically re-attempted by gRPC.

momento/cache_client.go
```diff
@@ -211,19 +219,33 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
...
+}
+
+// NewEagerCacheClient returns a new CacheClient with provided configuration, credential provider, and default TTL seconds arguments,
+// as well as eagerly attempting to establish gRPC connections
+func NewEagerCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration, eagerConnectTimeout time.Duration) (CacheClient, error) {
+	props := CacheClientProps{
+		Configuration:       configuration,
+		CredentialProvider:  credentialProvider,
+		DefaultTtl:          defaultTtl,
+		EagerConnectTimeout: eagerConnectTimeout,
```
pratik151192, line 237: Can we add a validation before constructing the props to make sure `eagerConnectTimeout >= 0` and is a valid duration in seconds.

internal/grpcmanagers/data_manager.go
```diff
@@ -69,3 +86,32 @@ func NewLocalDataGrpcManager(request *models.LocalDataGrpcManagerRequest) (*Data
...
+		case <-ctx.Done():
+			// Context timeout or cancellation occurred
+			return ctx.Err()
+		default:
+			// Check current state
+			state := gm.Conn.GetState()
+			switch state {
+			case connectivity.Ready:
+				// Connection is ready, exit the method
+				fmt.Println("connection is ready")
```
pratik151192, line 102: My bad as I had commented with this snippet just for test purposes. For a client-facing SDK, we usually don't want these messages to surface to them; hence, we use `log.debug` and not print it to STDOUT. If a client ever actually enables debug logging to debug various issues, this is okay to appear.

internal/grpcmanagers/data_manager.go
```diff
@@ -69,3 +86,32 @@ func NewLocalDataGrpcManager(request *models.LocalDataGrpcManagerRequest) (*Data
...
+				return nil
+			case connectivity.Idle, connectivity.Connecting:
+				fmt.Println("connection is " + state.String())
```
pratik151192, line 105: Same, lets make this `log.debug`

pratik151192 (commented):
Handful of nits to make it production ready and should be good to go! Thank you for your patience with this!

examples/lambda-examples/simple-get/README.md
```diff
@@ -0,0 +1,65 @@
...
+<img src="https://docs.momentohq.com/img/logo.svg" alt="logo" width="400"/>
+
+[![project status](https://momentohq.github.io/standards-and-practices/badges/project-status-official.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+[![project stability](https://momentohq.github.io/standards-and-practices/badges/project-stability-stable.svg)](https://github.com/momentohq/standards-and-practices/blob/main/docs/momento-on-github.md)
+
+<br>
+
+## Simple Get Lambda
+
+This repo contains an example AWS Lambda Function, built using AWS CDK, that sets and gets an item in a Momento cache.
```
cprice404, line 13:
```suggestion
This directory contains an example AWS Lambda Function, built using AWS CDK, that sets and gets an item in a Momento cache.
```

examples/lambda-examples/simple-get/README.md
```diff
@@ -0,0 +1,65 @@
...
+momento cache create cache
+```
+
+You can then set a value for the key:
+
+```commandline
+momento cache set key value
+```
+
+You can also create a cache and key using the [Momento Console](https://console.gomomento.com).
```
cprice404, line 63: I don't have a super strong opinion about this, but I do kind of agree with what @pratik151192 said about making the lambda do the create cache, set, and get, so that there are fewer steps to deal with in the README. not a blocker though, this way is okay too.

cprice404: oh it looks like you did change the lambda, but maybe just not the README?

momento/cache_client.go
```diff
@@ -211,19 +232,33 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
 // NewCacheClient returns a new CacheClient with provided configuration, credential provider, and default TTL seconds arguments.
 func NewCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration) (CacheClient, error) {
 	props := CacheClientProps{
-		Configuration:      configuration,
-		CredentialProvider: credentialProvider,
-		DefaultTtl:         defaultTtl,
+		Configuration:       configuration,
+		CredentialProvider:  credentialProvider,
+		DefaultTtl:          defaultTtl,
+		EagerConnectTimeout: 0,
```
cprice404, line 238: For the other SDKs we've been making eager the default, and setting this to `30` I believe. @nand4011 @pratik151192 ?

momento/cache_client.go
```diff
@@ -211,19 +232,33 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
...
+	}
+	return commonCacheClient(props)
+}
+
+// NewEagerCacheClient returns a new CacheClient with provided configuration, credential provider, and default TTL seconds arguments,
+// as well as eagerly attempting to establish gRPC connections
+func NewEagerCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration, eagerConnectTimeout time.Duration) (CacheClient, error) {
```
cprice404, line 245: I know it'll be an annoyingly long name but I think that since the main factory function will default to eager, then this one should be `NewCacheClientWithEagerConnectTimeout`? wdyt @pratik151192

cprice404 (commented):
looking good, thank you!

momento/cache_client.go
nand4011: Yeah, node is 30 seconds by default.

nand4011 (commented):
LGTM. I haven't worked in go yet so my code review won't be worth much. The only thing I wonder about is if we should think of a better name for the example lambda now that it does more than cal get.

examples/lambda-examples/simple-get/README.md
anitarua: I don't create a cache in the lambda yet, I can update that. 

And @nand4011, regarding the naming of it, I can restructure the directory and rename it so it's just a Lambda example rather than "simple get"

momento/cache_client.go
```diff
@@ -211,19 +232,34 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
 // NewCacheClient returns a new CacheClient with provided configuration, credential provider, and default TTL seconds arguments.
 func NewCacheClient(configuration config.Configuration, credentialProvider auth.CredentialProvider, defaultTtl time.Duration) (CacheClient, error) {
 	props := CacheClientProps{
-		Configuration:      configuration,
-		CredentialProvider: credentialProvider,
-		DefaultTtl:         defaultTtl,
+		Configuration:       configuration,
+		CredentialProvider:  credentialProvider,
+		DefaultTtl:          defaultTtl,
+		EagerConnectTimeout: 30 * time.Second,
```
pratik151192, line 238: @cprice404 I saw your comment here before and missed replying that time. I think this isn't the default here and @anitarua was sending it as 0 as this is the existing constructor where we want to disable eager connections by default.

cprice404: But I thought we were all aligned that we want eager connections to be the default?

momento/cache_client.go
```diff
@@ -211,19 +232,34 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
...
+	}
+	return commonCacheClient(props)
+}
+
+// NewCacheClientWithEagerConnectTimeout returns a new CacheClient with
+// provided configuration, credential provider, and default TTL seconds
+// arguments, as well as eagerly attempting to establish gRPC connections
```
cprice404, line 245: you should mention in this comment that a value of `0` can be used to indicate that you do not want an eager connection

momento/cache_client.go
pratik151192: Yes but this is the existing constructor. The new method is `NewCacheClientWithEagerConnectTimeout`. In other SDKs, the existing constructors do not establish an eager connection. And we decided on that because the existing constructors are sync and do not make a network call.

cprice404 (commented):
I'm +1 on this now sans the one minor addition to the docstring for the new factory that exposes the eager connection option.

Let's make sure we get alignment from @pratik151192 on the change to the default constructor but I think this is the right thing to do.

momento/cache_client.go
cprice404: But this isn't a constructor, it was already a factory function.

If we don't set the eager connection here then we will be forced to tell users to use a different factory function as the best practice. That seems undesirable given that this one is the most obvious / simple one that users will be tempted to change.

I understand that this is a behavior change for this function, but (a) it seems like the right thing to do, and (b) 90% of the eventual users of this SDK haven't started using it yet, so I want to make sure they have the best experience possible.

This approach is different from the ones you guys did in the other SDKs because we were already using factory functions here.

If you disagree and want to discuss we can set up a quick zoom for the 3 of us!

pratik151192: Oh I see; nah, I'm aligned with that. Thank you for the explanation!

pratik151192 (dismissed):
AWESOME!!!

pratik151192 (approved):

cprice404 (approved):
yes

PR #346, opened 2023-08-30
feat: add batchSet operation to batchutils

pratik151192:
Addressing one part of https://github.com/momentohq/dev-eco-issue-tracker/issues/445 to add `BatchSet`. Fast follow for `BatchSetNX` if customer asks for.

As a part of this change, I also re-organized the files so that each `batch` operation is in its own file as it was getting a bit bloated. Note that this won't break customer behavior as `Go` works on `modules` and the `module` for the files are same so from a client perspective their existing code won't break.

GPT also called out a bunch of bugs but I validated through @cprice404 that GPT is acting dumb for most of them. The only ones that seemed relevant to me are:

- A co-routine can get stuck and lead to deadlock so its better to wrap them around timeouts
- There's no limit on the `keys` or `item` sizes a client can pass which can potentially cause memory issues.

Let me know if you all want me to address these in this PR; else I can create a ticket for it and fast follow. 2 seems like something under client's control, while for 1 I believe we can wrap individual calls with a timeout and propagate as such. I didn't want to clutter this PR so happy to follow on it.


cprice404:
if you want to fast-follow on the timeouts thing, that would definitely be an improvement.

I think we can punt the other one for now.

batchutils/batch_delete.go
```diff
@@ -0,0 +1,105 @@
+package batchutils
```
cprice404, line 1: supernit - in future PRs it would be helpful to do a single PR first that contains no functional changes and just does the file re-org. Then file a follow-up PR with the changes. As this stands, I don't know whether there is anything I should be reviewing in this file or not.

If you do end up with PRs like this where you do non-functional re-organizational changes in the same PR as functional changes, then I would suggested adding your on comments to the PR through the github UI, saying "This file contains no functional changes", etc

cprice404: for now I'm going to assume there is nothing I need to look at in this file.

cprice404 (approved):

batchutils/batch_delete.go
pratik151192: Agreed @cprice404 100%. Will keep that in mind

cprice404: thanks!

PR #348, opened 2023-08-30
feat: add timeout to individual requests in batch call

pratik151192:
This PR adds a `request_timeout` for each individual `set`, `get`, or `delete` calls we make within the batch operations. This is just a hardening mechanism such that one if of the go-routines doesn't hang indefinitely and blocks the entire operation.

### DIFF
--- a/batchutils/batch_delete.go
+++ b/batchutils/batch_delete.go
@@ -3,6 +3,7 @@ package batchutils
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/momentohq/client-sdk-go/momento"
 )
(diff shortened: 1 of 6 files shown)


batchutils/batch_delete.go
```diff
@@ -15,16 +16,23 @@ func deleteWorker(
...
 ) {
 	for {
 		myKey := <-keyChan
 		if myKey == nil {
 			return
 		}
-		_, err := client.Delete(ctx, &momento.DeleteRequest{
+
+		timeoutCtx, cancel := context.WithTimeout(ctx, timeout)
+
```
eaddingtonwhite, line 28: Nit: @pratik151192 thoughts on calling 
```go
defer cancel()
``` 
up here on line 28 vs directly after Delete call?

pratik151192: Oh I did that @eaddingtonwhite but IntelliJ complained not to do that because this in a for loop

`Possible resource leak, 'defer' is called in the 'for' loop `

eaddingtonwhite (dismissed):

pratik151192:
@eaddingtonwhite just stupid git rebase stuff

eaddingtonwhite (approved):

PR #351, opened 2023-09-01
fix: return if the context was cancelled while polling for subscription items

pratik151192:
Addresses a bug where a client might have `canceled` the context they used to poll for a topic subscription.

### DIFF
--- a/momento/topic_subscription.go
+++ b/momento/topic_subscription.go
@@ -29,10 +29,31 @@ type topicSubscription struct {
 
 func (s *topicSubscription) Item(ctx context.Context) (TopicValue, error) {
 	for {
+		// Its totally possible a client just calls `cancel` on the `context` immediately after subscribing to an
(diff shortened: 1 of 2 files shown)


eaddingtonwhite:
Looks good maybe add quick test as well for this is good one make sure we cover.

momento/topic_subscription.go
```diff
@@ -29,10 +29,31 @@ type topicSubscription struct {
...
 	for {
+		// Its totally possible a client just calls `cancel` on the `context` immediately after subscribing to an
+		// item, so we should check that here.
+		select {
+		case <-ctx.Done():
+			// Context has been canceled, return an error
+			return nil, ctx.Err()
+		default:
+			// Proceed as is
+		}
```
pratik151192, line 40: The test will fail if you remove this

eaddingtonwhite (approved):

PR #352, opened 2023-09-08
chore: update pubsub example to branch on string vs bytes and rename MOMENTO_AUTH_TOKEN var to MOMENTO_API_KEY

anitarua:
Addresses the golang task in [#452](https://github.com/momentohq/dev-eco-issue-tracker/issues/452) and also renames the environment variable used in the examples from MOMENTO_AUTH_TOKEN to MOMENTO_API_KEY as per recent changes (might not be necessary though, can undo renaming if so).

### DIFF
--- a/examples/pubsub-example/main.go
+++ b/examples/pubsub-example/main.go
@@ -47,12 +47,19 @@ func pollForMessages(ctx context.Context, sub momento.TopicSubscription) {
 		if err != nil {
 			panic(err)
 		}
-		fmt.Printf("received message: '%v'\n", item)
+		switch msg := item.(type) {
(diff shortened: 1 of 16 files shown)


examples/pubsub-example/main.go
```diff
@@ -47,12 +47,19 @@ func pollForMessages(ctx context.Context, sub momento.TopicSubscription) {
 		if err != nil {
 			panic(err)
 		}
-		fmt.Printf("received message: '%v'\n", item)
+		switch msg := item.(type) {
```
anitarua, line 50: anything else needed here besides printing out the messages?

PR #355, opened 2023-09-18
feat: add batch set if not exists to batchutils

anitarua:
Addresses [#463](https://github.com/momentohq/dev-eco-issue-tracker/issues/463) by adding a client-side implementation of `MSETNX`, named `batchSetIfNotExists` here, as part of the `batchutils` package.

Also corrected a line in the makefile that was preventing `make lint` from working properly.

### DIFF
--- a/batchutils/batch_set_if_not_exists.go
+++ b/batchutils/batch_set_if_not_exists.go
@@ -3,6 +3,7 @@ package batchutils
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/momentohq/client-sdk-go/momento"
@@ -48,7 +49,8 @@ func BatchSetIfNotExists(ctx context.Context, props *BatchSetIfNotExistsRequest)
 			}
 		}
 	default:
-		return nil, nil, err
+		var message = fmt.Sprintf("Unexpected KeysExistResponse type: %T\n", resp)
+		return nil, nil, momento.NewMomentoError(momento.ClientSdkError, message, errors.New(message))
 	}
 
 	// If none of the keys exist, set the items using BatchSet
--- a/batchutils/batch_set_if_not_exists.go
+++ b/batchutils/batch_set_if_not_exists.go
@@ -29,15 +29,43 @@ func ExtractKeys(items Items) []momento.Key {
 	return list
 }
 
+// BatchSetIfNotExistsError contains a map associating failing cache keys with their specific errors.
+// Errors may be related to keys already existing or batch set failing.
+// It may be necessary to use a type assertion to access the errors:
+//
+// errors := (err.(*BatchSetIfNotExistsError)).Errors()
+type BatchSetIfNotExistsError struct {
+	errors map[momento.Key]error
+}
+
+func (e *BatchSetIfNotExistsError) Error() string {
+	return "Errors occurred during BatchSetIfNotExists; call Errors() to get a map of key -> errorType"
+}
+
+// Errors contains a map associating errors with their cache keys.
+func (e *BatchSetIfNotExistsError) Errors() map[momento.Key]error {
+	return e.errors
+}
+
+func CreateErrorMapping(keys []momento.Key, err error) map[momento.Key]error {
+	var errors = make(map[momento.Key]error, len(keys))
+	for i := 0; i < len(keys); i++ {
+		errors[keys[i]] = err
+	}
+	return errors
+}
+
 // BatchSetIfNotExists will set the key-value pairs ONLY if all keys don't already exist
-func BatchSetIfNotExists(ctx context.Context, props *BatchSetIfNotExistsRequest) (*BatchSetResponse, error) {
+func BatchSetIfNotExists(ctx context.Context, props *BatchSetIfNotExistsRequest) (*BatchSetResponse, *BatchSetIfNotExistsError) {
 	// First check if all keys exist or not
+	keys := ExtractKeys(props.Items)
 	resp, err := props.Client.KeysExist(ctx, &momento.KeysExistRequest{
 		CacheName: props.CacheName,
-		Keys:      ExtractKeys(props.Items),
+		Keys:      keys,
 	})
 	if err != nil {
-		return nil, err
+		var keysExistsError = &BatchSetIfNotExistsError{errors: CreateErrorMapping(keys, err)}
+		return nil, keysExistsError
 	}
 
 	switch result := resp.(type) {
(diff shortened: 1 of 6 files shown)


batchutils/batch_set_if_not_exists.go
```diff
@@ -0,0 +1,61 @@
...
+	switch result := resp.(type) {
+	case *responses.KeysExistSuccess:
+		// Check if any of the keys already exist
+		for _, keyExists := range result.Exists() {
+			if keyExists {
+				return nil, nil, momento.NewMomentoError(momento.AlreadyExistsError, "At least one key already exists", errors.New("at least one key already exists"))
+			}
+		}
+	default:
+		return nil, nil, err
```
pratik151192, line 51: This is interesting as at this point `err` will also be `nil` because of line 38. This isn't blocking this PR IMO as I see we are using this pattern in other files as well. We should ideally NEVER come here as either `err` will be populated or the `KeysExistSuccess` response type will be matched (unless we add a new response type). I wonder though if these cases are safe to `throw` with a message `IllegalStateError` because we never expect to come here. We bubble up this as `errUnexpectedGrpcResponse` in the SDK which is nice to see handled the way I thought it'd be.

```
func errUnexpectedGrpcResponse(r requester, grpcResp grpcResponse) momentoerrors.MomentoSvcErr {
	return momentoerrors.NewMomentoSvcErr(
		momentoerrors.InternalServerError,
		fmt.Sprintf(
			"%s request got an unexpected response %T '%s'",
			r.requestName(), grpcResp, grpcResp,
		),
		nil,
	)
}
```

I'm approving this PR as if we do decide on this, we should probably sweep through the `batchUtils` atleast and potentially even the SDK files (seems like we are handling on a quick look across SDK files); which i dont feel like is worth clubbing in here as theoretically we should never hit this case if we don't add a new response type. If we do add a new response type, then panicking or erroring here will be beneficial.

pratik151192 (approved):

batchutils/batch_set_if_not_exists.go
cprice404: yeah I think I would `return nil, nil, momento.NewMomentoError(something)` here.

batchutils/batch_set_if_not_exists.go
```diff
@@ -0,0 +1,61 @@
...
+func ExtractKeys(items Items) []momento.Key {
+	var list []momento.Key
+	for _, item := range items {
+		list = append(list, item.Key)
+	}
+	return list
+}
+
+// BatchSetIfNotExists will set the key-value pairs ONLY if all keys don't already exist
+func BatchSetIfNotExists(ctx context.Context, props *BatchSetIfNotExistsRequest) (*BatchSetResponse, *BatchSetError, error) {
```
cprice404, line 32: you two both have more context on this repo than I do, but in a vacuum (without other request functions on the screen to compare to), I don't understand why this signature returns a thruple with two error types as opposed to the more traditional pair with a single slot for the error?

cprice404: i took a look at a few other request types and it looks to me like the signatures are all just returning a pair (`success, error`), or just an error. Can we collapse the two different error types here into one?

anitarua: yeah I wasn't exactly sure how to best approach this. BatchSet returns a set of successful responses and a set of unsuccessful responses for each item in the batch, so I was going to just return that, but then realized BatchSetIfNotExists could also return different errors regarding whether the keys exist or not and whether there was anything wrong with the keysExist call.

anitarua: actually it looks like [Go 1.20 has a way to wrap multiple errors](https://go.dev/doc/go1.20#errors). I'll try to do that

cprice404: @pgautier404 @eaddingtonwhite do either of you have any thoughts on the best way to represent this?

@anitarua I don't know if we can assume our users will be on 1.20 or greater.

eaddingtonwhite: @anitarua @cprice404 I think you would want to make a BatchError interface which implements the built in golang `error` interface and then just return that only. This is approach AWS SDK takes as well for batch packages.  So BatchError would look something like this:

```go
type BatchError interface {
	// Satisfy the generic error interface.
	error
	// List of errors if any returned. Nil is returned if not set.
	BatchErrs() []error
}
```

And function would just look like
```go
func BatchSetIfNotExists(ctx context.Context, props *BatchSetIfNotExistsRequest) (*BatchSetResponse, error) {
  // impl
}
```

where `error` returned always implements `BatchError` so can iterate through errors like:
```go

rsp, err := batchutils.BatchSetIfNotExists(context.TODO(), &batchutils.BatchSetIfNotExistsRequest{})
if err != nil {
    if batchErrGroup, ok := err.(batchutils.BatchError); ok {
        for _, bErr := range batchErrGroup.BatchErrs() {
             fmt.Printf("%+v\n", bErr)
        }
    } else {
        fmt.Println("Error:", err.Error())
    }
}
```

pgautier404: @anitarua, @eaddingtonwhite, and @pgautier404 are brainstorming on this in DMs.

batchutils/batch_set_if_not_exists.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (e *BatchSetIfNotExistsError) Error() string {
+	return "Errors occurred during BatchSetIfNotExists; call Errors() to get a map of key -> errorType"
+}
+
+// Errors contains a map associating errors with their cache keys.
+func (e *BatchSetIfNotExistsError) Errors() map[momento.Key]error {
+	return e.errors
+}
```
anitarua, line 48: Least worst option? @pgautier404 @eaddingtonwhite @cprice404

batchutils/batch_set_if_not_exists.go
```diff
@@ -0,0 +1,98 @@
...
+func BatchSetIfNotExists(ctx context.Context, props *BatchSetIfNotExistsRequest) (*BatchSetResponse, *BatchSetIfNotExistsError) {
+	// First check if all keys exist or not
+	keys := ExtractKeys(props.Items)
+	resp, err := props.Client.KeysExist(ctx, &momento.KeysExistRequest{
+		CacheName: props.CacheName,
+		Keys:      keys,
+	})
+	if err != nil {
+		var keysExistsError = &BatchSetIfNotExistsError{errors: CreateErrorMapping(keys, err)}
+		return nil, keysExistsError
```
cprice404, line 68: Sorry for the churn / delay on this, @anitarua , it's a bit of a tricky API issue :)

After having thought about this more, I think this line is probably the point of contention. If one of the keys exists, I don't think that's an error case. It's just a different flavor of a success case.

So check out the API for SetIfNotExists:

https://docs.momentohq.com/cache/develop/api-reference#setifnotexists

It returns a response type that can either be "Stored", "NotStored", or "Error".

Another decent API to compare to is ItemGetTtl:

https://docs.momentohq.com/cache/develop/api-reference#itemgettype

That one returns hit, miss, or error; and if it's "hit", then it has an `itemType` property on the response.

So I think my suggestion here would be to try to make this API feel as much as possible like the existing `SetIfNotExists` API. That probably starts with changing this signature to return a custom `BatchSetIfNotExistsResponse` rather than the existing `BatchSetResponse`. And then I think the new `BatchSetIfNotExistsResponse` should start off as basically a copy of the existing `SetIfNotExistsResponse`, but then you would change the `Stored` flavor of it so that it had the same data members/accessors that you currently have on `BatchSetResponse`.

I think that makes the API cleaner and more familiar, and I'm hoping that eliminates one of the two error paths that was causing you problems, and then the second error path will be more straightforward. (But I am out of the loop on the convo for what the second error path looks like. I'm hoping it means you can just do the errors the same way they are done in `BatchSet`. But if that's not the case and you need more review/feedback, please ping me and I can dig in a little deeper!)

cprice404 (changes requested):

batchutils/batch_set_if_not_exists.go
anitarua: @cprice404 I think the `Stored` and `NotStored` response types are a good addition!

@pgautier404  pointed out though that there's still some weirdness around how we handle errors from the `KeysExist` call. I think the main question is whether we want to always return a mapping of keys to errors or if we want to return singleton errors for `KeysExist` related errors. Then we could get rid of the `CreateErrorMapping` function I'd made.

The current state of `BatchSetIfNotExists` is as follows:

- `KeysExist` call returns an error -> return `Error` with a map of all keys to the `KeysExist` error
- some keys already exist -> return `NotStored` with no map
- received unexpected `KeysExist` response -> return `Error` with a map of all keys to a new Momento error indicating the unexpected response
- `BatchSet` returns batch of errors -> return `Error` with a map of keys to the `BatchSet` errors
- everything was successful -> return `Stored` with a map of keys to the `BatchSet` responses

cprice404: I think that sounds about as good as we can do here. I'm +1 on what you described above, so feel free to merge/ship once pete approves

pgautier404 (approved):
LGTM!

PR #356, opened 2023-09-21
fix: don't check for connection state when trying to resubscribe

bruuuuuuuce:
The `getConnectivityState` api is an experimental api. A customer got into a state where the connection thought that it was in a correct connected state, but kept erroring out when trying to receive a message from our service.

This change removes the connectivity state check because
- its an experimental api
- there was no pausing between us saying the connectivity state was in the correct state, and us trying to recieve a message and it failing, causing an infinite loop

I have not been able to reproduce the issue that the customer was facing, but based up the feedback they gave and what I see inside of the code, I believe this should fix their issue

### DIFF
--- a/momento/topic_subscription.go
+++ b/momento/topic_subscription.go
@@ -10,7 +10,6 @@ import (
 	pb "github.com/momentohq/client-sdk-go/internal/protos"
 
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/connectivity"
 )
 
 type TopicSubscription interface {
@@ -82,11 +81,6 @@ func (s *topicSubscription) Item(ctx context.Context) (TopicValue, error) {
 }
 
 func (s *topicSubscription) attemptReconnect(ctx context.Context) {
-	// make sure the connection isn't ready, if its ready no need to reconnect
-	if s.topicManager.Conn.GetState() == connectivity.Ready {
-		s.log.Debug("connection is in ready state, not reconnecting")
-		return
-	}
 	// try and reconnect every n seconds. This will attempt to reconnect indefinetly
 	seconds := 5 * time.Second
 	for {


pgautier404 (approved):
LGTM

momento/topic_subscription.go
```diff
@@ -82,11 +81,6 @@ func (s *topicSubscription) Item(ctx context.Context) (TopicValue, error) {
 }
 
 func (s *topicSubscription) attemptReconnect(ctx context.Context) {
-	// make sure the connection isn't ready, if its ready no need to reconnect
-	if s.topicManager.Conn.GetState() == connectivity.Ready {
-		s.log.Debug("connection is in ready state, not reconnecting")
```
cprice404, line 87: i do not understand how this log message was getting spammed when there is a 5 second sleep just below this

bruuuuuuuce: there is a `return` just below it 😄

cprice404: 🤦

cprice404: okay next question, do you know why we added this in the first place? and is there any situation where we could end up calling this method when the connection was healthy, and thus end up unnecessarily reconnecting?

bruuuuuuuce: > is there any situation where we could end up calling this method when the connection was healthy, and thus end up unnecessarily reconnecting?

This was the original intent of why it was added, but this is the exact scenario that happened to the customer. They were in a state where the connection thought it was connected, but kept erroring out with `string=EOF`. I have not been able to replicate what they saw, but 
- this change should prevent the thousands of `connection is in a ready state, not reconnecting` that the customer saw
- `EOF` to me seems like the connection was in fact not in a healthy state, which we should probably try reconnecting on

Again the apis around the `conn` object are all experimental, so its very possible that they were incorrect in stating that they were connected correctly.

cprice404: your first bullet i get, and agree with.

but the other question i'm trying to understand is: under what circumstances do we call this method in the first place

bruuuuuuuce: This method gets called when we receive an error from the stream

```
if err := s.grpcClient.RecvMsg(rawMsg); err != nil {
			s.log.Error("stream disconnected, attempting to reconnect err:", fmt.Sprint(err))

			// Check if the context has been canceled before attempting to reconnect as the client
			// might have given up on the context
			select {
			case <-ctx.Done():
				// Context has been canceled, return an error
				return nil, ctx.Err()
			default:
				// Attempt to reconnect
				s.attemptReconnect(ctx)
			}
```

cprice404 (approved):

PR #360, opened 2023-10-20
fix: use just one grpc connection for topics

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/503

Publish and Subscribe now share grpc connections rather than having one for each.

### DIFF
--- a/internal/grpcmanagers/topic_manager.go
+++ b/internal/grpcmanagers/topic_manager.go
@@ -27,6 +27,7 @@ func NewStreamTopicGrpcManager(request *models.TopicStreamGrpcManagerRequest) (*
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
 		grpc.WithChainStreamInterceptor(interceptor.AddStreamHeaderInterceptor(authToken)),
+		grpc.WithChainUnaryInterceptor(interceptor.AddHeadersInterceptor(authToken)),
 	)
 
 	if err != nil {
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -15,7 +15,6 @@ import (
 
 type pubSubClient struct {
 	streamTopicManagers []*grpcmanagers.TopicGrpcManager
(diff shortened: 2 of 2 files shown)


internal/grpcmanagers/topic_manager.go
```diff
@@ -27,6 +27,7 @@ func NewStreamTopicGrpcManager(request *models.TopicStreamGrpcManagerRequest) (*
 		endpoint,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
 		grpc.WithChainStreamInterceptor(interceptor.AddStreamHeaderInterceptor(authToken)),
+		grpc.WithChainUnaryInterceptor(interceptor.AddHeadersInterceptor(authToken)),
```
anitarua, line 30: adds authorization header to any unary requests made using this connection

momento/pubsub_client.go
```diff
@@ -44,19 +43,11 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
 	}
-
-	unaryDataManager, err := grpcmanagers.NewUnaryDataGrpcManager(&models.DataGrpcManagerRequest{
-		CredentialProvider: request.CredentialProvider,
-		RetryStrategy:      retry.NewNeverRetryStrategy(),
-	})
-	if err != nil {
-		return nil, err
-	}
+	lastStreamTopicManager := streamTopicManagers[numChannels-1]
```
anitarua, line 46: not sure if reserving one specific channel is the best practice, could go the route of the `topicSubscribe` function instead, e.g.
```
topicManager := client.getNextStreamTopicManager()
clientStream, err := topicManager.StreamClient.Publish(...)
```

pgautier404: I like the approach in your comment better. That `lastStreamTopicManager` is still in the rotation for `getNextStreamTopicManager()` anyhow, so there's really no reason to treat it as a special purpose channel.

pgautier404 (approved):
LGTM!

PR #361, opened 2023-10-23
feat: use NumGrpcChannels to configure topic client instead of NumMaxSubscriptions

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/529

### DIFF
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -24,13 +24,20 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
 	var numChannels uint32
 	numSubscriptions := float64(request.TopicsConfiguration.GetMaxSubscriptions())
 	numGrpcChannels := request.TopicsConfiguration.GetNumGrpcChannels()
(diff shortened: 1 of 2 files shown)


momento/pubsub_client.go
```diff
@@ -23,11 +23,14 @@ var streamTopicManagerCount uint64
 func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, momentoerrors.MomentoSvcErr) {
 	var numChannels uint32
 	numSubscriptions := float64(request.TopicsConfiguration.GetMaxSubscriptions())
+	numGrpcChannels := request.TopicsConfiguration.GetNumGrpcChannels()
```
pgautier404, line 26: It seems a little dangerous to allow users to define both of these, as it's not clear that `numSubscriptions` would take precedence without reading the code. I think I'd prefer returning an InvalidArgument error if both are defined (possibly one that mentions `numSubscriptions` is deprecated). @cprice404 wdut?

cprice404 (approved):
lgtm ty

PR #362, opened 2023-10-24
feat: warn users if approaching grpc max concurrent streams limit, add method to close topics subscriptions

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/530

Uses an atomic int to count the number of actively used concurrent grpc streams. Allows us to warn users if they are approaching the grpc max concurrent streams limit given the number of grpc channels they have created. Currently warns users if they have <= 10 streams available given a maximum of 100*numGrpcChannels streams.

- Increment before calling `Subscribe`
- Decrement if `Subscribe` returned an error
- Increment before calling `Publish`
- Decrement after calling `Publish`
- Decrement after calling `Close` on a `TopicSubscription`
- Set number of streams to 0 when `Close` is called

### DIFF
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -70,14 +70,16 @@ func (client *pubSubClient) getNextStreamTopicManager() *grpcmanagers.TopicGrpcM
 	return topicManager
 }
 
-func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (*grpcmanagers.TopicGrpcManager, grpc.ClientStream, error) {
+func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (*grpcmanagers.TopicGrpcManager, grpc.ClientStream, context.Context, context.CancelFunc, error) {
 
 	checkNumConcurrentStreams(client.log)
 
-	atomic.AddInt64(&numGrpcStreams, 1)
+	// try withCancel on context
+	cancelContext, cancelFunction := context.WithCancel(ctx)
 
+	atomic.AddInt64(&numGrpcStreams, 1)
 	topicManager := client.getNextStreamTopicManager()
-	clientStream, err := topicManager.StreamClient.Subscribe(ctx, &pb.XSubscriptionRequest{
+	clientStream, err := topicManager.StreamClient.Subscribe(cancelContext, &pb.XSubscriptionRequest{
 		CacheName:                   request.CacheName,
 		Topic:                       request.TopicName,
 		ResumeAtTopicSequenceNumber: request.ResumeAtTopicSequenceNumber,
@@ -85,14 +87,15 @@ func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSu
 
 	if err != nil {
 		atomic.AddInt64(&numGrpcStreams, -1)
-		return nil, nil, err
+		cancelFunction()
+		return nil, nil, nil, nil, err
 	}
 
 	if numGrpcStreams > 0 && (int64(numChannels*100)-numGrpcStreams < 10) {
 		client.log.Info(fmt.Sprintf("WARNING: approaching grpc maximum concurrent stream limit, %d remaining of total %d streams\n", int64(numChannels*100)-numGrpcStreams, numChannels*100))
 	}
 
-	return topicManager, clientStream, err
+	return topicManager, clientStream, cancelContext, cancelFunction, err
 }
 
 func (client *pubSubClient) topicPublish(ctx context.Context, request *TopicPublishRequest) error {
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -92,7 +92,7 @@ func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSu
 	}
 
 	if numGrpcStreams > 0 && (int64(numChannels*100)-numGrpcStreams < 10) {
-		client.log.Info(fmt.Sprintf("WARNING: approaching grpc maximum concurrent stream limit, %d remaining of total %d streams\n", int64(numChannels*100)-numGrpcStreams, numChannels*100))
+		client.log.Warn(fmt.Sprintf("WARNING: approaching grpc maximum concurrent stream limit, %d remaining of total %d streams\n", int64(numChannels*100)-numGrpcStreams, numChannels*100))
 	}
 
 	return topicManager, clientStream, cancelContext, cancelFunction, err
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -89,7 +89,7 @@ func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSu
 	}
 
 	if numGrpcStreams > 0 && (int64(numChannels*100)-numGrpcStreams < 10) {
-		fmt.Printf("WARNING: approaching grpc maximum concurrent stream limit, %d remaining of total %d streams\n", int64(numChannels*100)-numGrpcStreams, numChannels*100)
+		client.log.Info(fmt.Sprintf("WARNING: approaching grpc maximum concurrent stream limit, %d remaining of total %d streams\n", int64(numChannels*100)-numGrpcStreams, numChannels*100))
 	}
 
 	return topicManager, clientStream, err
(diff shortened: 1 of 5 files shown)


momento/pubsub_client.go
```diff
@@ -109,7 +139,15 @@ func (client *pubSubClient) topicPublish(ctx context.Context, request *TopicPubl
...
 
 func (client *pubSubClient) close() {
+	atomic.AddInt64(&numGrpcStreams, (0 - numGrpcStreams))
 	for clientIndex := range client.streamTopicManagers {
 		defer client.streamTopicManagers[clientIndex].Close()
 	}
 }
+
+func checkNumConcurrentStreams() error {
+	if numGrpcStreams > 0 && numGrpcStreams == int64(numChannels*100) {
```
cprice404, line 149: I think I would tweak this slightly:

* check for >= instead of ==
* just log a warning if the limit is exceeded, don't error out
* return void

wdyt @pgautier404 ?

momento/pubsub_client.go
```diff
@@ -60,22 +62,47 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
 	return topicManager, clientStream, err
 }
 
 func (client *pubSubClient) topicPublish(ctx context.Context, request *TopicPublishRequest) error {
+	err := checkNumConcurrentStreams()
+	if err != nil {
+		return err
+	}
+
+	atomic.AddInt64(&numGrpcStreams, 1)
```
cprice404, line 104: at the risk of being a little less DRY, I think I'd recommend copy/pasting this line directly before and after each `Publish` call. you really want to minimize the number of LOC between the increment and decrement to make it easier to reason about this sort of thing.

in other languages you'd need to worry about the possibility of an exception being thrown on any of the lines between this and the publish. Less so in golang, but still a little more burden for future devs to have it separated.

You could also introduce a small wrapper function for the Publish calls if you wanted to encapsulate the inc, publish, and dec to a single spot in the code.

momento/pubsub_client.go
```diff
@@ -99,8 +127,10 @@ func (client *pubSubClient) topicPublish(ctx context.Context, request *TopicPubl
 				},
 			},
 		})
+		atomic.AddInt64(&numGrpcStreams, -1)
 		return err
 	default:
+		atomic.AddInt64(&numGrpcStreams, -1)
```
cprice404, line 133: this one won't be necessary if you move the increment like i suggested above, which is a nice bonus.

momento/pubsub_client.go
```diff
@@ -60,22 +62,47 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
 }
 
 func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (*grpcmanagers.TopicGrpcManager, grpc.ClientStream, error) {
+
+	err := checkNumConcurrentStreams()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	atomic.AddInt64(&numGrpcStreams, 1)
```
cprice404, line 77: I'm less familiar with the go SDK, but in the other SDKs when you call `subscribe`, the object that is returned to you has a method you can call to close/cancel the subscription. I assume we have that in golang too.

If that's true, I'm not immediately seeing where the corresponding decrement is that should happen in that case? maybe i'm just missing it though!

momento/pubsub_client.go
pgautier404: That sounds right to me. I don't think we're in a place where we can confidently throttle connections based on the counter.

momento/pubsub_client.go
```diff
@@ -109,7 +139,15 @@ func (client *pubSubClient) topicPublish(ctx context.Context, request *TopicPubl
 }
 
 func (client *pubSubClient) close() {
+	atomic.AddInt64(&numGrpcStreams, (0 - numGrpcStreams))
```
pgautier404, line 142:
```suggestion
	atomic.AddInt64(&numGrpcStreams, -numGrpcStreams)
```

momento/pubsub_client.go
```diff
@@ -60,22 +62,47 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
 		ResumeAtTopicSequenceNumber: request.ResumeAtTopicSequenceNumber,
 	})
+
+	if err != nil {
+		atomic.AddInt64(&numGrpcStreams, -1)
+		return nil, nil, err
+	}
+
+	if numGrpcStreams > 0 && (int64(numChannels*100)-numGrpcStreams < 10) {
+		fmt.Printf("WARNING: approaching grpc maximum concurrent stream limit, %d remaining of total %d streams\n", int64(numChannels*100)-numGrpcStreams, numChannels*100)
```
pgautier404, line 92: This should probably go to a log instead of to the console. It doesn't look like the pubSubClient struct has a logger property, so you'll need to add one in the struct and use the logger factory in `request.TopicsConfiguration` to create a logger to assign to it.

momento/pubsub_client.go
anitarua: I think you're right, but I don't think the Go SDK currently allows users to close a subscription from what I see. 

- `TopicClient.Subscribe()` calls `pubSubClient.topicSubscribe()` to get a `topicManager` and `clientStream`, which seem to have generated functions for closing a grpc connection or stream, respectively
- `TopicClient.Subscribe()` returns a `TopicSubscription` containing `topicManager` and `clientStream`
- The `TopicSubscription` interface allows access only to the `Item()` function, so a user wouldn't be able to access `topicManager.Conn.Close()` or `clientStream.Context().Done()` to close a connection or stream

Should the SDK provide a `Close()` or `Cancel()` function for users to explicitly unsubscribe and we decrement there? Otherwise, I think `TopicClient.Close()` is the only way subscriptions/streams are closed (as a result of closing the grpc connections).

anitarua: @eaddingtonwhite Just a heads up that we'll be adding the missing `Close` function to the `TopicSubscription` interface to allow users to close a topic subscription when they're done.

eaddingtonwhite: 👍

cprice404 (commented):
most recent commits look good, i think we are still waiting on a few more though?

momento/pubsub_client.go
```diff
@@ -56,29 +60,51 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
 	return topicManager
 }
 
-func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (*grpcmanagers.TopicGrpcManager, grpc.ClientStream, error) {
+func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSubscribeRequest) (*grpcmanagers.TopicGrpcManager, grpc.ClientStream, context.Context, context.CancelFunc, error) {
+
+	checkNumConcurrentStreams(client.log)
+
+	// try withCancel on context
+	cancelContext, cancelFunction := context.WithCancel(ctx)
```
anitarua, line 78: Did more digging and discovered `WithCancel` and this was the approach that worked.

momento/topic_subscription.go
```diff
@@ -98,7 +105,14 @@ func (s *topicSubscription) attemptReconnect(ctx context.Context) {
...
 			s.topicManager = newTopicManager
 			s.grpcClient = newStream
+			s.cancelContext = cancelContext
+			s.cancelFunction = cancelFunction
 			return
 		}
 	}
 }
+
+func (s *topicSubscription) Close() {
```
cprice404, line 115: in javascript we called this `unsubscribe`. In .NET we called it `Dispose` (because that is part of their `IDisposable` interface / pattern). If go has a similar `Closeable` pattern then we should use whatever name is idiomatic; if not, I'd go with `Unsubscribe` since that's what we did in JS.

momento/pubsub_client.go
cprice404: nice

momento/pubsub_client.go
```diff
@@ -56,29 +60,51 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
+
+	atomic.AddInt64(&numGrpcStreams, 1)
 	topicManager := client.getNextStreamTopicManager()
-	clientStream, err := topicManager.StreamClient.Subscribe(ctx, &pb.XSubscriptionRequest{
+	clientStream, err := topicManager.StreamClient.Subscribe(cancelContext, &pb.XSubscriptionRequest{
```
cprice404, line 82: i see, this makes sense

cprice404 (commented):
i have one question / nit about the name of the new public function, otherwise I think this LGTM now!

Would love to get Ellery and/or Pete to review the `Context` stuff since they both have more experience in golang than I do.

momento/topic_subscription.go
anitarua: A pattern like this? https://pkg.go.dev/v2ray.com/core/common#Closable

cprice404: perfect, ty!

and IIRC in golang you don't have to specify that you implement an interface, it just figures it out, is that right? just asking whether our our TopicSubscription interface needs to be marked as `Closeable`.

anitarua: The Go guide states that interfaces are implemented implicitly: https://go.dev/tour/methods/10

cprice404: great. thanks!

cprice404 (approved):

pgautier404 (approved):
Nice, LGTM

PR #366, opened 2023-11-17
fix: Update on-pull-request.yml

bruuuuuuuce:
Previously all builds were failing. This fixes and standardizes our actions with our other sdk repositories


.github/workflows/manual_release.yml
```diff
@@ -21,13 +21,53 @@ jobs:
 
       - name: Output release
         id: release
-        run: echo "::set-output name=release::${{ steps.semrel.outputs.version }}"
+        run: echo "release=${{ steps.semrel.outputs.version }}" >> $GITHUB_OUTPUT
```
bruuuuuuuce, line 24: set output is deprecated, this is same change we made to our javascript pr https://github.com/momentohq/client-sdk-javascript/pull/1022/files

anitarua (approved):
thanks for fixing it!!

PR #368, opened 2023-12-05
feat: add disposable tokens apis to golang sdk

bruuuuuuuce:


momento/disposable_token_scopes.go
```diff
@@ -0,0 +1,55 @@
...
+	return Permissions{
+		Permissions: []Permission{TopicPermission{Cache: selector, Topic: TopicNamePrefix{NamePrefix: topicNamePrefix}, Role: PublishOnly}},
+	}
+}
+
+func TopicNamePrefixSubscribeOnly(selector CacheSelector, topicNamePrefix string) DisposableTokenScope {
+	return Permissions{
+		Permissions: []Permission{TopicPermission{Cache: selector, Topic: TopicNamePrefix{NamePrefix: topicNamePrefix}, Role: SubscribeOnly}},
+	}
+}
```
pgautier404, line 55: I think we need `TopicName` versions of these as well?

pgautier404:
It looks like we're missing test coverage for all permissions scopes except `TopicPublishOnly`, tests for custom token scopes manually composed of permissions, and a test for `AllDataReadWrite. Not necessarily a blocker for this PR though if you could add a ticket for additional testing.

momento/disposable_token_scopes.go
bruuuuuuuce: I think these cover that case https://github.com/momentohq/client-sdk-go/pull/368/files#diff-7c0d5fa617251d8b0144289bc36396bb07a71549d6b265feaa84b7427f74fc05R15-R25?

bruuuuuuuce:
> It looks like we're missing test coverage for all permissions scopes except `TopicPublishOnly`

I think the rest of these are at the end of the test file https://github.com/momentohq/client-sdk-go/pull/368/files#diff-a51994a283c26d43a0949bfa49bcf706a7b5741d973ec01926eadca7bc89b977R1208, there are lots of permissions tests 😅

pgautier404:
> > It looks like we're missing test coverage for all permissions scopes except `TopicPublishOnly`
> 
> I think the rest of these are at the end of the test file https://github.com/momentohq/client-sdk-go/pull/368/files#diff-a51994a283c26d43a0949bfa49bcf706a7b5741d973ec01926eadca7bc89b977R1208, there are lots of permissions tests 😅

Ah, geez, yeah 🤦 I needed to pull.

momento/disposable_token_scopes.go
pgautier404: Ah, gotcha, great!

pgautier404 (approved):

anitarua (approved):
lgtm!

momento/permissions_scopes.go
```diff
@@ -0,0 +1,25 @@
+package momento
+
+func CacheReadWrite(selector CacheSelector) Permissions {
```
cprice404, line 3: This isn't necessarily a blocker, but in the other languages I feel like the 90% use case for these is to pass Strings. I know you can't do function overloading in golang but what does it actually look like to call this if you just want to specify a cache name? Can you think of any sane/user-friendly way to support either CacheSelector or string? like maybe if we added one set of functions for each?

momento/permission_scope.go
```diff
@@ -0,0 +1,114 @@
...
+}
+func (t TopicNamePrefix) TopicName() string {
+	return t.NamePrefix
+}
+
+type TopicPermission struct {
+	Role  TopicRole
+	Cache CacheSelector
+	Topic TopicSelector
+}
```
cprice404, line 91: same comment here, in the other languages there's usually a way to create one of these with strings. wondering what your thoughts are.

i know it's kind of a PITA to add examples before a release goes out but it'd be super helpful to see a bit of example code showing:

1. what it looks like for users to construct a scope via one of the predefined things like CacheReadWrite
2. what it looks like for users to construct a scope by explicitly listing permissions

i will look in the tests for a sec and see if i see what i'm looking for

momento/permissions_scopes.go
bruuuuuuuce: It would look something like this

```
CacheWriteOnly(CacheName{Name: "my cache name"})

CacheReadWrite(AllCaches{})

TopicSubscribeOnly(CacheName{Name: "a cache"}, TopicName{Name: "a topic"})
```

The tests have pretty good illustrations of what it looks like to call. The most sane way I can think of is adding duplicate methods since union types dont exist in golang

momento/permission_scope.go
bruuuuuuuce: I think the tests that I wrote for this feature show a pretty good example of what it would look like for a user to create permissions for predefined scopes. As for what it would look like to explicitly list out permissions, it would look something like what the predefined permissions are doing

```
Permissions{
		Permissions: []Permission{DisposableTokenCachePermission{Cache: selector, Item: CacheItemKeyPrefix{KeyPrefix: keyPrefix.asBytes()}, Role: WriteOnly}},
	}
```

cprice404 (commented):
overall this looks great and I am okay with going with the current API if the team supports it. before we merge I'd just like us to take one more moment to think about whether we care enough about a more succinct representation for specifying cache/topic selectors via simple String objects as per the comments I left. If we align that it is not worth changing things for that then I will be 👍 on the PR.

Interested in quick input on that topic from @bruuuuuuuce @pgautier404 and @eaddingtonwhite in particular.

momento/permissions_scopes.go
bruuuuuuuce: > Because of the fact that golang doesn't support overloading, then if we ship this PR as is, we can't support that ^^ in the future.

Thats true, but we could have something like `CacheNameWriteOnly("mycache")`

cprice404: > Thats true, but we could have something like `CacheNameWriteOnly("mycache")`

I like that idea and that's almost compelling enough to convince me, but, what about for like a TopicPublishOnly?

bruuuuuuuce: `TopicNamePublishOnlyCacheName` lol, my point is more that its forward compatible with adding strings

cprice404: taking to slack/zoom to speed up resolution :)

cprice404 (approved):
oh my god

PR #375, opened 2024-02-06
feat: add number of grpc channels as cache client config

anitarua:
Config allows creation of `numGrpcChannels` number of data clients in the CacheClient.

Did local testing and observed in the default case, there's 3 TCP connections (1 control client, 1 data client, 1 ping client). 
When using `WithNumGrpcChannels(4)`, there are 6 TCP connections (1 control client, 4 data client, 1 ping client).


pgautier404 (approved):
🚢

PR #378, opened 2024-02-14
feat: add read concern header

bruuuuuuuce:


cprice404 (approved):
Changes look good to me; new config API surface area looks good to me.

As far as I'm concerned we can time box the smoke testing on this and if we can't come up with a semi-reliable repro then we can go ahead and merge and release this.

momento/scalar_test.go
```diff
@@ -69,6 +69,8 @@ var _ = Describe("Scalar methods", func() {
 		Entry("with default cache name when the key and value are bytes", WithDefaultCache, Bytes([]byte{1, 2, 3}), Bytes("string"), "string", []byte("string")),
 		Entry("with default cache name when the value is empty", WithDefaultCache, String("key"), String(""), "", []byte("")),
 		Entry("with default cache name when the value is blank", WithDefaultCache, String("key"), String("  "), "  ", []byte("  ")),
+		Entry("using a consistent read concern client", WithConsistentReadConcern, String("key"), String("value"), "value", []byte("value")),
+		Entry("using a balanced read concern client", WithExpressReadConcern, Bytes([]byte{1, 2, 3}), Bytes("string"), "string", []byte("string")),
```
malandis, line 73: Typo? Documentation string says "using a balanced read concern" but the next parameter is Express. Also, did you mean to test all three or just two here?

malandis (commented):
One comment about a test case.

Separately: are we going to document the different read concerns in the client for the user in this PR? Later? Or will that be in the dev docs only?

momento/scalar_test.go
bruuuuuuuce: > are we going to document the different read concerns in the client for the user this PR

Yes good call. I meant to add a comment but was going to confirm the exact difference between the 3 first

> Typo

Yea...

> Also, did you mean to test all three or just two here

Good point. The default is `balanced` but doesnt hurt to add a third test with the default client

malandis (approved):

internal/interceptor/authorization_interceptor.go
```diff
@@ -3,16 +3,24 @@ package interceptor
...
-func AddHeadersInterceptor(authToken string) func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+func AddAuthHeadersInterceptor(authToken string) func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		return invoker(metadata.AppendToOutgoingContext(ctx, "authorization", authToken), method, req, reply, cc, opts...)
 	}
 }
 
+func AddReadConcernHeaderInterceptor(readConcern config.ReadConcern) func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
+		return invoker(metadata.AppendToOutgoingContext(ctx, "read_concern", string(readConcern)), method, req, reply, cc, opts...)
```
kvcache, line 20: This is not the correct header name. The correct header name is `read-concern`.

kvcache (changes requested):
need to use the correct header name

internal/interceptor/authorization_interceptor.go
bruuuuuuuce, line 20:
```suggestion
		return invoker(metadata.AppendToOutgoingContext(ctx, "read-concern", string(readConcern)), method, req, reply, cc, opts...)
```

PR #380, opened 2024-02-23
feat: add grpc config options and turn off keepalive for Lambda config

anitarua:
Adds GRPC configs for max message size and keepalive options. Those are passed in as `DialOptions` when constructing the GRPC connection in the `*GrpcManager` functions. The keepalive options are set only if nonzero values are passed in.

Adjusts the prebuilt configuration options to enable keepalives. 
The Lambda prebuilt config remains the same with keepalives disabled.

The `AuthClient` and `TopicClient` configs don't specify a `TransportStrategy`, so their grpc managers will just use the default settings (max message size 5mb, keepalives turned on with same values as in the cache configs). The data and ping grpc managers use the options set in the cache config objects. The control manager disables keepalives.


config/grpc-configuration.go
```diff
@@ -6,15 +6,86 @@ type GrpcConfigurationProps struct {
...
 	deadline time.Duration
+
+	// Indicates if it permissible to send keepalive pings from the client without any outstanding streams.
+	//
+	// NOTE: keep-alives are very important for long-lived server environments where there may be periods of time
+	// when the connection is idle. However, they are very problematic for lambda environments where the lambda
+	// runtime is continuously frozen and unfrozen, because the lambda may be frozen before the "ACK" is received
+	// from the server. This can cause the keep-alive to timeout even though the connection is completely healthy.
+	// Therefore, keep-alives should be disabled in lambda and similar environments.
+	keepAlivePermitWithoutCalls bool
```
pratik151192, line 17: should this and other `keepAlive` settings be a pointer like `* keepAlivePermitWithoutCalls`. GoLang default int value i believe is 0, so am wondering if we don't make it a pointer, and don't do a `nil` check on `grpc-channel-option` while setting these, they all become 0? And if that's case, not sure if the keepAlive will be turned on with 0 ms and set this param to false.

anitarua: Good call, I can switch them to pointers instead to be able to do the nil check

anitarua: I wasn't quite able to get the nil checks in without breaking the way GrpcConfiguration is specified in the prebuilt configs, but I added a check so that we don't pass in the keepalive settings unless `keepAlivePermitWithoutCalls` is specifically set to `true`

cprice404: > I wasn't quite able to get the nil checks in without breaking the way GrpcConfiguration is specified in the prebuilt config

can you say more about this? these are all new properties so I would have expected you'd just add them to the list of args to the struct constructor calls in the pre-built configs, but with `nil` values?

internal/models/requests.go
```diff
@@ -13,28 +13,34 @@ import (
...
 }
 
 type TopicStreamGrpcManagerRequest struct {
 	CredentialProvider auth.CredentialProvider
+	GrpcConfiguration  config.GrpcConfiguration
 }
 
 type PingGrpcManagerRequest struct {
 	CredentialProvider auth.CredentialProvider
+	GrpcConfiguration  config.GrpcConfiguration
```
cprice404, line 43: i'm not super familiar with this SDK but I'm mildly surprised to see these here. I would have guessed this would be a property on a client object somewhere, passed in at construction time, and that as such it wouldn't need to be part of the structs for the request objects?

config/configurations.go
```diff
@@ -43,7 +46,10 @@ func InRegionLatestWithLogger(loggerFactory logger.MomentoLoggerFactory) Configu
 		LoggerFactory: loggerFactory,
 		TransportStrategy: NewStaticTransportStrategy(&TransportStrategyProps{
 			GrpcConfiguration: NewStaticGrpcConfiguration(&GrpcConfigurationProps{
-				deadline: 1100 * time.Millisecond,
+				deadline:                    1100 * time.Millisecond,
+				keepAlivePermitWithoutCalls: true,
+				keepAliveTime:               5000 * time.Millisecond,
+				keepAliveTimeout:            1000 * time.Millisecond,
```
cprice404, line 52: If I'm reading this PR correctly, we weren't doing any keepalives in this SDK before. Is that correct? Do we know what the default values are if we don't specify them?

I'm asking because I want to make sure we are aware if we are changing the behavior here vs. previous releases of the SDKs.

I think that based on what we've seen so far, having keepalives enabled for long-lived/server-side apps is the right thing to do, and having them disabled for lambdas is the right thing to do, and I think that's what you've done here. But if they weren't previously enabled for the server-side cases then we just need to think a bit about whether there are any possible impacts on existing users; we do have some users doing various types of load tests with this SDK today. Can discuss on a zoom or something if there are questions to work out here.

config/grpc-configuration.go
anitarua: I can pass in nil values into the struct if I change `GrpcConfigurationProps` to accept pointers, but that would mean I can't pass in constants or literal values, only references to defined variables. If that's an acceptable switch to make, I can do that.

So the struct constructors would look like:
```
var NON_LAMBDA_KEEPALIVE_WITHOUT_CALLS = true
var NON_LAMBDA_KEEPALIVE_TIME = 5000 * time.Millisecond
var NON_LAMBDA_KEEPALIVE_TIMEOUT = 1000 * time.Millisecond

...
        GrpcConfiguration: NewStaticGrpcConfiguration(&GrpcConfigurationProps{
		deadline:                    5 * time.Second,
		keepAlivePermitWithoutCalls: &NON_LAMBDA_KEEPALIVE_WITHOUT_CALLS,
		keepAliveTime:               &NON_LAMBDA_KEEPALIVE_TIME,
		keepAliveTimeout:            &NON_LAMBDA_KEEPALIVE_TIMEOUT,
	}),
```

Instead of what it's doing currently:
```
...
        GrpcConfiguration: NewStaticGrpcConfiguration(&GrpcConfigurationProps{
		deadline:                    1100 * time.Millisecond,
		keepAlivePermitWithoutCalls: true,
		keepAliveTime:               5000 * time.Millisecond,
		keepAliveTimeout:            1000 * time.Millisecond,
	}),
```

cprice404: oh, geez. yeah i forgot about that. and I'm guessing you can't do `&(5000 * time.Millisecond)` either, right?

I think I am on the same page as you are now and would prefer to allow the literals to be passed in. But I don't know that we want to condition all of the keepalive flags on `keepAlivePermitWithoutCalls`, because that can legitimately be set to `false` in cases where you really do want the other two values to be honored.

What is the case where zero is ambiguous here?

config/configurations.go
cprice404: I think we probably want them enabled by default, I think the direction the PR is headed is good, but I just want to check in with stakeholders before we make the switch. I'll start a thread in slack.

Also - do we have `V1`s for these configs? if so we could ask them if they're currently using v1, and if they are then introducing v2 and updating Latest would be safe.

anitarua: There are no configs labeled as V1. There are:
* LaptopLatest
* LaptopLatestWithLogger
* InRegionLatest
* InRegionLatestWithLogger
* Lambda
* LambdaWithLogger

config/grpc-configuration.go
anitarua: Yeah `&(5000 * time.Millisecond)` was not accepted either.

And yeah, that's true, each keepalive option should be able to be set separately. If they're omitted from the constructor, they would default to:
* `keepAlivePermitWithoutCalls` -> false (0) -> we won't know if user specifically set it to false or if it was false by default
* `keepAliveTime` -> 0s but grpc will set any value <10s to 10s, but if it's not set, the default is infinity
* `keepAliveTimeout` -> 0s instead of the grpc default of 20s

internal/models/requests.go
anitarua: The `*GrpcManagerRequest` structs appear to just contain the arguments passed into the `New*GrpcManager` functions that construct the grpc connection. I wasn't sure how else to pass the configs through to the grpc manager functions.

config/grpc-configuration.go
cprice404: this sounds good to me, thanks!

internal/models/requests.go
cprice404: i see. okay!

cprice404 (approved):
I'm good with this! would be good to get pratik to sign off too though!

anitarua:
Okay, I think I will actually be making one more commit to this PR since we aligned on centralizing the default configs and making sure all clients go through the same grpc channel init process
cc: @pratik151192

config/configurations.go
```diff
@@ -31,11 +33,30 @@ func LaptopLatestWithLogger(loggerFactory logger.MomentoLoggerFactory) Configura
 	})
 }
 
+func LaptopWithLoggerAndKeepalive(loggerFactory logger.MomentoLoggerFactory) Configuration {
```
cprice404, line 36: I don't think we need a top-level public function for this? And also since you set the defaults in the grpc-channel-options.go now I think these pre-builts can just rely on that, right? Like, in the `LaptopLatestWithLogger` function we can do something like:

```
config := defaultStaticGrpcConfiguration();
config.deadline = ...
```

and then pass that in to the NewStaticTransportStrategy constructor or something?

config/configurations.go
cprice404: it's okay to change the semantics, we confirmed it w/rt that existing user.

Also, any config that has the word `Latest` in it, we are advertising to users that they should be aware that we might change it over time, so it's okay.

Also, all of our pre-built configs should either be `V1` or `V2` or `Latest`. If we have one that is just `Lambda` without `Latest`, that's a bug, and we should go ahead and fix it while you're in here.

config/configurations.go
```diff
@@ -54,18 +58,23 @@ func InRegionLatestWithLogger(loggerFactory logger.MomentoLoggerFactory) Configu
...
 
-func LambdaWithLogger(loggerFactory logger.MomentoLoggerFactory) Configuration {
+func LambdaLatestWithLogger(loggerFactory logger.MomentoLoggerFactory) Configuration {
 	return NewCacheConfiguration(&ConfigurationProps{
 		LoggerFactory: loggerFactory,
 		TransportStrategy: NewStaticTransportStrategy(&TransportStrategyProps{
 			GrpcConfiguration: NewStaticGrpcConfiguration(&GrpcConfigurationProps{
 				deadline: 1100 * time.Millisecond,
-			}),
+			}).WithKeepAliveDisabled(),
```
anitarua, line 77: by default, `NewStaticGrpcConfiguration` will enable keepalives, but this copy constructor will override them and disable keepalives

config/transport_strategy.go
```diff
@@ -25,23 +27,126 @@ type TransportStrategy interface {
...
+	maxSendMessageLength        int
+	maxReceiveMessageLength     int
 }
 
+// Constructs new GrpcConfiguration to tune lower-level grpc settings.
+// Note: keepalive settings are enabled by default, use WithKeepAliveDisabled() to disable all of them,
+// or use the appropriate copy constructor to override individual settings.
 func NewStaticGrpcConfiguration(grpcConfiguration *GrpcConfigurationProps) *StaticGrpcConfiguration {
+	// We set keepalive values to defaults because we can't tell if users set them to zero values
+	// or if the settings were just omitted from the props struct, thus defaulting them to zero values.
```
anitarua, line 52: some quirkiness here because Go sets the omitted fields of the GrpcConfigurationProps to their zero values, so we won't know if a user set the keepalive settings to zeros or if they just happened to default to their zero values

I made copy constructors to override the keepalive settings after the initial StaticGrpcConfiguration object is constructed

internal/grpcmanagers/control_manager.go
```diff
@@ -19,15 +18,18 @@ type ScsControlGrpcManager struct {
...
 
 func NewScsControlGrpcManager(request *models.ControlGrpcManagerRequest) (*ScsControlGrpcManager, momentoerrors.MomentoSvcErr) {
-	config := &tls.Config{
-		InsecureSkipVerify: false,
-	}
 	authToken := request.CredentialProvider.GetAuthToken()
 	endpoint := fmt.Sprint(request.CredentialProvider.GetControlEndpoint(), ControlPort)
+
+	// Override grpc config to disable keepalives
+	controlConfig := config.NewStaticGrpcConfiguration(&config.GrpcConfigurationProps{}).WithKeepAliveDisabled()
```
anitarua, line 25: turns off keepalives for the control client to match the other sdks

internal/grpcmanagers/grpc-channel-options.go
```diff
@@ -0,0 +1,41 @@
...
+func GrpcChannelOptionsFromGrpcConfig(grpcConfig config.GrpcConfiguration) []grpc.DialOption {
+	// Default to 5mb message sizes and keepalives turned on (defaults are set in NewStaticGrpcConfiguration)
+	return []grpc.DialOption{
+		grpc.WithDefaultCallOptions(
+			grpc.MaxCallRecvMsgSize(grpcConfig.GetMaxReceiveMessageLength()),
+			grpc.MaxCallSendMsgSize(grpcConfig.GetMaxSendMessageLength()),
+		),
+		grpc.WithKeepaliveParams(
+			keepalive.ClientParameters{
+				PermitWithoutStream: grpcConfig.GetKeepAlivePermitWithoutCalls(),
```
cprice404, line 22: So this is the only place in this whole PR where we are actually translating from our bespoke config model to the actual grpc library config model, right?

If so - do we need to condition these so that we only set them if our bespoke values are not zero? In other words, in the underlying grpc library code, is there a difference between setting these to zero and not setting them at all? or are they treating zeroes as "unset", like we are?

cprice404 (commented):
overall i think this looks good but just had one question.

And I'd definitely like to get two approvals on this before it goes in just because it's a decent amount of new API surface area.

Thanks for keeping at it! I know this has been a tricky one.

internal/grpcmanagers/grpc-channel-options.go
anitarua: No, there is no difference.
I got stuck on this question for a while but eventually put in some print statements and saw that instantiating `keepalive.ClientParameters` but not setting the values was equivalent to creating that struct and setting them each to their zero values. I think the grpc library treats the zeros as "unset" like we're doing.

config/transport_strategy.go
```diff
@@ -25,23 +27,126 @@ type TransportStrategy interface {
...
-	deadline time.Duration
+	deadline                    time.Duration
+	keepAlivePermitWithoutCalls bool
+	keepAliveTimeout            time.Duration
+	keepAliveTime               time.Duration
+	maxSendMessageLength        int
+	maxReceiveMessageLength     int
 }
 
+// Constructs new GrpcConfiguration to tune lower-level grpc settings.
```
bruuuuuuuce, line 47:
```suggestion
// NewStaticGrpcConfiguration constructs a new GrpcConfiguration to tune lower-level grpc settings.
```

config/transport_strategy.go
bruuuuuuuce: image

cprice404 (approved):
ship

PR #388, opened 2024-03-07
chore: by default support lazy connections

bruuuuuuuce:
when using the `Dial` method inside of grpc-go, the client automatically tries to connect to the server. This is not the default behavior that we want, and is not the default behavior of other grpc language impl. 

https://github.com/grpc/grpc-go/blob/99ded5c4905feb5d6ef305114c6f8a7a942011c7/clientconn.go#L222-L225

This pr addresses this by calling `NewClient`, which does not try to automatically connect to the server, unless an `eagerConnection` is specified. If the user wants to eagerly connect (which is currently only an option for data plane client), then the behavior is unchanged


cprice404 (approved):
:ty:

PR #390, opened 2024-03-08
fix: race condition that can cause batchutils to leak goroutines

cprice404:
The batchutils use a goroutine to distribute items/keys to the
workers for batchget, batchset, batchdelete.

In these goroutines, we have a `select` block in a `for` loop,
which is blocking waiting until the context is closed. However,
in the `default` case of that select block, we try to push a
`nil` onto the channel. This is intended to be a signaling mechanism
to tell the workers that they can exit.

However, this `default` branch of the `select` may be reached
many times between when the workers complete and when the context
is cancelled. If they hit it enough times, they will fill up the
channel to its maximum size, and enter into a blocking state
where they are trying to push another `nil` onto the channel even
after all of the workers have exited. This push will never succeed
because the channel is full and no more workers are in existence
to empty the channel.

In this scenario, even though the context has been cancelled, the
goroutine will never exit the default branch of the select so that
it can execute the next iteration of the for loop, where it would
notice the cancellation and exit. Thus, the goroutine remains
running forever and leaks memory for all of the objects that it
has a reference to.

In this commit we get rid of the `default` branch of the `select`,
and instead, we simply require the caller to pass in the number of
workers as an argument. This allows us to push exactly one `nil`
onto the channel for each worker, thus preventing the race condition
and the goroutine leak.

### DIFF
--- a/batchutils/batch_delete.go
+++ b/batchutils/batch_delete.go
@@ -95,7 +95,7 @@ func BatchDelete(ctx context.Context, props *BatchDeleteRequest) *BatchDeleteErr
 		}()
 	}
 
-	go keyDistributor(cancelCtx, props.Keys, keyChan)
+	go keyDistributor(cancelCtx, props.Client.Logger(), props.MaxConcurrentDeletes, props.Keys, keyChan)
 
 	// wait for the workers to return
 	wg.Wait()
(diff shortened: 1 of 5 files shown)


pgautier404 (approved):
:ty:

pratik151192 (approved):

PR #392, opened 2024-03-08
chore: simplify logic of batchutils functions

cprice404:
This commit gets rid of a goroutine in the batch utils functions that is arguably unnecessary.

When we are populating the worker input channels, we were restricting the size of the channel to be only as large as the max number of workers. Then we needed a producer goroutine to feed the keys into that channel as the workers removed them.

However, since the response channel is already sized to handle the entire result set, and the items we put into the input channel are already in memory, sizing the channel so that it is large enough to hold pointers to all of them seems like a reasonable thing to do. It shouldn't increase memory usage much, and it simplifies the code considerably. We get rid of a goroutine and a context that we would otherwise need to be careful about leaking.

### DIFF
--- a/batchutils/batch_operations.go
+++ b/batchutils/batch_operations.go
@@ -1,10 +1,7 @@
 package batchutils
 
 import (
-	"context"
 	"time"
-
-	"github.com/momentohq/client-sdk-go/config/logger"
 
 	"github.com/momentohq/client-sdk-go/momento"
 )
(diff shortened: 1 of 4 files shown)


batchutils/batch_operations.go
```diff
@@ -19,28 +17,28 @@ func getRequestTimeout(propsTimeout *time.Duration) (requestTimeout time.Duratio
...
-
-	for {
-		select {
-		case <-ctx.Done():
-			logger.Trace("keyDistributor context done, exiting for loop")
-			return
-		}
-	}
-}
+//func keyDistributor(ctx context.Context, logger logger.MomentoLogger, numWorkers int, keys []momento.Key, keyChan chan momento.Key) {
```
pgautier404, line 20: Is there a reason you're keeping this block comment around?

batchutils/batch_set.go
```diff
@@ -102,35 +101,32 @@ func (e *BatchSetResponse) Responses() map[momento.Value]responses.SetResponse {
...
-
-	for {
-		select {
-		case <-ctx.Done():
-			logger.Trace("itemDistributor context done, exiting for loop")
-			return
-		}
-	}
-}
+//func itemDistributor(ctx context.Context, logger logger.MomentoLogger, numWorkers int, items []BatchSetItem, itemChan chan BatchSetItem) {
```
pgautier404, line 104: Same question as above re: block comment.

batchutils/batch_operations.go
cprice404: nope

cprice404:
this pr is for discussion. i'm not super experienced with golang so if there is an important reason not to increase the size of the channel, I don't feel strongly that this change must go in.

pgautier404 (approved):
LGTM!

PR #393, opened 2024-03-08
feat: add getBatch and setBatch methods

anitarua:
Adds getBatch and setBatch methods that are built into the CacheClient

### DIFF
--- a/responses/get_batch.go
+++ b/responses/get_batch.go
@@ -31,9 +31,9 @@ func (resp GetBatchSuccess) ValueMapStringString() map[string]string {
 	ret := make(map[string]string)
 	for idx, element := range resp.responses {
 		switch e := element.(type) {
-		case GetMiss:
+		case *GetMiss:
 			ret[string(resp.keys[idx])] = ""
-		case GetHit:
+		case *GetHit:
 			ret[string(resp.keys[idx])] = e.ValueString()
 		}
 	}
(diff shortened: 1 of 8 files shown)


responses/get_batch.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+// ValueMapStringString returns the data as a Map whose keys and values are utf-8 strings,
+// decoded from the underlying byte arrays. Misses are represented as empty strings.
+func (resp GetBatchSuccess) ValueMapStringString() map[string]string {
+	ret := make(map[string]string)
+	for idx, element := range resp.responses {
+		switch e := element.(type) {
+		case *GetMiss:
+			ret[string(resp.keys[idx])] = ""
```
pgautier404, line 35: It's possible and valid for a user to set a key with an empty string as the value, in which case it isn't possible to determine the difference between a miss and a hit. Can we use `nil` for this instead?

anitarua: I tried nil initially but it would break the build, even when I declare a map of type `map[string]*string` :/
> `cannot use nil as string value in assignment`

cprice404: what do we do in dictionarygetfields

cprice404: looks like it might just omit the missing keys from the map, which i think is better than using an empty string

anitarua: okay, I can do that instead

cprice404 (approved):
batch

PR #398, opened 2024-03-19
chore: clean up and return new sdk ConnectionError when eager connection fails

anitarua:


momento/scs_data_client.go
```diff
@@ -96,5 +96,8 @@ func (client scsDataClient) Connect() error {
 	defer cancel()
 
 	err := client.grpcManager.Connect(ctx)
+	if err != nil {
+		client.grpcManager.Close()
+	}
```
anitarua, line 101: I think `cancel` is supposed to free up resources anyway, but I figured it wouldn't hurt to call `Close` here as well if the eager connection fails

momento/cache_client_test.go
```diff
@@ -109,4 +109,20 @@ var _ = Describe("CacheClient", func() {
...
+			sharedContext.CredentialProvider,
+			sharedContext.DefaultTtl,
+			1*time.Millisecond,
+		)
+
+		Expect(client).To(BeNil())
+		Expect(err).NotTo(BeNil())
+		var momentoErr MomentoError
+		if errors.As(err, &momentoErr) {
+			Expect(momentoErr.Code()).To(Equal(ConnectionError))
```
anitarua, line 125: Added a test to verify the error will surface as type `ConnectionError`

momento/cache_client.go
```diff
@@ -240,7 +240,7 @@ func commonCacheClient(props CacheClientProps) (CacheClient, error) {
 			if err != nil {
 				logger := props.Configuration.GetLoggerFactory().GetLogger("CacheClient")
 				logger.Debug("Failed to connect to the server within the given eager connection timeout:", err.Error())
-				return nil, convertMomentoSvcErrorToCustomerError(momentoerrors.ConvertSvcErr(err))
```
pratik151192, line 243: hmm ideally the caller here should wrap any kind of `error` into a generic `connectionError`. That would allow you to not throw `ConnectionErrors` at each instance within `connect` and just return error as is. In Java we anyway had to throw an exception so I chose to throw `Connection` but in this case it's simpler to return the error as GoLang has the tuple return semantic.

anitarua: oh that's a much better approach, don't know why I didn't do that in the first place 😄 will fix

pratik151192 (approved):

PR #405, opened 2024-04-05
fix: upgrade grpc lib and use NewClient instead of Dial

anitarua:
Last part of https://github.com/momentohq/dev-eco-issue-tracker/issues/684


internal/grpcmanagers/data_manager.go
```diff
@@ -56,6 +55,10 @@ func (dataManager *DataGrpcManager) Close() momentoerrors.MomentoSvcErr {
 }
 
 func (gm *DataGrpcManager) Connect(ctx context.Context) error {
+	// Dial would connect in the background, but NewClient remains in IDLE until first RPC
+	// or until we actually call Connect here
+	gm.Conn.Connect()
+
```
anitarua, line 61: without this, eager connection context would timeout every time

pratik151192: Does the `Connect()` return an `Error`? If so, we should fail fast and log and return the error the client. That would be inline with other SDKs.

anitarua: hmm, does not appear to return anything: https://github.com/grpc/grpc-go/blob/v1.63.2/clientconn.go#L657

pratik151192: ugh that's weird. Can you wrap this with a 10 or 30 second timeout here so that we do not run into the case of being blocked on network for long periods.

anitarua: I'm not sure if that's necessary. The call to our data manager `Connect` function is already wrapped in the eager connection timeout in `scsDataClient`: https://github.com/momentohq/client-sdk-go/blob/main/momento/scs_data_client.go#L95 

Would another timeout wrapper add anything?

pratik151192: Oh nice I didn't realize it had that. We're all good then!

pratik151192 (approved):

PR #407, opened 2024-04-09
chore: add leaderboards code snippets and example program

anitarua:
Adds a leaderboard example file as well as dev docs code snippets for each leaderboard method


examples/doc-examples-go-apis.go
```diff
@@ -475,3 +478,129 @@ func example_API_SetBatch() {
...
+
+func example_API_LeaderboardGetRank() {
+	getRankResponse, err := leaderboard.GetRank(ctx, momento.LeaderboardGetRankRequest{
+		Ids: []uint32{123, 456},
+	})
+	if err != nil {
+		panic(err)
+	} else {
+		switch r := getRankResponse.(type) {
+		case *responses.LeaderboardFetchSuccess:
```
cprice404, line 573: what other `type`s are available here? Or is it just `Success`?

Is this how we modeled other `Fetch` APIs? or are they like, Hit + Miss?

anitarua: There's only [FetchSuccess](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/messages/responses/leaderboard/leaderboard-fetch.ts), no Hit/Miss. 

We'd waffled on Success/Error vs Found/NotFound/Error and settled on Success/Error after the bug bash: https://github.com/momentohq/client-sdk-javascript/pull/966

examples/doc-examples-go-apis.go
```diff
@@ -475,3 +478,129 @@ func example_API_SetBatch() {
...
+		panic(err)
+	} else {
+		switch r := fetchByScoreResponse.(type) {
+		case *responses.LeaderboardFetchSuccess:
+			fmt.Printf("Successfully fetched elements by score:\n")
+			for _, element := range r.Values() {
+				fmt.Printf("ID: %d, Score: %f, Rank: %d\n", element.Id, element.Score, element.Rank)
+			}
+		}
+	}
```
pgautier404, line 542: In the next few examples I'd suggest pulling the switch code out of the else clause. It's a little more compact and I think a little more idiomatic.

examples/leaderboards-example/main.go
```diff
@@ -0,0 +1,142 @@
...
+	// Get leaderboard length
+	lengthResponse, err := leaderboard.Length(ctx)
+	if err != nil {
+		panic(err)
+	} else {
+		switch r := lengthResponse.(type) {
+		case *responses.LeaderboardLengthSuccess:
+			fmt.Printf("Leaderboard length: %d\n", r.Length())
+		}
+	}
```
pgautier404, line 67: Same comment as above regarding pulling this response processing code out of the else clauses of these examples.

PR #413, opened 2024-06-04
feat: add storage client

pgautier404:


auth/credential_provider.go
```diff
@@ -156,6 +169,7 @@ func processV1Token(decodedBase64Token []byte) (*tokenAndEndpoints, momentoerror
 			ControlEndpoint: fmt.Sprintf("control.%s", tokenData["endpoint"]),
 			CacheEndpoint:   fmt.Sprintf("cache.%s", tokenData["endpoint"]),
 			TokenEndpoint:   fmt.Sprintf("token.%s", tokenData["endpoint"]),
+			StoreEndpoint:   fmt.Sprintf("storage.%s", tokenData["endpoint"]),
```
pgautier404, line 172: Waiting on final confirmation for the endpoint.

momento/store_client.go
```diff
@@ -0,0 +1,168 @@
...
+
+	resp, err := c.storeDataClient.get(ctx, request)
+	if err != nil {
+		return nil, momentoerrors.ConvertSvcErr(err)
+	}
+
+	return resp, nil
+}
+
+func (c defaultPreviewStoreClient) Put(ctx context.Context, request *StorePutRequest) (responses.StorePutResponse, momentoerrors.MomentoSvcErr) {
```
pgautier404, line 143: This was called `Put` in the node version of this client but is called `Get` in the protos. Which should we be using?

momento/store_put.go
```diff
@@ -0,0 +1,7 @@
+package momento
+
+type StorePutRequest struct {
+	StoreName string
+	Key       string
```
pgautier404, line 5: For all of these request types, I am solely using `string` as the data type for the `Key`. This is the type in the protos, and there is no need to swap from string to bytes while processing these keys. This is inconsistent with the `Key` type (an alias for the `Value` type) used in the cache client, but implementing these using the `Key` type will require some extra work for basically no benefit. I'm open to discussion and happy to go ahead and use the `Key` type here if folks think the consistency with the cache client key type is important.

momento/store_value.go
```diff
@@ -0,0 +1,17 @@
...
+	isStoreValue()
+}
+
+type Integer int
+
+type Double float64
+
+func (String) isStoreValue() {}
+
+func (Bytes) isStoreValue() {}
```
pgautier404, line 13: I'm reusing the `String` and `Bytes` types from `value.go`, but I think it might make more sense to keep them separate. Maybe rename these to `StoreInteger` and so on? Also applies to the symbols in the `StoreValueType` enum in `responses/store_get.go`?

momento/store_put.go
```diff
@@ -0,0 +1,10 @@
+package momento
+
+type StorePutRequest struct {
+	// StoreName is the name of the store to put the value in.
+	StoreName string
+	// Key is the key to put the value for.
+	Key string
```
pgautier404, line 7: For all of these request types, I am solely using `string` as the data type for the key. This is the type in the protos, and there is no need to swap from string to bytes while processing these keys as there was with the cache client.

This is inconsistent with the Key type (an alias for the Value type) used in the cache client, but implementing these using the Key type will require extra work for basically no benefit. I'm open to discussion and happy to go ahead and use the Key type here if folks think the consistency with the cache client key type is important.

auth/credential_provider.go
```diff
@@ -156,6 +169,7 @@ func processV1Token(decodedBase64Token []byte) (*tokenAndEndpoints, momentoerror
 			ControlEndpoint: fmt.Sprintf("control.%s", tokenData["endpoint"]),
 			CacheEndpoint:   fmt.Sprintf("cache.%s", tokenData["endpoint"]),
 			TokenEndpoint:   fmt.Sprintf("token.%s", tokenData["endpoint"]),
+			StorageEndpoint: fmt.Sprintf("storage.%s", tokenData["endpoint"]),
```
pgautier404, line 172: Waiting on final confirmation for endpoint name.

momento/storage_value.go
```diff
@@ -0,0 +1,19 @@
...
+
+// Integer plain old int.
+type Integer int
+
+// Double backed by float64 as Go doesn't have a double type.
+type Double float64
+
+func (String) isStorageValue() {}
+
+func (Bytes) isStorageValue() {}
```
pgautier404, line 15: I'm reusing the `String` and `Bytes` types from `value.go`, but I think it might make more sense to keep them separate. Maybe rename these to StorageInteger and so on? Also applies to the symbols in the StorageValueType enum in responses/store_get.go?

momento/storage_set.go
```diff
@@ -0,0 +1,10 @@
+package momento
+
+type StorageSetRequest struct {
+	// StoreName is the name of the store to put the value in.
+	StoreName string
+	// Key is the key to put the value for.
+	Key string
```
pgautier404, line 7: For all of these request types, I am solely using string as the data type for the key. This is the type in the protos, and there is no need to swap from string to bytes while processing these keys as there was with the cache client.

This is inconsistent with the Key type (an alias for the Value type) used in the cache client, but implementing these using the Key type will require extra work for basically no benefit. I'm open to discussion and happy to go ahead and use the Key type here if folks think the consistency with the cache client key type is important.

examples/store-example/main.go
```diff
@@ -0,0 +1,92 @@
...
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// This is possible because I've moved the TryGet* funcs to the StorageGetResponse interface. This
+	// retains the ability to explicitly check for the success type but allows users to grab values
+	// without needing to do a type assertion.
+	if tryGetResp.ValueType() == responses.STRING {
+		myStr, gotStr := tryGetResp.TryGetValueString()
```
cprice404, line 52: does this pattern already exist elsewhere in this SDK? I would have probably expected `myStr, err` as the return types here.

auth/credential_provider.go
cprice404: I asked in slack and this is still the expected name.

internal/models/requests.go
```diff
@@ -101,3 +106,21 @@ type LeaderboardClientRequest struct {
...
+type CreateStoreRequest struct {
+	StoreName string
+}
+
+type DeleteStoreRequest struct {
+	StoreName string
+}
+
+type ListStoresRequest struct {
+	NextToken string
```
cprice404, line 125: I believe in most of the SDKs we are not exposing this for now (for ListCaches). Not sure what we are doing in go?

cprice404: oh maybe this isn't public, maybe it's just for the proto interop?

momento/list_stores.go
```diff
@@ -0,0 +1,6 @@
+package momento
+
+type ListStoresRequest struct {
+	// Token to continue paginating through the list. It's used to handle large paginated lists.
+	NextToken string
```
cprice404, line 5: okay so this one definitely looks public, let's decide whether we actually want to surface this or not

momento/storage_set.go
cprice404: nope - this was a deliberate choice, you are doing the right thing here.

momento/storage_value.go
cprice404: i agree with your instinct to keep them separate - but would that force you to rename them with `Storage` prefix? or can you just use the same names here since they are in a different file? I guess the thing that matters is the `package`, eh?

In that case yeah I might give them different names like `StorageValueInteger` etc.

cprice404 (commented):
left a handful of questions but mostly lgtm.

Once the service is available we should probably ask @eaddingtonwhite to try this out and see if he has any feedback.

responses/storage_get.go
```diff
@@ -0,0 +1,101 @@
...
+
+const (
+	BYTES   StorageValueType = "BYTES"
+	STRING  StorageValueType = "STRING"
+	INTEGER StorageValueType = "INTEGER"
+	DOUBLE  StorageValueType = "DOUBLE"
+)
+
+// StorageGetResponse is the base response type for a store get request.
+type StorageGetResponse interface {
```
anitarua, line 13: has it been decided if the get response should modeled as Success/Error or Hit/Miss/Error?

responses/storage_get.go
```diff
@@ -0,0 +1,101 @@
...
+	isStoreGetResponse()
+	ValueType() StorageValueType
+	TryGetValueString() (string, bool)
+	TryGetValueBytes() ([]byte, bool)
+	TryGetValueInteger() (int, bool)
+	TryGetValueDouble() (float64, bool)
+}
+
+// StorageGetSuccess indicates a successful store get request.
+type StorageGetSuccess struct {
```
anitarua, line 23: is it possible to add one more layer here so that StorageGetSuccess is an interface and there can be subtypes like StorageGetSuccessBytes, StorageGetSuccessInteger, etc?

momento/list_stores.go
pgautier404: The golang cache client does unfortunately use `NextToken` in the `ListCachesRequest`. I think it would be a reasonable inconsistency (and generally less confusing) to remove it from `ListStoresRequest`. If you agree I'll yank it out.

cprice404: i'm tempted to remove it from both of them :)

I'm guessing that even in the existing ListCaches API, because of how go works with default values and stuff, it's possible that anyone constructing a listCaches request is not passing it in anyway...

examples/store-example/main.go
pgautier404: This pattern hasn't show up in the SDK so far, but it is a pretty common golang idiom ("comma, ok"), especially in situations where it is important to differentiate between values that are nil or empty because they are missing/invalid and values that were explicitly set to nil or empty. For these responses, I expect error conditions to arise and be handled as a result of the `client.Get` call that returns the response. Once we have a response, asking for the value from the response is closer to a type cast or a hash lookup, which are both operations that use the "comma, ok" idiom. This allows us to return typed default empty values while also returning the valid/invalid status of the data access attempt. So, for example, if I have a response with a string value, `myStr, ok := resp.TryGetValueDouble()` will return `(0, false)`, alerting the user that their try failed. This idiom also gives users an another way in addition to `ValueType()` to determine the response data type, as `_, ok := resp.TryGetXyzType` will yield a true value for `ok` if the type is the valid underlying type and false otherwise. Not really very efficient, but part of the spirit of the idiom.

All that said, this is more or less equivalent to the `resp, err` return signature you're more used to seeing, and I'm not opposed to using that here instead.

responses/storage_get.go
pgautier404: Yes, we decided on Success/Error.

responses/storage_get.go
pgautier404: Sounds like we're going to discuss the preferred approach to take with this soon. Thanks for the call out!

momento/list_stores.go
pgautier404: Yep, and the doc example doesn't use it either. I'll go ahead and commit a change that removes `NextToken` from the list requests for both caches and stores.

examples/store-example/main.go
pgautier404: I updated this example to use an actual `ok` variable to make the pattern more recognizable and explicit.

momento/storage_client.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/momentohq/client-sdk-go/internal/services"
+	"github.com/momentohq/client-sdk-go/responses"
+)
+
+var storageDataClientCount uint64
+
+// PREVIEW Momento Storage Client
+//
+// WARNING: the API for this client is not yet stable and may change without notice.
+// Please contact Momento if you would like to try this preview.
```
anitarua, line 22: added docstring for PreviewStorageClient, can update wording if needed

momento/storage_data_client.go
```diff
@@ -0,0 +1,130 @@
...
+	case *pb.XStoreValue_BytesValue:
+		return responses.NewStoreGetSuccess_Bytes(responses.BYTES, val.GetBytesValue()), nil
+	case *pb.XStoreValue_StringValue:
+		return responses.NewStoreGetSuccess_String(responses.STRING, val.GetStringValue()), nil
+	case *pb.XStoreValue_DoubleValue:
+		return responses.NewStoreGetSuccess_Float64(responses.DOUBLE, val.GetDoubleValue()), nil
+	case *pb.XStoreValue_IntegerValue:
+		return responses.NewStoreGetSuccess_Integer(responses.INTEGER, int(val.GetIntegerValue())), nil
+	default:
+		return nil, momentoerrors.NewMomentoSvcErr(momentoerrors.UnknownServiceError, "Unknown store value type", nil)
```
anitarua, line 128: slight deviation from what we'd aligned on, I think:
`Get` will return an error if the item is not found rather than returning a nil value

responses/storage_get.go
```diff
@@ -0,0 +1,100 @@
...
+)
+
+// StorageGetResponse is the base response type for a store get request.
+type StorageGetResponse interface {
+	isStoreGetResponse()
+	ValueType() StorageValueType
+	ValueString() (string, bool)
+	ValueBytes() ([]byte, bool)
+	ValueInteger() (int, bool)
+	ValueFloat64() (float64, bool)
```
anitarua, line 19: another slight deviation from what we'd aligned on, I think:
these value accessors use the "comma, ok" idiom instead of returning a type error.

I think the "comma, ok" pattern is still more idiomatic here because there's really only one reason why these accessors could go wrong (i.e. tried to access the wrong type), whereas the "comma, err" idiom appears to be used when potentially multiple different errors could be returned.

If we want to be more in line with the other SDKs, though, we can still use the "comma, err" idiom here and create a new `TypeError` error code

momento/storage_client.go
```diff
@@ -0,0 +1,212 @@
...
+type PreviewStorageClient interface {
+	// CreateStore creates a new store if it does not exist.
+	CreateStore(ctx context.Context, request *CreateStoreRequest) (responses.CreateStoreResponse, error)
+	// DeleteStore deletes a store and all the items within it.
+	DeleteStore(ctx context.Context, request *DeleteStoreRequest) (responses.DeleteStoreResponse, error)
```
anitarua, line 27: I updated the PreviewStorageClient to return `error` rather than `momentoerrors.MomentoSvcErr` so it's in line with the other clients.

This also fixed an issue in one of the tests (was expecting sdk error `InvalidArgumentError` but had received `ClientSdkError` due to mapping an sdk error to a `MomentoSvcErr`)

cprice404 (commented):
just clearing this off of y review list for now, please add me again when it's ready for 👀 !

anitarua:
Tried out the AWS DDB Golang sdk to put and get an item and they use type conversions to get items back:
```go
// Put a value in DDB table
	_, err = ddb.PutItem(ctx, &dynamodb.PutItemInput{
		Item: map[string]types.AttributeValue{
			"something": &types.AttributeValueMemberS{Value: "my-value"},
		},
		TableName: aws.String(tableName),
	})
	if err != nil {
		log.Fatalf("DDB: failed to put item, %v", err)
		panic(err)
	}
	fmt.Printf("DDB: put item\n")

	// Get the value
	resp, err := ddb.GetItem(ctx, &dynamodb.GetItemInput{
		Key:       map[string]types.AttributeValue{"something": &types.AttributeValueMemberS{Value: "my-value"}},
		TableName: aws.String(tableName),
	})
	if err != nil {
		log.Fatalf("DDB: failed to get item, %v", err)
		panic(err)
	}
	itemString, ok := resp.Item["something"].(*types.AttributeValueMemberS)
	if ok {
		fmt.Printf("Got item as string: %s\n", itemString.Value)
	} else {
		fmt.Printf("Not a string\n")
	}

	itemBool, ok := resp.Item["something"].(*types.AttributeValueMemberBOOL)
	if ok {
		fmt.Printf("Got item as bool: %v\n", itemBool.Value)
	} else {
		fmt.Printf("Not a bool\n")
	}
```

momento/storage_delete.go
```diff
@@ -0,0 +1,8 @@
+package momento
+
+type StorageDeleteRequest struct {
+	// StoreName is the name of the store to delete.
```
malandis, line 4:
```suggestion
	// StoreName is the name of the store to delete the key from.
```

momento/storage_set.go
```diff
@@ -0,0 +1,10 @@
+package momento
+
+type StoragePutRequest struct {
+	// StoreName is the name of the store to put the value in.
+	StoreName string
+	// Key is the key to put the value for.
+	Key string
+	// Value is the `StorageValue` value to put in the store.
+	Value StorageValue
+}
```
malandis, line 10: Recommend renaming the file to `storage_put.go`

responses/list_stores.go
```diff
@@ -0,0 +1,49 @@
...
+		nextToken: nextToken,
+		stores:    stores,
+	}
+}
+
+// NextToken Next Page Token returned by Simple Store Service along with the list of stores.
+// If nextToken is present, then this token must be provided in the next call to continue paginating through the list.
+// This is done by setting this value in ListStoresRequest.
+func (resp ListStoresSuccess) NextToken() string {
+	return resp.nextToken
```
malandis, line 28: Does this need to be here in this version?

responses/storage_set.go
```diff
@@ -0,0 +1,11 @@
...
+
+// StoragePutResponse is the base response type for a store put request.
+type StoragePutResponse interface {
+	isStorePutResponse()
+}
+
+// StoragePutSuccess indicates a successful store put request.
+type StoragePutSuccess struct{}
+
+func (StoragePutSuccess) isStorePutResponse() {}
```
malandis, line 11: recommend renaming file to `storage_put.go`.

malandis (commented):
LGTM aside from some minor cosmetic observations

internal/protos/cacheclient.proto
```diff
@@ -24,7 +24,11 @@ service Scs {
   rpc Set (_SetRequest) returns (_SetResponse) {}
   rpc SetBatch (_SetBatchRequest) returns (stream _SetResponse) {}
   rpc SetIf (_SetIfRequest) returns (_SetIfResponse) {}
-  rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {}
+
+  // Deprecated because we have SetIf - Absent to cover this case.
+  rpc SetIfNotExists (_SetIfNotExistsRequest) returns (_SetIfNotExistsResponse) {
+    option deprecated = true;
+  }
```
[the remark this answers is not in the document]
anitarua: oh, I think this snuck in when I updated the protos to use `StoragePut`

momento/storage_scalar_test.go
```diff
@@ -0,0 +1,254 @@
...
+		store := uuid.NewString()
+
+		// Missing key simply returns found=false
+		resp, found, err := sharedContext.StorageClient.Get(sharedContext.Ctx, &StorageGetRequest{
+			StoreName: sharedContext.StoreName,
+			Key:       key,
+		})
+		Expect(found).To(BeFalse())
+		Expect(err).To(BeNil())
+		Expect(resp).To(BeNil())
```
anitarua, line 57: updated the tests to reflect expected `resp, found, err` tuple

Get will return:
- `resp, true, nil` for item was found
- `nil, false, nil` for item not found
- `nil, false, err` for store not found

momento/storage_client.go
```diff
@@ -0,0 +1,211 @@
...
+
+	if _, err := prepareName(request.Key, "Key"); err != nil {
+		return nil, false, err
+	}
+
+	resp, err := c.getNextStorageDataClient().get(ctx, request)
+	if err != nil {
+		if err.Code() == NotFoundError && err.Message() == "Element not found" {
+			return nil, false, nil
+		}
```
pgautier404, line 171: I think we should go ahead and deprecate the `NotFoundError` and split it into separate errors for cache, store, and item not found in `scs_errors.go`'s `ConvertSvcErr` method. The [commit in the JS client](https://github.com/momentohq/client-sdk-javascript/pull/1339/files#diff-b548ebabc0b407e63e6aef9030d12c4cc2a8a593b60f80ef0cd23030b14dd6b7R86) posted this morning gives a good sense of how to handle this case. That gives us a fully differentiated set of errors and you can just check for the `StoreNotFound` error code here.

config/storage_config.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"time"
+
+	"github.com/momentohq/client-sdk-go/config/logger"
+)
+
+type storageConfiguration struct {
+	loggerFactory     logger.MomentoLoggerFactory
+	transportStrategy TransportStrategy
+	numGrpcChannels   uint32
```
cprice404, line 12: do we actually have this wired up yet? it's okay to include it if we do, but this is not a requirement for v0/bugbash

momento/storage_client.go
cprice404: yes please, follow the JS PR, and we need to be checking the `err` header from the metadata, not string matching on the message. the message could change over time.

momento/storage_scalar_test.go
```diff
@@ -46,14 +47,16 @@ var _ = Describe("Store scalar methods", func() {
...
+		// Missing key simply returns found=false
+		resp, found, err := sharedContext.StorageClient.Get(sharedContext.Ctx, &StorageGetRequest{
 			StoreName: sharedContext.StoreName,
 			Key:       key,
 		})
-		Expect(err).To(HaveMomentoErrorCode(NotFoundError))
-		Expect(err.Error()).To(ContainSubstring("Element not found"))
+		Expect(found).To(BeFalse())
+		Expect(err).To(BeNil())
+		Expect(resp).To(BeNil())
```
cprice404, line 57: ❤️

cprice404 (commented):
i like the way this looks if @pgautier404 does. I second his comment about the error types but otherwise 🚢 from me

config/storage_config.go
anitarua: yeah it was already wired up when I started working on this

responses/storage_get.go
pgautier404, line 19: I believe we're supposed to [defer to the name in the proto](https://www.notion.so/momentohq/Notes-on-PreviewStorageClient-796758f5af6e48619075d1d708127a7b?pvs=4#5aafe67c5ded4393be9e1714ed3ec7c4) for mismatched data types like this, so afaik this should be `ValueDouble` instead.

momento/storage_data_client.go
pgautier404: This is no longer true, right? As I understand it, the error returned on line 114 will prompt the storage client to return `nil, false, nil` for the `Get` call now?

anitarua: Right, but it still won't compile without a `default` case there

momento/storage_client.go
anitarua: I made the change. Gotta note that there wasn't a clear way for me to retrieve metadata from a DeleteStore RPC call as per the [grpc metadata docs](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#unary-call)

The fix for now is to assume if it gets a default CacheNotFoundError, it should return a StoreNotFoundError instead. Left a `TODO` to loop back to this later as needed.

internal/momentoerrors/scs_errors.go
```diff
@@ -36,8 +37,12 @@ const (
 	AuthenticationError = "AuthenticationError"
 	// LimitExceededError occurs when request rate, bandwidth, or object size exceeded the limits for the account.
 	LimitExceededError = "LimitExceededError"
-	// NotFoundError occurs when a cache with specified name doesn't exist.
-	NotFoundError = "NotFoundError"
+	// CacheNotFoundError occurs when a cache with specified name doesn't exist.
+	CacheNotFoundError = "CacheNotFoundError"
```
pgautier404, line 41: This needs to retain the string value `NotFoundError` for backwards compatibility.

internal/momentoerrors/scs_errors.go
```diff
@@ -38,6 +39,12 @@ const (
 	LimitExceededError = "LimitExceededError"
 	// NotFoundError occurs when a cache with specified name doesn't exist.
 	NotFoundError = "NotFoundError"
+	// CacheNotFoundError occurs when a cache with specified name doesn't exist.
+	CacheNotFoundError = "CacheNotFoundError"
```
pgautier404, line 43: Sorry, I specifically meant that the string value for `CacheNotFoundError` needs to be `NotFoundError`, to match the value of the `NotFoundError` which existing user code will be expecting.

pgautier404: But, yes, `NotFoundError` should remain in the code as well like you have it now.

anitarua: oh gotcha, np, will fix that

responses/storage_get.go
```diff
@@ -0,0 +1,100 @@
...
+// NewStoreGetSuccess_Bytes returns a new StorageGetSuccess containing the supplied byte slice value.
+func NewStoreGetSuccess_Bytes(valueType StorageValueType, value []byte) *StorageGetSuccess {
+	return &StorageGetSuccess{
+		valueType:  valueType,
+		valueBytes: &value,
+	}
+}
+
+// NewStoreGetSuccess_Float64 returns a new StorageGetSuccess containing the supplied float64 value.
+func NewStoreGetSuccess_Float64(valueType StorageValueType, value float64) *StorageGetSuccess {
```
pgautier404, line 87: Can you go ahead and rename this to `NewStoreSuccess_Double` for consistency please?

internal/momentoerrors/scs_errors.go
```diff
@@ -37,10 +37,8 @@ const (
 	AuthenticationError = "AuthenticationError"
 	// LimitExceededError occurs when request rate, bandwidth, or object size exceeded the limits for the account.
 	LimitExceededError = "LimitExceededError"
-	// NotFoundError occurs when a cache with specified name doesn't exist.
-	NotFoundError = "NotFoundError"
```
pgautier404, line 40: Sorry, I left a followup comment but probably after you started redoing it :-) The `NotFoundError` needs to go ahead and remain in the code and should be marked deprecated.

eaddingtonwhite (approved):

PR #419, opened 2024-06-07
feat: retry initial topic subscriptions on LimitExceeded

cprice404:
This commit modifies the logic for the initial topic subscription
attempts so that if we get a LimitExceeded error, we will retry
the subscription if there are multiple grpc channels configured.
We try up to one time per channel, on the assumption that we
may get connected to another server where we are still beneath
our subscription limit.

Tested locally against an environment that had 2 servers, with
some code that tries to create 200 subscriptions (with a limit
of 100 per server), and 10 grpc channels. Observed that the first
150-160 usually connected on their first attempt. After that,
began to observe some intermittent retries, but almost always
ended up with 200 successful subscriptions.

### DIFF
--- a/momento/topic_client.go
+++ b/momento/topic_client.go
@@ -4,6 +4,7 @@ package momento
 import (
 	"context"
 	"fmt"
+
 	"github.com/momentohq/client-sdk-go/internal/grpcmanagers"
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


pgautier404 (approved):

bruuuuuuuce (approved):
:shipit:

PR #427, opened 2024-06-24
chore: update metadata message signifying item not found

pgautier404:


malandis (commented):
Can we isolate this PR to just the element -> item metadata change?

pgautier404:
> Can we isolate this PR to just the element -> item metadata change?


Everything in this PR is related to that change and preparation for the control plane metadata coming online. It's all part of the same body of work and I'd like it all to go in now to set me up for making the second necessary change when the backend is ready.

malandis (approved):

malandis:
> > Can we isolate this PR to just the element -> item metadata change?
> 
> Everything in this PR is related to that change and preparation for the control plane metadata coming online. It's all part of the same body of work and I'd like it all to go in now to set me up for making the second necessary change when the backend is ready.

Ok. Just a tad more work for me to review and place in context.

PR #433, opened 2024-06-26
chore: update get responses

pgautier404:


cprice404:
@pgautier404  - if @eaddingtonwhite doesn't have bandwidth to review this today I think we should go ahead and merge and release it (pending CI issues), it's at least a good step forward in case a customer tries to start using this SDK soon.

eaddingtonwhite (approved):
This is looking much better @pgautier404 thanks for swarming on this I think got all my feed back in. Really like the developer ux more here now. Lets :shipit: can keep iterating based on more user feedback.

PR #444, opened 2024-07-01
chore: add strings to Describe()s and It()s for test segmentation

pgautier404:
This commit modifies the contents of the test descriptions in `Describe()` and `It()` calls so the tests can be segmented by client type. These segments are intended for use by the canaries and allow us to enable and disable features under test in a more granular (cell-based) way. A companion PR will modify the canaries to invoke the tests by segment instead of invoking the entire suite as it did previously. When that PR is verified and merged, a third PR will re-enable the storage tests that are currently disabled.


bruuuuuuuce (approved):
:shipit:

PR #446, opened 2024-07-05
fix: storage get response should use Found and NotFound, should refer to float64 type as Float instead of Double

anitarua:
- **fix: storage client value should use Float instead of Double**
- **fix: storage get response should use Found and NotFound cases**


cprice404:
I think Pete already tackled this in a different PR? https://github.com/momentohq/client-sdk-go/pull/433

PR #447, opened 2024-07-05
fix: ensure one-time headers are actually sent on only first request with non-empty info

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/933

The grpc interceptor route was causing the header to be sent on every request (once added, it couldn't be removed). This PR switches the SDK to use the per-request metadata route to send first-time headers instead. 

Verified that the dashboard will see the correct header once on the first cache data client request rather than on every request.

Subsequent requests show up as empty-valued headers in the dashboard.
I think this is true for all the SDKs ([see comment](https://github.com/momentohq/dev-eco-issue-tracker/issues/933#issuecomment-2214838194)) -> appears to be a query issue, I think it's fixed

Currently no control plane requests will send first-time headers since they won't appear on the dashboard.
All the data clients will send first-time headers but I wasn't able to see them in the dashboard.


.github/workflows/on-push-to-main.yml
```diff
@@ -99,7 +99,7 @@ jobs:
           default-branch: main
           changelog-types: '[{"type":"feat","section":"Features","hidden":false},{"type":"fix","section":"Bug Fixes","hidden":false},{"type":"chore","section":"Miscellaneous","hidden":false}]'
           extra-files: |
-            internal/interceptor/first_time_headers_interceptor.go
+            internal/metadata.go
```
malandis, line 102: For my own benefit, what is the purpose of this property?

internal/metadata.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if !FirstTimeHeadersSent {
+		FirstTimeHeadersSent = true
+		headers = append(headers, "runtime-version", "golang:"+runtime.Version())
+		headers = append(headers, "agent", "golang:"+string(clientType)+":"+Version)
+	}
+	return metadata.AppendToOutgoingContext(
+		ctx, headers...,
+	)
+}
```
malandis, line 35: Supposing a developer has multiple service clients, eg a topic client and a cache client, then only one of the clients will send the agent header. The other will not, because `FirstTimeHeadersSent` is global. You could solve this by making `FirstTimeHeadersSent` a map keyed by the client type.

momento/storage_data_client.go
```diff
@@ -47,17 +48,11 @@ func (client *storageDataClient) Close() {
...
-}
-
 func (client *storageDataClient) delete(ctx context.Context, request *StorageDeleteRequest) (responses.StorageDeleteResponse, momentoerrors.MomentoSvcErr) {
 	ctx, cancel := context.WithTimeout(ctx, client.requestTimeout)
 	defer cancel()
 
-	requestMetadata := metadata.NewOutgoingContext(
-		ctx, client.CreateNewMetadata(request.StoreName),
-	)
+	requestMetadata := internal.CreateMetadata(ctx, internal.Store, "store", request.StoreName)
```
malandis, line 55: Since this line is repeated and the parameter `"store"` repeated, it makes sense to make a function like `internal.CreateStoreMetadata` that puts these details in one place. Same goes for the other clients.

malandis (commented):
Very thorough. I have a couple questions and one suggestion.

.github/workflows/on-push-to-main.yml
anitarua: it's so [release-please can update specified lines in arbitrary files](https://github.com/googleapis/release-please/blob/main/docs/customizing.md#updating-arbitrary-files)

in this case, I put the `// x-release-please-version` annotation on line 11 of `metadata.go` so that the `agent` header will always send the latest released version number

malandis: Thanks. That's cool!

malandis (approved):

PR #457, opened 2024-07-15
chore: add dev docs snippets for storage client

anitarua:
Addresses https://github.com/momentohq/dev-eco-issue-tracker/issues/944

- Adds storage client snippets and a cheat sheet for dev docs 
- Fixes Makefile target for building examples
- Adds Makefile target and github actions step for running the dev docs snippets in CI

### DIFF
--- a/examples/doc-examples-go-apis.go
+++ b/examples/doc-examples-go-apis.go
@@ -7,6 +7,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/momentohq/client-sdk-go/storageTypes"
 	"github.com/momentohq/client-sdk-go/utils"
 
 	"github.com/google/uuid"
@@ -24,7 +25,28 @@ var (
 	cacheName         string
 	leaderboardClient momento.PreviewLeaderboardClient
 	leaderboard       momento.Leaderboard
+	storageClient     momento.PreviewStorageClient
+	storeName         string
 )
+
+func RetrieveApiKeyFromYourSecretsManager() string {
+	return "your-api-key"
+}
+
+func example_API_CredentialProviderFromString() {
+	apiKey := RetrieveApiKeyFromYourSecretsManager()
+	credentialProvider, err := auth.NewStringMomentoTokenProvider(apiKey)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func example_API_CredentialProviderFromEnvVar() {
+	credentialProvider, err := auth.NewStringMomentoTokenProvider("MOMENTO_API_KEY")
+	if err != nil {
+		panic(err)
+	}
+}
 
 func example_API_InstantiateCacheClient() {
 	context := context.Background()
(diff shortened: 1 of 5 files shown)


examples/doc-examples-go-apis.go
```diff
@@ -24,8 +25,29 @@ var (
...
+func example_API_CredentialProviderFromString() {
+	apiKey := RetrieveApiKeyFromYourSecretsManager()
+	credentialProvider, err := auth.NewStringMomentoTokenProvider(apiKey)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func example_API_CredentialProviderFromEnvVar() {
+	credentialProvider, err := auth.NewStringMomentoTokenProvider("MOMENTO_API_KEY")
```
cprice404, line 45: this might be calling the wrong fn?

examples/doc-examples-go-apis.go
```diff
@@ -632,3 +654,128 @@ func example_patterns_WriteThroughCaching() {
 		Value:     momento.String(value),
 	})
 }
+
+func example_API_Storage_InstantiateClient() {
+	credentialProvider, err := auth.NewStringMomentoTokenProvider("MOMENTO_API_KEY")
```
cprice404, line 659: same here

examples/doc-examples-go-apis.go
```diff
@@ -632,3 +654,128 @@ func example_patterns_WriteThroughCaching() {
...
+	getResp, err := storageClient.Get(ctx, &momento.StorageGetRequest{
+		StoreName: storeName,
+		Key:       "key",
+	})
+	if err != nil {
+		panic(err)
+	}
+
+	// If the value was not found, the response's Value will be nil.
+	if getResp.Value() == nil {
```
cprice404, line 727: I thought that `getResp` itself would be `nil`? @pgautier404 ?

examples/doc-examples-go-apis.go
```diff
@@ -632,3 +654,128 @@ func example_patterns_WriteThroughCaching() {
...
+		fmt.Printf("Got the bytes %b\n", t)
+	case storageTypes.Float:
+		fmt.Printf("Got the float %f\n", t)
+	case storageTypes.Int:
+		fmt.Printf("Got the integer %d\n", t)
+	case nil:
+		fmt.Println("Got nil")
+	}
+
+	// If you know the type you're expecting, you can assert it directly:
```
cprice404, line 746: I think let's put this one above the switch example, it's probably the more common use case

examples/doc-examples-go-apis.go
```diff
@@ -632,3 +654,128 @@ func example_patterns_WriteThroughCaching() {
...
+	intVal, ok := getResp.Value().(storageTypes.Int)
+	if !ok {
+		fmt.Println("Illegal type assertion")
```
cprice404, line 749: let's make this wording more clear: the value was not at int (bonus points if it's easy to print out the name of the type that it actually was, but I suspect that would be too much code and the point of this code block is to be super short)

cprice404 (commented):
lgtm other than comments i left; Probably good for @pgautier404 and @eaddingtonwhite to look too though.

Also - can you please make sure that CI is running these examples? I think it should have caught the error with the env var credential provider.

Also - after this goes in you will need to make a PR against dev docs to add the cheat sheet page etc, before you close the ticket. e.g.:

https://github.com/momentohq/public-dev-docs/pull/1050
https://github.com/momentohq/public-dev-docs/pull/1058

thanks!

examples/doc-examples-go-apis.go
pgautier404: @eaddingtonwhite  and I discussed this and, after some back and forth with it, decided that this approach was more consistent with the rest of the client APIs (and incidentally with DynamoDB, which returns nil items on miss). We just had another talk about it and I put up a [draft PR](https://github.com/momentohq/client-sdk-go/pull/458) that returns a `nil` response instead. There are advantages and disadvantages to both and it might be good to have a discussion over that PR and align on the better approach.

cc @cprice404 @anitarua

cprice404: i am totally fine with going the `.Value() == nil` approach, I only raised the question because I thought it was different from where we left off last time I was looking at this with you.

I am in favor of consistency w/our other Clients and with DDB so this approach is fine with me. Most important thing here is to make a call and move forward. So my vote is that if @eaddingtonwhite is good with this approach then we just go with it.

pgautier404: Ok, we'll stick with this pattern. Thanks.

examples/docs-examples/main.go
```diff
@@ -627,8 +673,214 @@ func example_patterns_WriteThroughCaching() {
...
+	if err != nil {
+		panic(err)
+	}
+
+	// If the value was not found, the response's Value will be nil.
+	if getResp.Value() == nil {
+		fmt.Println("Got nil")
+	}
+
+	// Then get the value from the found response.
```
pgautier404, line 755: I think this comment can be deleted.

examples/docs-examples/main.go
```diff
@@ -627,8 +673,214 @@ func example_patterns_WriteThroughCaching() {
...
+	client.DeleteCache(ctx, &momento.DeleteCacheRequest{
+		CacheName: cacheName,
+	})
+	storageClient.DeleteStore(ctx, &momento.DeleteStoreRequest{
+		StoreName: storeName,
+	})
+}
+
+func main() {
+	defer cleanup()
```
pgautier404, line 820: Perfect, thank you.

examples/docs-examples/main.go
```diff
@@ -715,7 +715,10 @@ func example_API_Storage_ListStores() {
 
 	switch r := resp.(type) {
 	case *responses.ListStoresSuccess:
-		log.Printf("Found stores %+v", r.Stores())
+		log.Printf("Found stores:")
+		for _, store := range r.Stores() {
+			log.Printf("\tStore name: %s", store.Name())
+		}
```
pgautier404, line 721: Can you please add newlines to the end of each of these `Printf`s?

pgautier404 (approved):
🚢

PR #458, opened 2024-07-16
chore: return nil response from storage get miss

pgautier404:
To allow the storage client Get response to be `nil` we had to convert the function's response type to a pointer. While this is inconsistent with the other response types, it does seem to work pretty naturally in the flow of the consumer code, which you can see in the storage-example in this PR. One disadvantage to implementing it this way is the possibility of a runtime panic if users try to access `getResponse.Value()` without first ensuring that `getResponse` isn't null.

### DIFF
--- a/examples/storage-example/main.go
+++ b/examples/storage-example/main.go
@@ -64,8 +64,8 @@ func main() {
 		panic(err)
 	}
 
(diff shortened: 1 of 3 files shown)


cprice404:
reading this description nudges me a little further in favor of the other approach.

PR #472, opened 2024-08-01
chore: remove tests that require account session token

pgautier404:


momento/auth_client_test.go
```diff
@@ -1255,609 +1231,4 @@ var _ = Describe("auth auth-client", func() {
 		})
 	})
 
-	Describe("Generate api keys", func() {
-		var sessionTokenClient AuthClient
-		var authTestCache1 string
-		var authTestCache2 string
-
-		BeforeEach(func() {
-			sessionCredsProvider, err := auth.NewEnvMomentoTokenProvider("TEST_SESSION_TOKEN")
```
cprice404, line 1264: do we have versions of all of these tests that are using a regular API key instead of a session token?

If not, should we just change this one line and keep these tests?

pgautier404: No, these tests only ran using a session token. When I replace the `sessionsCredsProvider` with the shared credential provider using a regular API key all of the tests fail with:

```
[FAILED] Failed to create superuser api key for creating auth testing cache client: PermissionError: Unauthorized: Token is not authorized to make this call
  rpc error: code = PermissionDenied desc = Unauthorized: Token is not authorized to make this call
```

cprice404: you can't create a superuser key using an API key but i think all of the other tests should work

pgautier404: That's not a test. It's in `BeforeEach()`. Let me do some more digging.

cprice404 (approved):

PR #473, opened 2024-08-07
fix: use sync.Map to track first-time agent headers

cprice404:
Prior to this commit we were using a regular `map` object for
keeping track of whether the first-time headers have been
sent for a given client. These `map` objects are not safe for
concurrent access, and will cause the program to crash if there
is a concurrent read while a write is occurring.

This commit switches to a `sync.Map` which is safe for concurrent
access.

### DIFF
--- a/internal/metadata.go
+++ b/internal/metadata.go
@@ -8,14 +8,6 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-//	var FirstTimeHeadersSent = map[ClientType]bool{
-//		Cache:       false,
-//		Store:       false,
-//		Leaderboard: false,
-//		Topic:       false,
-//		Ping:        false,
-//		Auth:        false,
-//	}
 var FirstTimeHeadersSent sync.Map
 
 func init() {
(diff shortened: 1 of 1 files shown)


internal/metadata.go
```diff
@@ -3,18 +3,30 @@ package internal
...
-	Ping:        false,
-	Auth:        false,
+//	var FirstTimeHeadersSent = map[ClientType]bool{
+//		Cache:       false,
+//		Store:       false,
+//		Leaderboard: false,
+//		Topic:       false,
+//		Ping:        false,
+//		Auth:        false,
+//	}
```
pgautier404, line 18: Remove?

pgautier404 (dismissed):

anitarua (approved):
tested locally and I saw the headers make it through to the dashboard

internal/metadata.go
cprice404: oops thanks

anitarua (approved):

pgautier404 (approved):

PR #476, opened 2024-08-09
feat: return topics subscription items with value, publisher id, and sequence number

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/974

Adds a new method `Event()` to allow topics subscriptions to return objects that implement the `TopicEvent` interface. A `TopicEvent` can be a `TopicItem`, `TopicHeartbeat`, or `TopicDiscontinuity`.

Preserves the existing `Item()` method that returns a `TopicValue` (string or bytes) directly. This is still useful for users who don't care about accessing the extra detailed items. 

For users who want the extra information, they would likely need to set up a loop and switch statements like this:

```go
        for {
		event, err := s.Event(ctx)
		if err != nil {
			return nil, err
		}

		switch e := event.(type) {
		case TopicItem:
			switch msg := e.GetValue().(type) {
				case momento.String:
					...
				case momento.Bytes:
					...
			}
		case TopicHeartbeat:
			...
		case TopicDiscontinuity:
			...
		}
	}
```


anitarua:
cc: @cprice404 @pgautier404 

Calling out that this is a backwards incompatible change that modifies the `Item()` method. Pete suggested creating a new method with a new return value instead. I can make those changes when/if we're aligned that that's the best way forward.

momento/topic_item.go
```diff
@@ -0,0 +1,27 @@
...
+
+func (m TopicItem) GetValue() Value {
+	return m.Message
+}
+
+func (m TopicItem) GetValueString() string {
+	return m.Message.asString()
+}
+
+func (m TopicItem) GetValueBytes() []byte {
```
cprice404, line 17: So, if we are going to cross this rubicon in this PR (making an alternate function to access the more advanced data), I think we should probably also go ahead and account for how we are going to expose other message types besides the normal Message/Value ones. e.g. if we decide we want to expose heartbeats and discontinuities, how would we surface those here?

I think in general the path you are on here (introducing a new "advanced" or "detailed" function to keep the existing one simple) is the right path, but we need to account for those different message types so that we don't end up needing to add a 3rd one later. If you have any ideas for what you think that could look like, feel free to sketch them! otherwise we can round up a few folks for a call to brainstorm on it whenever.

momento/topic_subscription.go
```diff
@@ -93,6 +94,77 @@ func (s *topicSubscription) Item(ctx context.Context) (TopicValue, error) {
 	}
 }
 
+func (s *topicSubscription) DetailedItem(ctx context.Context) (*TopicItem, error) {
```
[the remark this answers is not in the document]
anitarua: Yes, we can totally do that

momento/topic_item.go
```diff
@@ -0,0 +1,27 @@
+package momento
+
+type TopicItem struct {
+	Message             Value
+	PublisherId         String
+	TopicSequenceNumber uint64
+}
+
+func (m TopicItem) GetValue() Value {
```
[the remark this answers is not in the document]
anitarua: Went back to play around with this and it is actually possible to use `Value`, it would just need to be typecasted first through `string(value)` or `[]byte(value)`.

I can remove the convenience functions then if that seems sufficient.

anitarua:
Converted back to draft while we're discussing best path forward. 

I sketched out what it could look like to return an interface, DetailedTopicItem, that can be extended to include Discontinuity and Heartbeat in the future, as well as TopicMessage with the string/byte values, publisher id, and sequence number.

What do y'all think?

momento/topic_subscription.go
```diff
@@ -31,6 +32,24 @@ type topicSubscription struct {
...
+
+		switch item := item.(type) {
+		case TopicMessage:
+			return item.GetValue(), nil
+			// case TopicHeartbeat:
+			// 	continue
+			// case TopicDiscontinuity:
+			// 	continue
+		}
+	}
```
anitarua, line 49: a user could set up this loop themselves if they wanted to poll for detailed messages, such as heartbeats and discontinuities, in addition to the string/byte messages in the future

anitarua: though extending the interface would require users to update their switch statements, which would be another breaking change, right?

pgautier404: Go switch statements aren't required to be exhaustive, so it would be safe to add these in the future.

momento/topic_client_test.go
```diff
@@ -96,7 +96,71 @@ var _ = Describe("topic-client", func() {
...
+				Value:     value,
+			})
+			if err != nil {
+				panic(err)
+			}
+		}
+		time.Sleep(time.Millisecond * 100)
+		cancelFunction()
+
+		for i, receivedItem := range receivedItems {
```
cprice404, line 153: can you also make an assertion here about the total number of messages received?

And we always get at least one heartbeat, could you make an assertion about that?

momento/topic_item.go
```diff
@@ -0,0 +1,59 @@
...
+// }
+
+// func NewTopicDiscontinuity(lastKnownSequenceNumber uint64, newSequenceNumber uint64) TopicDiscontinuity {
+// 	return TopicDiscontinuity{
+// 		lastKnownSequenceNumber: lastKnownSequenceNumber,
+// 		newSequenceNumber:       newSequenceNumber,
+// 	}
+// }
+
+// func (TopicDiscontinuity) isDetailedTopicItem() {}
```
cprice404, line 31: I think that as long as we are crossing this bridge we should just go ahead and add these in now. That way we can really play with the API and make sure it feels okay.

momento/topic_item.go
```diff
@@ -0,0 +1,59 @@
...
+
+type TopicMessage struct {
```
cprice404, line 33: have we not already used this name somewhere else?

Also are we using "message" terminology here or "Item"?

anitarua: Apparently TopicMessage has not been used yet. I can totally switch back to TopicItem since that's more consistent

momento/topic_client_test.go
anitarua: Running locally, I observe that a heartbeat is not always the first item received. In the test, I allowed the subscription to stay on longer and assert that we receive at least one heartbeat afterwards.

pgautier404:
I think this looks good! I pulled down the branch to play with coding against it, and I didn't get much help from my IDE with options for the switch on the event type. For people unfamiliar with what types an Event could be, this is going to be confusing. I don't think it's a blocker for this PR, but I think we should add a docstring with an example of matching the event and then matching the event value (basically the sample in the PR description). It looks like this is kind of a weakness of this SDK in general, so a broader audit of docstrings might be in order.

anitarua:
> I did notice that GetPublisherId() is always returning a blank string, so if that's supposed to be populated you might check out what's going on there.

Apparently the token client wasn't piping `TokenId` through when creating a new disposable token. Fixed and confirmed publisher id is printed now when it exists

cprice404:
> I think this looks good! I pulled down the branch to play with coding against it, and I didn't get much help from my IDE with options for the switch on the event type. For people unfamiliar with what types an Event could be, this is going to be confusing. I don't think it's a blocker for this PR, but I think we should add a docstring with an example of matching the event and then matching the event value (basically the sample in the PR description). It looks like this is kind of a weakness of this SDK in general, so a broader audit of docstrings might be in order.
> 
> I did notice that `GetPublisherId()` is always returning a blank string, so if that's supposed to be populated you might check out what's going on there.

Just to clarify on this one - we're saying that today, in cases where we already expect users to switch on a response type, the IDE isn't great at hinting; therefore the situation for this new API is basically at parity with previous APIs? (just want to make sure we're not regressing or didn't miss some trick that we had used on the previous APIs)

cprice404:
@anitarua when you think you're done making changes to this for now, let's see if we can get @eaddingtonwhite to pull down the branch and try it out and see if he has any feedback.

anitarua:
> Just to clarify on this one - we're saying that today, in cases where we already expect users to switch on a response type, the IDE isn't great at hinting; therefore the situation for this new API is basically at parity with previous APIs? (just want to make sure we're not regressing or didn't miss some trick that we had used on the previous APIs)

Yeah, I think that's correct. Tried out several switch statements locally and type hints were pretty nonexistent 😅
I can add docstrings for `Item()` and `Event()` in this PR, and I agree an audit and more examples would be good in future

momento/topic_client_test.go
cprice404: i wonder if we are swallowing the first heartbeat in the code then? Because I could have sworn that the server always sends one as the first message so that we can make sure we didn't hit something like "cache not found".

if we are swallowing that first one that's okay though, just asking for the sake of sanity check.

cprice404 (approved):
this lgtm now but would like to get Ellery to approve before we ship

nice

eaddingtonwhite (dismissed):
@anitarua this is great! Played around with the new Topic Event type and feels nice also love all the new information can get as a user if want! Also nice work keeping everything backwards compatible. 

I left some nits on godocs. IMO we could probably take a pass in multiple places in our SDK we are a bit inconsistent with how we document functions for end users vs our other SDK's. Tried to call out how it impacts end user discoverability in this PR though. We can address in follow up PR if want though.

Nice work!!

momento/topic_subscription.go
```diff
@@ -14,7 +14,49 @@ import (
...
+	//	if err != nil {
+	//		panic(err)
+	//	}
+	//	switch msg := item.(type) {
+	//	case momento.String:
+	//		fmt.Printf("received message as string: '%v'\n", msg)
+	//	case momento.Bytes:
+	//		fmt.Printf("received message as bytes: '%v'\n", msg)
+	//	}
 	Item(ctx context.Context) (TopicValue, error)
```
anitarua, line 30: @eaddingtonwhite is this what you meant by putting the godocs on the interface? Or did you mean just put it all as a top level TopicSubscription interface go doc?

eaddingtonwhite: No this is what I meant! Looks good!

eaddingtonwhite (approved):
Nice @anitarua thanks for docs fixes!

pgautier404 (approved):
:shipit:

PR #477, opened 2024-08-13
chore: clean up and try to speed up tests

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/958

Tests were consistently taking anywhere from 18-31 minutes in CI recently. This PR attempts to refactor the tests to speed them up. The tests pass in 3-4 minutes now!

Improvements:

- Uses [BeforeSuite and AfterSuite](https://onsi.github.io/ginkgo/#suite-setup-and-cleanup-beforesuite-and-aftersuite) instead of BeforeEach and AfterEach to initialize test resources (clients, caches, stores, etc). BeforeEach created resources before each "spec" (we have at least 500 specs whereas we have only 3 suites: auth, batchutils, and momento).
- Uses BeforeEach only to provide unique names for topics and collections.
- Adds new helper functions for creating unique string and byte keys.
- Control plane tests that use cache client with default cache name now use their own dedicated cache client to avoid polluting the shared context.

### DIFF
--- a/momento/momento_suite_test.go
+++ b/momento/momento_suite_test.go
@@ -5,16 +5,29 @@ import (
 	"testing"
 
 	"github.com/momentohq/client-sdk-go/momento"
+	helpers "github.com/momentohq/client-sdk-go/momento/test_helpers"
 	"github.com/momentohq/client-sdk-go/responses"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
 )
 
+var sharedContext helpers.SharedContext
+
 func TestMomento(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Momento Suite")
 }
+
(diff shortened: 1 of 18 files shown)


momento/momento_suite_test.go
```diff
@@ -5,17 +5,30 @@ import (
...
 
 	"github.com/momentohq/client-sdk-go/momento"
+	helpers "github.com/momentohq/client-sdk-go/momento/test_helpers"
 	"github.com/momentohq/client-sdk-go/responses"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
 )
 
+var sharedContext helpers.SharedContext
```
anitarua, line 15: sharedContext is now available from the start of the ginkgo test suite rather than in `BeforeEach` closures

anitarua:
@pgautier404 If desired, I can also pin the staticcheck version in the Makefile in this PR

pgautier404 (approved):
🚢

PR #479, opened 2024-08-15
chore: slow down the tests a bit to be mindful of service limits

anitarua:
Adds 100ms of delay before each test in the control, dictionary, leaderboard, list, scalar, set, and sorted set tests Describe blocks to slow down the tests a bit.

Currently, the tests are running a bit too fast and hitting limits when they're run as canaries so hitting throttling errors. These failures unnecessarily alert oncall and increasing canary limits further would not be a good use of resources.

With the 100ms addition, the local test runtime moves from ~3min to ~5min.

Can make the delay longer or more targeted to specific tests if needed.

### DIFF
--- a/momento/control_test.go
+++ b/momento/control_test.go
@@ -3,6 +3,7 @@ package momento_test
 import (
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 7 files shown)


anitarua:
raised limits instead

PR #480, opened 2024-08-15
refactor: organize makefile phony declarations and add echo statements

malandis:
We tidy up the Makefile per best practices: group PHONY declarations at the top and add echo statements for each target for clarity.


cprice404 (approved):
yes

PR #481, opened 2024-08-16
chore: add per-service makefile targets

malandis:
Adds per-service targets to the Makefile. Also adds a separate
`install-ginkgo` target that is a prerequisite for any of the test
targets. This prerequisite ensures anyone just wanting to run the
tests has the minimal set of dependencies installed.

In a followup PR we will make per-endpoint targets as well.


anitarua (approved):

pgautier404 (approved):
LGTM!

PR #482, opened 2024-08-19
chore: try to make flaky tests less flaky

anitarua:
Saw several canary test failures from these two tests, likely flaky due to timing issues.

- Relaxed the timing requirement for the scalar test assert statement.
- Relaxed the timeout for the topics test and switched to using `WithCancel` instead of `WithTimeout` for greater clarity.

Checked that removing the context-checking part of `Item()` will make the test fail ([reference comment](https://github.com/momentohq/client-sdk-go/pull/351#discussion_r1313303122) from PR that added this test). I think the intent of this test was to check what happens when a `Item()` is called with a cancelled context (should return err) and I think that's still true after this revision.


cprice404 (approved):
k

PR #486, opened 2024-08-21
test: identify services under test w/labels and setup conditionally

malandis:
Modifies the tests to use labels to identify each service under
test. We then update the makefile to call ginkgo with a label filter,
and can then programmatically detect if (for example) the storage
service is under test. When the storage service is under test, we set
up the test resources. When the storage service is not under test, we
do not do that.

Previously we always stood up storage service test resources, which
caused errors in places where the storage service was not deployed
even if we weren't specifically testing the storage service.

Previously we used the ginkgo `focus` feature to match a regex on
the suite name. As far as I can tell, we cannot access the focus regex
programmatically, hence we use the labels instead. Because of this,
at some point we should consider renaming the test suite names as
they have been structured assuming we are using focus regex, eg

`Describe("auth auth-client", ...`

### DIFF
--- a/batchutils/batch_operations_test.go
+++ b/batchutils/batch_operations_test.go
@@ -16,7 +16,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("batch-utils", func() {
+var _ = Describe("batch-utils", Label("cache-service"), func() {
 
 	var (
 		ctx       context.Context
--- a/batchutils/batch_set_test.go
+++ b/batchutils/batch_set_test.go
@@ -16,7 +16,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("batch-utils", func() {
+var _ = Describe("batch-utils", Label("cache-service"), func() {
 
 	var (
 		ctx       context.Context
--- a/momento/momento_suite_test.go
+++ b/momento/momento_suite_test.go
@@ -13,6 +13,11 @@ import (
 )
 
 var sharedContext helpers.SharedContext
+var AUTH_SERVICE_LABEL = "auth-service"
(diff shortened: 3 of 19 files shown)


batchutils/batch_set_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("batch-utils", func() {
+var _ = Describe("batch-utils", Label("cache-service"), func() {
```
malandis, line 19: This should really be using `momento_test.CACHE_SERVICE_LABEL`.

batchutils/batch_operations_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("batch-utils", func() {
+var _ = Describe("batch-utils", Label("cache-service"), func() {
```
malandis, line 19: This should really be using momento_test.CACHE_SERVICE_LABEL.

cprice404 (approved):
🙈 , wfm if @anitarua is good w/it

momento/momento_suite_test.go
```diff
@@ -22,13 +27,38 @@ func TestMomento(t *testing.T) {
...
 	sharedContext.Close()
 })
 
+// This assumes that when we narrow tests to a specific service, we are
+// doing so with labels (as per the Makefile).
+//
+// If we want to focus tests based on test regex pattern, we will need to
+// update this function to check the test regex pattern instead of labels.
+func includesStorageTests() bool {
+	labels := CurrentSpecReport().Labels()
```
anitarua, line 46: whoa handy!

anitarua (approved):
lgtm I think

PR #488, opened 2024-08-27
fix: use unique test keys

malandis:
Some of the tests weren't using unique test keys. We fix where appropriate. The collection tests still scope a collection name too broadly; we fix the dictionary tests and leave the other collections for future PRs. Importantly we fix the `item get type` and `item get ttl` tests to use unique keys.

### DIFF
--- a/momento/scalar_test.go
+++ b/momento/scalar_test.go
@@ -1237,10 +1237,16 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 	})
 
 	Describe("ItemGetType", func() {
+		scalarName := NewRandomMomentoString()
+		dictionaryName := NewRandomString()
+		listName := NewRandomString()
+		setName := NewRandomString()
+		sortedSetName := NewRandomString()
+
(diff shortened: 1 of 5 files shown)


momento/scalar_test.go
```diff
@@ -1296,27 +1302,29 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
...
 				Ttl:       ttl,
 			})
 			Expect(err).To(BeNil())
 			resp, err := sharedContext.Client.ItemGetTtl(sharedContext.Ctx, &ItemGetTtlRequest{
 				CacheName: sharedContext.CacheName,
-				Key:       String("hi"),
+				Key:       key,
 			})
 			Expect(err).To(BeNil())
 			Expect(resp).To(BeAssignableToTypeOf(&ItemGetTtlHit{}))
```
malandis, line 1330: This test has been consistently flaky. We will see if the lack of a random key here was the cause.

cprice404 (approved):
lord's work

PR #491, opened 2024-09-04
fix: detailed sub item test canary error

rishtigupta:
## PR Description:
Since [Momento Topics do not guarantee order delivery of messages](https://docs.momentohq.com/topics/webhooks/overview#:~:text=Momento%20does%20not%20guarantee%20order%20for%20event%20delivery.), it is possible for the items to go out of sync which seems to be happening with "Publishes and receives detailed subscription items" test (keeps failing in go-lang canary). 

This commit removes the equality check from the test, and rather checks if the received item is one of the elements in expectedValues array.

### DIFF
--- a/momento/topic_client_test.go
+++ b/momento/topic_client_test.go
@@ -157,7 +157,8 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 		for i, receivedItem := range receivedItems {
 			switch r := receivedItem.(type) {
 			case TopicItem:
-				Expect(r.GetValue()).To(Equal(expectedValues[i]))
+				// Since Momento Topics do not guarantee delivery order, we can't check for equality
+				Expect(r.GetValue()).To(BeElementOf(expectedValues))
 				Expect(r.GetTopicSequenceNumber()).To(BeNumerically(">", 0))
 				Expect(r.GetTopicSequenceNumber()).To(Equal(uint64(i + 1)))
 			case TopicHeartbeat:
--- a/momento/topic_client_test.go
+++ b/momento/topic_client_test.go
@@ -165,7 +165,6 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 		for i, receivedItem := range receivedItems {
 			switch r := receivedItem.(type) {
 			case TopicItem:
-				//fmt.Println("Received item - i, value: ", i, r.GetValue())
 				Expect(r.GetValue()).To(Equal(expectedValues[i]))
 				Expect(r.GetTopicSequenceNumber()).To(BeNumerically(">", 0))
 				Expect(r.GetTopicSequenceNumber()).To(Equal(uint64(i + 1)))
--- a/momento/topic_client_test.go
+++ b/momento/topic_client_test.go
@@ -154,11 +154,19 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 
 		numberOfHeartbeats := 0
 		numberOfDiscontinuities := 0
+
+		for i, receivedItem := range receivedItems {
+			if r, ok := receivedItem.(TopicItem); ok {
+				fmt.Printf("Received item value %d: %v\n", i, r.GetValue())
+				fmt.Println("Received item topic sequence number: ", r.GetTopicSequenceNumber())
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


anitarua (dismissed):
🙌

malandis:
We also want to test that we receive all the expected items, right? Previously we tested getting all the items in the precise order. Now we only test the received items are in the expected set.

momento/topic_client_test.go
```diff
@@ -157,7 +157,8 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 		for i, receivedItem := range receivedItems {
 			switch r := receivedItem.(type) {
 			case TopicItem:
-				Expect(r.GetValue()).To(Equal(expectedValues[i]))
+				// Since Momento Topics do not guarantee delivery order, we can't check for equality
+				Expect(r.GetValue()).To(BeElementOf(expectedValues))
 				Expect(r.GetTopicSequenceNumber()).To(BeNumerically(">", 0))
 				Expect(r.GetTopicSequenceNumber()).To(Equal(uint64(i + 1)))
```
anitarua, line 163: oh wait also, I think this topic sequence number assertion would fail for items received out of order as well

anitarua: or wait, maybe the publishes were received out of order by the service, but would be returned with incrementing topics sequence number to the client anyway

cprice404: would it be helpful to add logging to this test to help us understand what it is happening until we get it resolved?

momento/topic_client_test.go
```diff
@@ -157,7 +157,8 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 		for i, receivedItem := range receivedItems {
 			switch r := receivedItem.(type) {
 			case TopicItem:
-				Expect(r.GetValue()).To(Equal(expectedValues[i]))
+				// Since Momento Topics do not guarantee delivery order, we can't check for equality
```
cprice404, line 160: I'm not sure that this is true. The docs that you linked are for webhooks, not regular topics. Let me ask.

cprice404: I just asked Dylan; the order can change for webhooks, but for subscriptions the order is strict. So I think for now if this test is failing somewhat regularly then we should just add more logging to help us understand what's happening.

rishtigupta:
Just adding logs to see what we got in the received items array

momento/topic_client_test.go
```diff
@@ -154,9 +154,18 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 
 		numberOfHeartbeats := 0
 		numberOfDiscontinuities := 0
+
+		for i, receivedItem := range receivedItems {
+			if r, ok := receivedItem.(TopicItem); ok {
+				fmt.Printf("Received item value %d: %v\n", i, r.GetValue())
+				fmt.Println("Received item topic sequence number: ", r.GetTopicSequenceNumber())
+			}
+		}
```
cprice404, line 163: do these get printed okay when run via the test runner? i know sometimes those test runners like to swallow stdout/stderr

cprice404 (commented):
just one question about the test runner console handling

momento/topic_client_test.go
```diff
@@ -154,9 +154,18 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
...
+
 		for i, receivedItem := range receivedItems {
 			switch r := receivedItem.(type) {
 			case TopicItem:
+				//fmt.Println("Received item - i, value: ", i, r.GetValue())
```
anitarua, line 168: extra comment? or should the print statements go in this loop instead?

momento/topic_client_test.go
rishtigupta: I get this output when running locally:
```
topic-client Publishes and receives detailed subscription items [topics-service]
/Users/rishtigupta/workplace/client-sdk-go/momento/topic_client_test.go:99
Received item value 0: aaa
Received item topic sequence number:  1
Received item value 1: [1 2 3]
Received item topic sequence number:  2
• [10.176 seconds]
```

momento/topic_client_test.go
rishtigupta: removed this commented log.

cprice404 (approved):

PR #494, opened 2024-09-05
chore: debug canary tests

rishtigupta:
## PR Description:

`**Test: [topics-service] "Publishes and receives detailed subscription items":**`
The test is failing because it's comparing the first received value (aaa) against the second expected value ([1, 2, 3]), which is causing the mismatch. This mismatch is likely because we have extra non-published items, such as a heartbeat, being included in the receivedItems array. The key issue is that we are not filtering out non-TopicItem events (e.g., TopicHeartbeat, TopicDiscontinuity) before doing the comparison with expectedValues.

Hence, this commit compares only the received **TopicItem values** to the expectedValues. 

`**Test: [cache-service][scalar-tests] "accurately reports the remaining TTL for a key":**`
My hunch is that it can be either of the 3 issues:
1. The TTL might not be exactly what is expected due to small timing discrepancies between setting the TTL and when it is retrieved. This might happen due to network latency or minor processing delays.
2. If ItemGetTtl is called immediately after setting the key, there may be too little difference between the ttl and RemainingTtl(), causing the comparison to fail.
3. If the system clocks between the client and server are not perfectly synchronized, it can cause discrepancies (maybe?)

I also found that gingko has an inbuilt  `BeNumerically` matcher, which is designed specifically to compare numerical values. So, instead of manual boolean checker, using inbuilt matcher to compare original ttl with remaining ttl.

### DIFF
--- a/momento/scalar_test.go
+++ b/momento/scalar_test.go
@@ -1322,6 +1322,8 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 				Ttl:       ttl,
 			})
 			Expect(err).To(BeNil())
+			// Small delay to allow for processing time
+			time.Sleep(time.Millisecond * 10)
 			resp, err := sharedContext.Client.ItemGetTtl(sharedContext.Ctx, &ItemGetTtlRequest{
 				CacheName: sharedContext.CacheName,
 				Key:       key,
@@ -1330,7 +1332,11 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 			Expect(resp).To(BeAssignableToTypeOf(&ItemGetTtlHit{}))
 			switch result := resp.(type) {
 			case *ItemGetTtlHit:
-				Expect(ttl > result.RemainingTtl()).To(BeTrue())
+				fmt.Println(`original ttl`, ttl)
+				fmt.Println(`result.RemainingTtl()`, result.RemainingTtl())
+				const gracePeriod = time.Millisecond * 50
+				// Expect remaining TTL to be less than the original TTL (with grace period)
+				Expect(ttl-gracePeriod > result.RemainingTtl()).To(BeTrue())
 				Expect(result.RemainingTtl() > (time.Second * 30)).To(BeTrue())
 			default:
 				Fail(fmt.Sprintf("expected ItemGetTtlHit but got %s", result))
--- a/momento/scalar_test.go
+++ b/momento/scalar_test.go
@@ -1322,8 +1322,7 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 				Ttl:       ttl,
 			})
 			Expect(err).To(BeNil())
-			// Small delay to allow for processing time
-			time.Sleep(time.Millisecond * 10)
+
 			resp, err := sharedContext.Client.ItemGetTtl(sharedContext.Ctx, &ItemGetTtlRequest{
 				CacheName: sharedContext.CacheName,
 				Key:       key,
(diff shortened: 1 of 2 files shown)


momento/scalar_test.go
```diff
@@ -1322,6 +1322,8 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 				Ttl:       ttl,
 			})
 			Expect(err).To(BeNil())
+			// Small delay to allow for processing time
+			time.Sleep(time.Millisecond * 10)
```
cprice404, line 1326: I always prefer to avoid sleeps if possible; if you're not getting a Miss on the subsequent line then I think this shouldn't be necessary?

momento/scalar_test.go
```diff
@@ -1330,7 +1332,11 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 			Expect(resp).To(BeAssignableToTypeOf(&ItemGetTtlHit{}))
 			switch result := resp.(type) {
 			case *ItemGetTtlHit:
-				Expect(ttl > result.RemainingTtl()).To(BeTrue())
+				fmt.Println(`original ttl`, ttl)
+				fmt.Println(`result.RemainingTtl()`, result.RemainingTtl())
```
cprice404, line 1336: this is a good idea, thanks for adding

momento/scalar_test.go
```diff
@@ -1330,7 +1332,11 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
 			Expect(resp).To(BeAssignableToTypeOf(&ItemGetTtlHit{}))
 			switch result := resp.(type) {
 			case *ItemGetTtlHit:
-				Expect(ttl > result.RemainingTtl()).To(BeTrue())
```
cprice404, line 1333: when this failed before did this assertion print anything useful? Or did it just say like, false != true?

momento/scalar_test.go
```diff
@@ -1330,7 +1332,11 @@ var _ = Describe("cache-client scalar-methods", Label(CACHE_SERVICE_LABEL), func
...
+				fmt.Println(`original ttl`, ttl)
+				fmt.Println(`result.RemainingTtl()`, result.RemainingTtl())
+				const gracePeriod = time.Millisecond * 50
+				// Expect remaining TTL to be less than the original TTL (with grace period)
+				Expect(ttl-gracePeriod > result.RemainingTtl()).To(BeTrue())
 				Expect(result.RemainingTtl() > (time.Second * 30)).To(BeTrue())
```
cprice404, line 1340: or was it this assertion that failed?

momento/topic_client_test.go
```diff
@@ -155,25 +155,28 @@ var _ = Describe("topic-client", Label(TOPICS_SERVICE_LABEL), func() {
 		numberOfHeartbeats := 0
 		numberOfDiscontinuities := 0
 
-		for i, receivedItem := range receivedItems {
+		// Collect only TopicItem events for comparison
+		receivedTopicItems := []TopicItem{}
+		for _, receivedItem := range receivedItems {
 			if r, ok := receivedItem.(TopicItem); ok {
```
cprice404, line 161: nice catch on this, I agree this looks like it's probably the culprit.

nit: I think the switch pattern is a little more readable than this and you could still apply your same fix, right?

cprice404 (commented):
handful of questions/comments.

Nice find on the topic publish test!

momento/scalar_test.go
rishtigupta: yes, just expected true, got false

momento/scalar_test.go
rishtigupta: This is the assertion that is failing:
`Expect(ttl > result.RemainingTtl()).To(BeTrue())`

malandis (approved):
LGTM

cprice404 (approved):
yes

PR #500, opened 2024-09-10
chore: add epsilon sleep durations/reduce default ttl on tests

rishtigupta:
## PR Description
This commit reduces the ttl on some of the tests. 
For the tests that verify default ttl, adding epsilon sleep durations.


cprice404 (approved):
thank you

PR #507, opened 2024-09-13
feat: improve logger interface to accept `any`

cprice404:
The MomentoLogger interface currently only accepts `...string` arguments
for its varargs. These are intended to allow logger implementations to
interpolate the varargs into the template string using functions like
`fmt.Sprintf`, which support `...any` arguments.

The current interface is too restrictive and makes the logger more
difficult to use than it should be. This change will make it more
compatible with real production logging systems.

### DIFF
--- a/config/logger/momento-logger.go
+++ b/config/logger/momento-logger.go
@@ -1,11 +1,11 @@
 package logger
 
 type MomentoLogger interface {
-	Trace(message string, args ...string)
-	Debug(message string, args ...string)
-	Info(message string, args ...string)
-	Warn(message string, args ...string)
-	Error(message string, args ...string)
+	Trace(message string, args ...any)
+	Debug(message string, args ...any)
+	Info(message string, args ...any)
+	Warn(message string, args ...any)
+	Error(message string, args ...any)
 }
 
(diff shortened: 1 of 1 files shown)


rishtigupta (dismissed):
:shipit:

cprice404:
@eaddingtonwhite @poppoerika - As I've been consuming our go SDK for some benchmarking and other work, I've gotten more hands-on with extending the logging interface to use real, production-quality logging systems, and have found this change to be necessary.

I believe this should be backward compatible with the existing API as it is simply broadening the types allowed for these signatures, but I want to be careful not to break any existing customer code, so I would appreciate input as to whether this seems like a safe change.

cprice404:
@eaddingtonwhite @poppoerika - it's not 100% backward compatible; anyone who has made their own custom implementation of `MomentoLogger` will have to change their signatures from `...string` to `...any`. I think it should be a trivial and desirable change in most cases, and I don't know if we have any users who have even implemented this interface, but I want to be cautious.

pgautier404 (approved):

eaddingtonwhite (approved):

PR #515, opened 2024-09-17
chore: Use the consistent reads flag in integration tests

nand4011:
This will make some tests get throttled locally. If this works on github and the canaries, we will follow up with local fixes.


rishtigupta (approved):
:shipit: 💪

PR #518, opened 2024-09-18
chore: Add a flag and make command to use consistent reads in tests

nand4011:
Add a flag and make command to use consistent reads in tests. This lets us avoid using them locally, making local development easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,10 @@ test: install-ginkgo
 	@echo "Running tests..."
 	@ginkgo ${GINKGO_OPTS} ${TEST_DIRS}
 
+prod-test: install-ginkgo
+	@echo "Running tests with consistent reads..."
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -80,7 +84,7 @@ test-auth-service: install-ginkgo
 
 test-cache-service: install-ginkgo
 	@echo "Testing cache service..."
-	@ginkgo ${GINKGO_OPTS} --label-filter cache-service ${TEST_DIRS}
+	@CONSISTENT_READS=1 ginkgo ${GINKGO_OPTS} --label-filter cache-service ${TEST_DIRS}
```
nand4011, line 87: This will make the canaries run the cache tests with consistent reads.

cprice404 (approved):

PR #522, opened 2024-09-26
chore: add delay between retries when topic subscription limit is reached

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1020

Added a delay of 500ms between retries if subscription limit is reached (can change the delay if another value would be better). Allowed the loop to retry forever instead of up to a max number of attempts. 

Tested locally and the log message appears to be correctly incrementing the retry attempts and the timing difference appears to be every 500ms as expected.

```
anita@Anitas-Laptop examples % go run topic-example/main.go
[2024-09-26T21:28:13Z] INFO (CacheClient): Creating cache with name: test-cache
[2024-09-26T21:28:13Z] INFO (CacheClient): Cache with name 'test-cache' already exists, skipping
[2024-09-26T21:28:19Z] INFO (topic-client): Topic subscription limit reached, checking to see if subscription is eligible for retry
[2024-09-26T21:28:19Z] INFO (topic-client): Retrying topic subscription in 500ms due to subscription limit; retry attempt 1
[2024-09-26T21:28:20Z] INFO (topic-client): Topic subscription limit reached, checking to see if subscription is eligible for retry
[2024-09-26T21:28:20Z] INFO (topic-client): Retrying topic subscription in 500ms due to subscription limit; retry attempt 2
[2024-09-26T21:28:20Z] INFO (topic-client): Topic subscription limit reached, checking to see if subscription is eligible for retry
[2024-09-26T21:28:20Z] INFO (topic-client): Retrying topic subscription in 500ms due to subscription limit; retry attempt 3
[2024-09-26T21:28:21Z] INFO (topic-client): Topic subscription limit reached, checking to see if subscription is eligible for retry
[2024-09-26T21:28:21Z] INFO (topic-client): Retrying topic subscription in 500ms due to subscription limit; retry attempt 4
[2024-09-26T21:28:22Z] INFO (topic-client): Topic subscription limit reached, checking to see if subscription is eligible for retry
[2024-09-26T21:28:22Z] INFO (topic-client): Retrying topic subscription in 500ms due to subscription limit; retry attempt 5
```

### DIFF
--- a/momento/topic_client.go
+++ b/momento/topic_client.go
@@ -98,10 +98,12 @@ func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscri
 			if rpcError != nil {
 				if rpcError.Code() == codes.ResourceExhausted {
 					c.log.Info("Topic subscription limit reached, checking to see if subscription is eligible for retry")
+					// Failed attempts only need to be incremented before we retry due to resource exhausted error.
+					failedAttempts += 1
+					continue
 				}
-				failedAttempts += 1
-				continue
 			}
+			// For all other errors, we return the error and exit the loop.
 			return nil, momentoerrors.ConvertSvcErr(err)
 		}
 		break
--- a/momento/topic_client.go
+++ b/momento/topic_client.go
@@ -98,9 +98,9 @@ func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscri
 			if rpcError != nil {
 				if rpcError.Code() == codes.ResourceExhausted {
 					c.log.Info("Topic subscription limit reached, checking to see if subscription is eligible for retry")
-					failedAttempts += 1
-					continue
 				}
+				failedAttempts += 1
+				continue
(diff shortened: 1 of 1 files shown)


momento/topic_client.go
```diff
@@ -82,9 +76,11 @@ func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscri
 
 	firstMsg := new(pb.XSubscriptionItem)
 
-	for failedAttempts < maxAttempts {
+	failedAttempts := uint32(0)
```
cprice404, line 79: i see below on line 101 where we are incrementing the failedAttempts counter; currently it looks like we only increment it if it's a ResourceExhausted error.  that seems like it will suffice for the specific bug I reported but I'm wondering if we might want to increment that in a more general branch of the error handling?

cprice404 (approved):

cprice404 (approved):

PR #540, opened 2024-10-07
feat: support topic sequence page

malandis:
# Overview
This PR updates the protos selectively to include the topic sequence page changes from [this PR on the client protos repo](https://github.com/momentohq/client-protos/pull/298). In the future we will update to the released protos.

There are two changes: the topic subscription logic and the topic subscribe logic.

# Topic Subscription changes
Updates the topic subscription to take the sequence page into account:
- Keeps track of last known sequence page in addition to sequence number so we can reconnect gracefully
- Includes sequence page detail for topic item and discontinuity objects

# Topic Subscribe changes
Previously the topic client `Subscribe` method took a `TopicSubscription` as input but ignored the resume at sequence number. While the sequence number was used _internally_ on reconnects and not expected to be used externally, we feel it could be confusing if a user sets the sequence number and we ignore it.

Since we also add the sequence page, we also pass that to the subscribe RPC.

# To test/confirm once service is live
- [x] Suppose we subscribe to a long-running topic and the sequence page is not zero. Will the first message we receive still be a heartbeat, or a discontinuity?
  - heartbeat, then discontinuity
- [x] Test re-subscription on a sequence page change works


malandis:
Opening for early review. Once the service is live, I will update the protos and rebase.

momento/topic_subscription.go
```diff
@@ -167,6 +169,7 @@ func (s *topicSubscription) attemptReconnect(ctx context.Context) {
 			CacheName:                   s.cacheName,
 			TopicName:                   s.topicName,
 			ResumeAtTopicSequenceNumber: s.lastKnownSequenceNumber,
+			SequencePage:                s.lastKnownSequencePage,
```
anitarua, line 172: I think you may need to pass sequence page through in the [pubsub client](https://github.com/momentohq/client-sdk-go/blob/main/momento/pubsub_client.go#L90) too

malandis: Great find. Fixed in [f679607](https://github.com/momentohq/client-sdk-go/pull/540/commits/f6796075f2d2e411834a53332c43cefa98f90127)

malandis:
Ready for final review.

anitarua (approved):
lgtm!

PR #551, opened 2024-10-29
fix: add message wrapper to errors and improve resource exhausted error message by using metadata

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1038

- When a RESOURCE_EXHAUSTED error is received, we now parse the `err` metadata in order to report the cause in the message wrapper. If the metadata is not available, we try to use string matching on the error details.
- Adds the missing message wrapper component for consistency with the other SDKs.
- Adds ability to get header and trailer metadata (as per the [docs](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#unary-call)) from each RPC call across the various clients. Otherwise, metadata is not available for parsing.
- Deprecates `AlreadyExistsError` in favor of `CacheAlreadyExistsError` and `StoreAlreadyExistsError` for consistency with the other SDKs.

### DIFF
--- a/internal/momentoerrors/scs_errors.go
+++ b/internal/momentoerrors/scs_errors.go
@@ -176,23 +176,23 @@ func determineLimitExceededMessageWrapper(errorMessage string, metadata ...metad
 			case "element_size_limit_exceeded":
 				messageWrapper = ElementSizeLimitExceeded
 			}
-		} else {
-			// If err metadata is not available, try string matching on the
-			// error details to return the most specific message wrapper.
-			lowerCasedMessage := strings.ToLower(errorMessage)
-			if strings.Contains(lowerCasedMessage, "subscribers") {
-				messageWrapper = TopicSubscriptionsLimitExceeded
-			} else if strings.Contains(lowerCasedMessage, "operations") {
-				messageWrapper = OperationsRateLimitExceeded
-			} else if strings.Contains(lowerCasedMessage, "throughput") {
-				messageWrapper = ThroughputRateLimitExceeded
-			} else if strings.Contains(lowerCasedMessage, "request limit") {
-				messageWrapper = RequestSizeLimitExceeded
-			} else if strings.Contains(lowerCasedMessage, "item size") {
-				messageWrapper = ItemSizeLimitExceeded
-			} else if strings.Contains(lowerCasedMessage, "element size") {
-				messageWrapper = ElementSizeLimitExceeded
-			}
+		}
+	} else {
+		// If err metadata is not available, try string matching on the
+		// error details to return the most specific message wrapper.
+		lowerCasedMessage := strings.ToLower(errorMessage)
+		if strings.Contains(lowerCasedMessage, "subscribers") {
+			messageWrapper = TopicSubscriptionsLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "operations") {
+			messageWrapper = OperationsRateLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "throughput") {
+			messageWrapper = ThroughputRateLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "request limit") {
+			messageWrapper = RequestSizeLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "item size") {
+			messageWrapper = ItemSizeLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "element size") {
+			messageWrapper = ElementSizeLimitExceeded
 		}
 	}
 	return messageWrapper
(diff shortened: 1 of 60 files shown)


momento/requester.go
```diff
@@ -30,7 +31,7 @@ func errUnexpectedGrpcResponse(r requester, grpcResp grpcResponse) momentoerrors
 type requester interface {
 	hasCacheName
 	initGrpcRequest(client scsDataClient) error
-	makeGrpcRequest(metadata context.Context, client scsDataClient) (grpcResponse, error)
+	makeGrpcRequest(requestMetadata context.Context, client scsDataClient) (grpcResponse, []metadata.MD, error)
```
anitarua, line 34: had to update this signature to be able to get the response metadata in the cache data client

internal/momentoerrors/scs_errors.go
```diff
@@ -63,60 +76,204 @@ const (
...
+					} else if strings.Contains(lowerCasedMessage, "throughput") {
+						messageWrapper = ThroughputRateLimitExceeded
+					} else if strings.Contains(lowerCasedMessage, "request limit") {
+						messageWrapper = RequestSizeLimitExceeded
+					} else if strings.Contains(lowerCasedMessage, "item size") {
+						messageWrapper = ItemSizeLimitExceeded
+					} else if strings.Contains(lowerCasedMessage, "element size") {
+						messageWrapper = ElementSizeLimitExceeded
+					}
+				}
```
malandis, line 188: Let's extract this to a function

internal/momentoerrors/scs_errors.go
```diff
@@ -63,60 +76,204 @@ const (
...
 func ConvertSvcErr(err error, metadata ...metadata.MD) MomentoSvcErr {
 	if grpcStatus, ok := status.FromError(err); ok {
 		switch grpcStatus.Code() {
 		case codes.InvalidArgument:
-			return NewMomentoSvcErr(InvalidArgumentError, grpcStatus.Message(), err)
+			return newMomentoSvcErr(
+				InvalidArgumentError,
+				grpcStatus.Message(),
+				err,
+				"Invalid argument passed to Momento client",
```
malandis, line 98: Here and else let's extract any constant strings to a single place. That will ensure maintainability across SDKs and going forward. I recommend pulling them out into a separate file

internal/momentoerrors/scs_errors.go
```diff
@@ -63,60 +76,204 @@ const (
...
+					case "request_size_limit_exceeded":
+						messageWrapper = RequestSizeLimitExceeded
+					case "item_size_limit_exceeded":
+						messageWrapper = ItemSizeLimitExceeded
+					case "element_size_limit_exceeded":
+						messageWrapper = ElementSizeLimitExceeded
+					}
+				} else {
+					// If err metadata is not available, try string matching on the
+					// error details to return the most specific message wrapper.
```
malandis, line 173: I think there's a logic error here.

Currently:
```
- if there is a metadata array
  - when entry indexed at 1 meets conditions...
     - map the metadata entry to our message
 - otherwise when metadata entry 1 does not meet those criteria, string match on the grpc error message
```

We should also string match on the grpc error message when there is no metadata at all, right?

malandis (commented):
This is a comprehensive change.

Suggestions for improvement:
- Centralize all static strings to one place
- Extract any complicated error logic to a function instead of directly in a switch

This way we can write unit tests for various more complicated things, and maintain uniformity over the messages.

internal/momentoerrors/scs_errors.go
```diff
@@ -176,23 +176,23 @@ func determineLimitExceededMessageWrapper(errorMessage string, metadata ...metad
...
+		} else if strings.Contains(lowerCasedMessage, "operations") {
+			messageWrapper = OperationsRateLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "throughput") {
+			messageWrapper = ThroughputRateLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "request limit") {
+			messageWrapper = RequestSizeLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "item size") {
+			messageWrapper = ItemSizeLimitExceeded
+		} else if strings.Contains(lowerCasedMessage, "element size") {
+			messageWrapper = ElementSizeLimitExceeded
```
malandis, line 195: You also need this block when errData[0] is empty string or `ok` is false right?

malandis (commented):
One last question otherwise good to go.

internal/momentoerrors/scs_errors.go
anitarua: correct. I think I've fixed that in [b548402](https://github.com/momentohq/client-sdk-go/pull/551/commits/b548402771d024c13b140be46b587c01765e33e1)

malandis (approved):

PR #558, opened 2024-11-19
feat: document all rpc's and if they are idempotent or not

malandis:
Based on the current cache service proto, we add in all RPCs. For
maintainability, we arrange them in the same order as in the proto
file. For the RPCs that are new relative to when this was last
updated, add a comment and mark whether they are retryable or not.


rishtigupta (approved):
:shipit: ty!

rishtigupta (approved):

pgautier404 (approved):

PR #559, opened 2024-11-21
feat: add replica-reads test

rishtigupta:
## PR Description:
- Add replica-reads test using balanced read concern.

## Issue
https://github.com/momentohq/dev-eco-issue-tracker/issues/1049

### DIFF
--- a/momento/replica_reads_test.go
+++ b/momento/replica_reads_test.go
@@ -1,0 +1,77 @@
+package momento_test
+
+import (
+	"context"
+	"fmt"
+	"github.com/google/uuid"
+	"github.com/momentohq/client-sdk-go/momento"
+	"github.com/momentohq/client-sdk-go/responses"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"math/rand"
+	"sync"
+	"time"
+)
+
+var _ = Describe("replica-reads", Label(CACHE_SERVICE_LABEL), func() {
+	It(`should read the latest value after replication delay using balanced read concern`, func() {
+		client, _ := sharedContext.GetClientPrereqsForType("withBalancedReadConcern")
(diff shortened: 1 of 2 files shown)


momento/replica_reads_test.go
```diff
@@ -0,0 +1,78 @@
...
+		client, _ := sharedContext.GetClientPrereqsForType("withBalancedReadConcern")
+		numTrials := 10
+		delayBetweenTrials := 100 * time.Millisecond
+		replicationDelay := 1 * time.Second
+
+		trialFn := func(ctx context.Context, trialNumber int, wg *sync.WaitGroup, resultChan chan error) {
+			defer wg.Done()
+
+			// Start this trial with its own delay plus a random delay
+			startDelay := time.Duration(trialNumber)*delayBetweenTrials + time.Duration(rand.Intn(20)-10)*time.Millisecond
```
malandis, line 28: Is there a typo here? `time.Duration(rand.Intn(20)-10)` is the same as `time.Duration(rand.Intn(10))`? Or are the parens mismatched?

malandis (commented):
Looks good. One question about the delay schedule.

momento/replica_reads_test.go
rishtigupta: That was intentional from my end. I wanted a random jitter that could be positive (up to 10 ms) or negative (down to -10 ms). 

- `time.Duration(rand.Intn(20)-10)` generates a random value between -10 and 10, allowing the delay to either shorten or extend a little in both directions.
- `time.Duration(rand.Intn(10))` generates a random value between 0 and 10, so it can only add extra delay.


The first approach adds random variation to the delay, which I thought would make our operations more unpredictable.

anitarua (approved):
lgtm!

malandis (approved):

PR #562, opened 2024-11-26
fix(loadgen): adjust total count in loadgen example

malandis:
The previous commit to the loadgen had a typo where we double counted unavailable
instead of adding limit exceeded.


pgautier404 (approved):
Thank you!

PR #576, opened 2025-01-16
feat: add CredentialProvider constructor for momento-local connections

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1105

Adds `NewMomentoLocalProvider()` method to construct a `CredentialProvider` that allows cache clients to connect to momento-local. To support this, we needed to add a config for allowing insecure connections as momento-local does not use TLS. 
Referenced the structure used in JS sdk (References: [credential provider](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/auth/credential-provider.ts) and [auth utils](https://github.com/momentohq/client-sdk-javascript/blob/main/packages/core/src/internal/utils/auth.ts)).

### DIFF
--- a/internal/grpcmanagers/data_manager.go
+++ b/internal/grpcmanagers/data_manager.go
@@ -3,7 +3,6 @@ package grpcmanagers
 import (
 	"context"
 	"errors"
-	"fmt"
 
 	"github.com/momentohq/client-sdk-go/internal/interceptor"
 	"github.com/momentohq/client-sdk-go/internal/models"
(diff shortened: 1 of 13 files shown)


internal/grpcmanagers/data_manager.go
```diff
@@ -30,25 +27,15 @@ func NewUnaryDataGrpcManager(request *models.DataGrpcManagerRequest) (*DataGrpcM
 
 	var conn *grpc.ClientConn
 	var err error
-	if request.EagerConnect {
```
anitarua, line 33: Also did a small refactoring, removes unnecessary if block since grpc.NewClient would be called either way with the same arguments to create a grpc connection.

Checking the eager connection bool isn't relevant here; the thing that triggers eager connection is calling the `Connect()` method below, which we do when an eager connection timeout is set ([reference).](https://github.com/momentohq/client-sdk-go/blob/c6510e2a18a05dd6e06316f7866cbc581597f905/momento/cache_client.go#L241-L242)

pgautier404: Can we go ahead and remove the `EagerConnect` field from the `DataGrpcManagerRequest` struct since we're not using it? It looks like it's only used here if I'm not missing something.

anitarua: ah good catch!

pgautier404 (dismissed):
This looks great. I would like to remove the `EagerConnect` field from `DataGrpcManagerRequest`, but that is fine to do in a followup.

pgautier404 (approved):

PR #583, opened 2025-02-04
feat: configure topic client to use separate publish and subscribe grpc channels

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1124

Under high load, grpc channels that serve both publish and subscribe requests fail to recover when the service redeploys. Creating separate unary and stream grpc channels appears to fix this problem.

Creates separate unary and stream grpc channels within the topic client (defaults to 4 unary and 4 stream channels).
Adds topic config methods `WithNumStreamGrpcChannels()` and `WithNumUnaryGrpcChannels()` to allow overriding the default number. Also adds topics-specific grpc config and transport strategy to the topics config struct to allow setting these values.

Now returns an error if user attempts to start a new subscription when numStreamChannels*100 is exceeded:
```
panic: LimitExceededError: Number of grpc streams: 402; number of channels: 4; max concurrent streams: 400; Already at maximum number of concurrent grpc streams, cannot make new subscribe requests
```

Verified that these changes work with the loadgen setup I'd been using for testing. Survived two failovers, used 1 topic, 1500 users, default 4 unary channels, `WithNumStreamGrpcChannels(15)` for the 1500 subscribers.

### DIFF
--- a/config/topics-grpc-configuration.go
+++ b/config/topics-grpc-configuration.go
@@ -3,9 +3,9 @@ package config
 import "time"
 
 type TopicsGrpcConfigurationProps struct {
-	// number of milliseconds the client is willing to wait for an RPC to complete before it is terminated
+	// The number of milliseconds the client is willing to wait for an RPC to complete before it is terminated
 	// with a DeadlineExceeded error.
-	deadline time.Duration
+	client_timeout time.Duration
 
 	// The maximum message length the client can send to the server.  If the client attempts to send a message
 	// larger than this size, it will result in a RESOURCE_EXHAUSTED error.
@@ -26,13 +26,13 @@ type TopicsGrpcConfigurationProps struct {
 
 // GrpcConfiguration Encapsulates gRPC configuration tunables.
 type TopicsGrpcConfiguration interface {
-	// GetDeadline Returns number of milliseconds the client is willing to wait for an RPC to complete before
+	// GetClientTimeout Returns number of milliseconds the client is willing to wait for an RPC to complete before
 	// it is terminated with a DeadlineExceeded error.
-	GetDeadline() time.Duration
+	GetClientTimeout() time.Duration
 
-	// WithDeadline Copy constructor for overriding the client-side deadline. Returns a new GrpcConfiguration
+	// WithClientTimeout Copy constructor for overriding the client-side deadline. Returns a new GrpcConfiguration
 	// with the specified client-side deadline
-	WithDeadline(deadline time.Duration) TopicsGrpcConfiguration
+	WithClientTimeout(deadline time.Duration) TopicsGrpcConfiguration
 
 	// GetKeepAlivePermitWithoutCalls returns bool indicating if it is permissible to send keepalive pings from the client without any outstanding calls.
 	GetKeepAlivePermitWithoutCalls() bool
--- a/config/topics-grpc-configuration.go
+++ b/config/topics-grpc-configuration.go
@@ -1,0 +1,105 @@
+package config
+
+import "time"
+
+type TopicsGrpcConfigurationProps struct {
+	// number of milliseconds the client is willing to wait for an RPC to complete before it is terminated
+	// with a DeadlineExceeded error.
+	deadline time.Duration
+
+	// The maximum message length the client can send to the server.  If the client attempts to send a message
+	// larger than this size, it will result in a RESOURCE_EXHAUSTED error.
+	maxSendMessageLength int
+
+	// The maximum message length the client can receive from the server.  If the server attempts to send a message
+	// larger than this size, it will result in a RESOURCE_EXHAUSTED error.
+	maxReceiveMessageLength int
+
+	// NumStreamGrpcChannels represents the number of GRPC channels the topic client
+	// should open and work with for stream operations (i.e. topic subscriptions).
+	numStreamGrpcChannels uint32
+
+	// NumUnaryGrpcChannels represents the number of GRPC channels the topic client
+	// should open and work with for unary operations (i.e. topic publishes).
+	numUnaryGrpcChannels uint32
+}
+
+// GrpcConfiguration Encapsulates gRPC configuration tunables.
+type TopicsGrpcConfiguration interface {
+	// GetDeadline Returns number of milliseconds the client is willing to wait for an RPC to complete before
+	// it is terminated with a DeadlineExceeded error.
+	GetDeadline() time.Duration
+
+	// WithDeadline Copy constructor for overriding the client-side deadline. Returns a new GrpcConfiguration
+	// with the specified client-side deadline
+	WithDeadline(deadline time.Duration) TopicsGrpcConfiguration
+
+	// GetKeepAlivePermitWithoutCalls returns bool indicating if it is permissible to send keepalive pings from the client without any outstanding calls.
+	GetKeepAlivePermitWithoutCalls() bool
+
+	// WithKeepAlivePermitWithoutCalls Copy constructor for overriding the keepalive permit without calls.
+	// Indicates if it permissible to send keepalive pings from the client without any outstanding streams.
+	//
+	// NOTE: keep-alives are very important for long-lived server environments where there may be periods of time
+	// when the connection is idle. However, they are very problematic for lambda environments where the lambda
+	// runtime is continuously frozen and unfrozen, because the lambda may be frozen before the "ACK" is received
+	// from the server. This can cause the keep-alive to timeout even though the connection is completely healthy.
+	// Therefore, keep-alives should be disabled in lambda and similar environments.
+	WithKeepAlivePermitWithoutCalls(keepAlivePermitWithoutCalls bool) TopicsGrpcConfiguration
+
+	// GetKeepAliveTimeout returns number of milliseconds the client will wait for a response from a keepalive or ping.
+	GetKeepAliveTimeout() time.Duration
+
+	// WithKeepAliveTimeout Copy constructor for overriding the keepalive timeout. After waiting for a duration of this time,
+	// if the keepalive ping sender does not receive the ping ack, it will close the transport.
+	//
+	// NOTE: keep-alives are very important for long-lived server environments where there may be periods of time
+	// when the connection is idle. However, they are very problematic for lambda environments where the lambda
+	// runtime is continuously frozen and unfrozen, because the lambda may be frozen before the "ACK" is received
+	// from the server. This can cause the keep-alive to timeout even though the connection is completely healthy.
+	// Therefore, keep-alives should be disabled in lambda and similar environments.
+	WithKeepAliveTimeout(keepAliveTimeout time.Duration) TopicsGrpcConfiguration
+
+	// GetKeepAliveTime returns the interval at which to send the keepalive or ping.
+	GetKeepAliveTime() time.Duration
+
+	// WithKeepAliveTime Copy constructor for overriding the keepalive time.
+	// After a duration of this time the client/server pings its peer to see if the transport is still alive.
+	//
+	// NOTE: keep-alives are very important for long-lived server environments where there may be periods of time
+	// when the connection is idle. However, they are very problematic for lambda environments where the lambda
+	// runtime is continuously frozen and unfrozen, because the lambda may be frozen before the "ACK" is received
+	// from the server. This can cause the keep-alive to timeout even though the connection is completely healthy.
+	// Therefore, keep-alives should be disabled in lambda and similar environments.
+	WithKeepAliveTime(keepAliveTime time.Duration) TopicsGrpcConfiguration
+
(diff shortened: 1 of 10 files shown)


config/topics_config.go
```diff
@@ -5,18 +5,30 @@ import (
...
 
 type topicsConfiguration struct {
-	loggerFactory    logger.MomentoLoggerFactory
-	maxSubscriptions uint32
-	numGrpcChannels  uint32
+	loggerFactory     logger.MomentoLoggerFactory
+	maxSubscriptions  uint32 // DEPRECATED
+	numGrpcChannels   uint32 // DEPRECATED
+	numStreamChannels uint32
+	numUnaryChannels  uint32
```
nand4011, line 12: Why are these in the top level configuration and not in the grpc configuration? Aren't they specific to grpc?

config/topics_config.go
```diff
@@ -28,26 +40,54 @@ type TopicsConfiguration interface {
...
 	// the topic client should open and work with.
 	GetNumGrpcChannels() uint32
 
+	// Deprecated: please use the WithNumStreamGrpcChannels and WithNumUnaryGrpcChannels overrides
+	// instead to tune the number of GRPC channels for stream and unary operations, respectively.
+	// Defaults to 4 stream channels and 4 unary channels otherwise.
+	//
+	// WithNumGrpcChannels is currently implemented to create the specified number of GRPC connections
+	// (each GRPC connection can multiplex 100 subscribers/publishers). Defaults to 1.
 	WithNumGrpcChannels(numGrpcChannels uint32) TopicsConfiguration
```
nand4011, line 63: Is numGrpcChannels used anymore after this pr? Should we have this setter override the new stream and unary fields so that existing customers who call this won't have their channel counts switched, or do we expect that no one is manually calling this?

config/topics_config.go
anitarua: oh good point, I can move these to the GrpcConfig

momento/pubsub_client.go
```diff
@@ -18,41 +18,28 @@ import (
...
+	log                     logger.MomentoLogger
+	numGrpcStreams          atomic.Int64 // TODO: encapsulate in grpc manager
 }
 
-var streamTopicManagerCount atomic.Uint64
-var numGrpcStreams atomic.Int64
-var numChannels uint32
-
 func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, momentoerrors.MomentoSvcErr) {
-	numSubscriptions := float64(request.TopicsConfiguration.GetMaxSubscriptions())
```
anitarua, line 31: > Is numGrpcChannels used anymore after this pr? Should we have this setter override the new stream and unary fields so that existing customers who call this won't have their channel counts switched, or do we expect that no one is manually calling this?

@nand4011 all of the numSubscriptions and numGrpcChannel stuff are removed here, but there's definitely at least one customer who uses `WithNumGrpcChannels`.

Leaves a good question: what should the deprecated config override methods `WithNumGrpcChannels` and `WithMaxSubscriptions` do? 

I assumed we'd just use the default number of channels (4 unary, 4 stream)

nand4011: I'm not sure what we should do. If the user set the minimum number of channels because they need more to handle their scale, then `WithNumGrpcChannels` should probably override both `numStreamChannels` and `numUnaryChannels` with that number. Otherwise, we could split `numGrpcChannels` between stream and unary, but that seems more complex. I'd probably go with overriding both stream and unary. We should get more opinions though.

config/topics_config.go
anitarua: Actually, as I began implementing it, I realized it didn't seem to be a cross-cutting grpc config across all the clients, and updating the GrpcConfiguration interface could affect those other clients too.

Would it make sense if I just keep these in the topics config?

nand4011: Drat. Unfortunately we've been pretty inconsistent about implementing per-service configuration objects. Even the SDKs that have separate grpc configurations per service tend to give them all the same interface. I didn't realize that the topic client config in go doesn't use a transport strategy to encapsulate the low level options like this.

I feel like the most correct thing to do would be to add a topics_transport_strategy and a topics_grpc_configuration, and stop assuming that a client config will contain an implementation of grpc_configuration, but that is a more complicated change.

We could add numStreamChannels and numUnaryChannels at the top level and then change their getters and setters to reach into a transport layer in the future if we want.

anitarua: Oh yeah, a topics-specific grpc configuration should work. I think it wouldn't be too bad a change, I'll make it here

momento/topic_subscription.go
```diff
@@ -110,18 +114,24 @@ func (s *topicSubscription) Event(ctx context.Context) (TopicEvent, error) {
 			select {
 			case <-ctx.Done():
 				{
-					s.log.Info("Subscription context is done; closing subscription.")
+					// s.log.Info("Subscription context is done; closing subscription.")
+					s.momentoTopicClient.numGrpcStreams.Add(-1)
```
malandis, line 118: The user can still call `Close` on the subscription, which will also decrement the number of streams, right?

momento/topic_subscription.go
```diff
@@ -110,18 +114,24 @@ func (s *topicSubscription) Event(ctx context.Context) (TopicEvent, error) {
...
+					s.log.Debug("[Event RecvMsg] Context done, numGrpcStreams: %d", s.momentoTopicClient.numGrpcStreams.Load())
 					return nil, ctx.Err()
 				}
 			case <-s.cancelContext.Done():
 				{
-					s.log.Info("Subscription context is cancelled; closing subscription.")
+					// s.log.Info("Subscription context is cancelled; closing subscription.")
+					s.momentoTopicClient.numGrpcStreams.Add(-1)
```
malandis, line 125: Same here. A user may call `Close` explicitly, which decrements the number of streams, then calls cancel, which will be handled here again.

momento/pubsub_client.go
```diff
@@ -18,41 +19,37 @@ import (
...
+		numStreamChannels = request.TopicsConfiguration.GetNumStreamGrpcChannels()
+	} else if request.TopicsConfiguration.GetNumGrpcChannels() > 0 {
+		// numGrpcChannels is deprecated, but we'll use it to set both numUnaryChannels and numStreamChannels
+		// in case there are customers still using it.
+		numStreamChannels = request.TopicsConfiguration.GetNumGrpcChannels()
+	} else if request.TopicsConfiguration.GetMaxSubscriptions() > 0 {
+		// maxSubscriptions is deprecated, but we'll use it to set numStreamChannels
+		// in case there are customers still using it.
+		numStreamChannels = uint32(math.Ceil(float64(request.TopicsConfiguration.GetMaxSubscriptions()) / 100.0))
+	}
```
malandis, line 50: Should this logic be in the configuration object itself?

momento/topic_subscription.go
anitarua: I'll double check this, good callout

momento/pubsub_client.go
anitarua: that makes more sense, will move this logic there

malandis: On second thought, the defaults should be in one place (here) but the defaults documented or specialized on the objects.

momento/pubsub_client.go
anitarua: Could also default to 4 unary channels and make WithNumGrpcChannels update only number of subscribe channels

config/topics_transport_strategy.go
```diff
@@ -0,0 +1,278 @@
...
+
+func NewTopicsStaticTransportStrategy(props *TopicsTransportStrategyProps) TopicsTransportStrategy {
+	return &TopicsStaticTransportStrategy{
+		grpcConfig: props.GrpcConfiguration,
+	}
+}
+
+// Returns the more generic GrpcConfiguration interface instead of the more specific TopicsGrpcConfiguration.
+// Used by underlying grpc manager to set cross-cutting grpc settings.
+func (s *TopicsStaticTransportStrategy) GetGrpcConfig() GrpcConfiguration {
```
anitarua, line 261: topics grpc manager expected the GrpcConfiguration type, so I returned that here. 

But I think I can implement an `IGrpcConfiguration` interface that grpcmanagers accept instead, or I could just make the topics grpc manager accept  `TopicsGrpcConfiguration` instead

anitarua: Upon further discussion, will go the `TopicsGrpcConfiguration` route for topics grpc manager. 
There should be no need for the topics-specific grpc manager to accept the grpc configs of other clients.

momento/pubsub_client.go
anitarua: addressed in https://github.com/momentohq/client-sdk-go/pull/583/commits/9bccd29d4f6b441eeec405c97e5afe0d3d4d2d0f

momento/topic_subscription.go
anitarua: addressed here: https://github.com/momentohq/client-sdk-go/pull/583/commits/6e00309735ae200fb68bd4c4c447e31effc896a2

config/topics-grpc-configuration.go
```diff
@@ -0,0 +1,105 @@
+package config
+
+import "time"
+
+type TopicsGrpcConfigurationProps struct {
+	// number of milliseconds the client is willing to wait for an RPC to complete before it is terminated
+	// with a DeadlineExceeded error.
+	deadline time.Duration
```
malandis, line 8: called `client_timeout` in other SDKs

config/topics-grpc-configuration.go
```diff
@@ -0,0 +1,105 @@
...
+	// NumUnaryGrpcChannels represents the number of GRPC channels the topic client
+	// should open and work with for unary operations (i.e. topic publishes).
+	numUnaryGrpcChannels uint32
+}
+
+// GrpcConfiguration Encapsulates gRPC configuration tunables.
+type TopicsGrpcConfiguration interface {
+	// GetDeadline Returns number of milliseconds the client is willing to wait for an RPC to complete before
+	// it is terminated with a DeadlineExceeded error.
+	GetDeadline() time.Duration
```
malandis, line 31: `GetClientTimeout`

config/topics-grpc-configuration.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// WithDeadline Copy constructor for overriding the client-side deadline. Returns a new GrpcConfiguration
+	// with the specified client-side deadline
+	WithDeadline(deadline time.Duration) TopicsGrpcConfiguration
```
malandis, line 35: `WithClientTimeout`

config/topics_transport_strategy.go
```diff
@@ -0,0 +1,271 @@
...
+	GetNumUnaryGrpcChannels() uint32
+
+	// WithNumUnaryGrpcChannels is currently implemented to create the specified number of GRPC connections
+	// for unary operations. Each GRPC connection can multiplex 100 concurrent publish requests.
+	// Defaults to 4.
+	WithNumUnaryGrpcChannels(numUnaryGrpcChannels uint32) TopicsTransportStrategy
+}
+
+type TopicsStaticGrpcConfiguration struct {
+	deadline                    time.Duration
```
malandis, line 50: clientTimeout

momento/pubsub_client.go
```diff
@@ -31,22 +29,12 @@ type pubSubClient struct {
 }
 
 func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, momentoerrors.MomentoSvcErr) {
-	// NOTE: This is hard-coded for now but we may want to expose it via TopicConfiguration in the future,
-	// as we do with some of the other clients. Defaults to keep-alive pings enabled.
-	grpcConfig := config.NewStaticGrpcConfiguration(&config.GrpcConfigurationProps{})
+	grpcConfig := request.TopicsConfiguration.GetTransportStrategy().GetGrpcConfig()
 
-	// Default to using 4 grpc channels for subscriptions
+	// Default to using 4 grpc channels for subscriptions.
```
malandis, line 34: Should externalize any magic numbers of defaults to static variables; here `DEFAULT_NUM_STREAM_CHANNELS`

momento/pubsub_client.go
```diff
@@ -60,14 +48,10 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
 		streamTopicManagers = append(streamTopicManagers, streamTopicManager)
 	}
 
-	// Default to using 4 grpc channels for publishes
+	// Default to using 4 grpc channels for publishes.
```
malandis, line 51: Same for num unary channels

malandis (commented):
Looks good just a couple minor naming issues

malandis (approved):

PR #587, opened 2025-02-27
refactor: type internal subscribe client and prefer `Recv`

malandis:
Previously the internal subscribe client was typed as
`grpc.ClientStream`, which is a low-level type and not intended for
use when dealing with streams.

We refactor the code to type with generated, specific type
`Pubsub_SubscribeClient`. We refactor the variable names to reflect
this, preferring the name `subscribeClient`. With that we use the
higher level `Recv` method instead of `RecvMsg`.


anitarua (approved):
lgtm! good find!

PR #588, opened 2025-03-01
fix: set grpc deadline on publish requests and keep count of active subscriptions in stream topic grpc managers

anitarua:
Ensure the configurable client timeout for topic client is actually used in publish requests.

Refactor the `numGrpcStreams` counter so that each subscription-specific grpc manager keeps count of active streams instead. This ensures that subscribe requests won't get queued up on full grpc channels in edge cases (like retry storms when a server redeploy happens).

As a result, we're able to remove the old "retry subscriptions when limit exceeded error received" stuff from the topic client. 

Tested locally using the basic topic-example program:

<details>
<summary>101 subscriptions and 2 stream channels when cell subscription limit set to default of 100</summary>

```
[2025-03-04T21:00:04Z] DEBUG (topic-client): Starting new subscription on grpc channel 1 which now has 51 streams
[2025-03-04T21:00:04Z] DEBUG (topic-client): Starting new subscription, total number of streams now: 101
[2025-03-04T21:00:04Z] DEBUG (topic-client): Starting new subscription with new sequence number and sequence page.
[2025-03-04T21:00:04Z] DEBUG (topic-client): failed to receive first message from subscription: rpc error: code = ResourceExhausted desc = Too many subscribers (limit: 100)
[2025-03-04T21:00:04Z] WARN (topic-client): Topic subscription limit reached for this account; please contact us at support@momentohq.com
panic: LimitExceededError: Too many subscribers (limit: 100)
rpc error: code = ResourceExhausted desc = Too many subscribers (limit: 100)
```

</details>

<details>
<summary>101 subscriptions and 1 stream channel</summary>

```
[2025-03-04T21:02:06Z] DEBUG (topic-client): Starting new subscription, total number of streams now: 99
[2025-03-04T21:02:06Z] DEBUG (topic-client): Starting new subscription with new sequence number and sequence page.
[2025-03-04T21:02:06Z] DEBUG (topic-client): Starting new subscription on grpc channel 0 which now has 100 streams
[2025-03-04T21:02:06Z] WARN (topic-client): WARNING: approaching grpc maximum concurrent stream limit, 0 remaining of total 100 streams

[2025-03-04T21:02:06Z] DEBUG (topic-client): Starting new subscription, total number of streams now: 100
[2025-03-04T21:02:06Z] DEBUG (topic-client): Starting new subscription with new sequence number and sequence page.
panic: LimitExceededError: Number of grpc streams: 100; number of channels: 1; max concurrent streams: 100; Already at maximum number of concurrent grpc streams, cannot make new subscribe requests
```

</details>

### DIFF
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync/atomic"
+	"time"
 
 	"github.com/momentohq/client-sdk-go/config/logger"
 	"github.com/momentohq/client-sdk-go/internal"
@@ -29,6 +30,7 @@ type pubSubClient struct {
 	endpoint                string
 	log                     logger.MomentoLogger
 	numGrpcStreams          atomic.Int64 // TODO: encapsulate in grpc manager
+	requestTimeout          time.Duration
 }
 
 func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, momentoerrors.MomentoSvcErr) {
@@ -66,6 +68,13 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
 		unaryTopicManagers = append(unaryTopicManagers, unaryTopicManager)
 	}
 
+	var timeout time.Duration
+	if request.TopicsConfiguration.GetClientSideTimeout() < 1 {
+		timeout = defaultRequestTimeout
+	} else {
+		timeout = request.TopicsConfiguration.GetClientSideTimeout()
+	}
+
 	return &pubSubClient{
 		numUnaryChannels:    numUnaryChannels,
 		unaryTopicManagers:  unaryTopicManagers,
@@ -73,6 +82,7 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
 		streamTopicManagers: streamTopicManagers,
 		endpoint:            request.CredentialProvider.GetCacheEndpoint(),
 		log:                 request.Log,
+		requestTimeout:      timeout,
 	}, nil
 }
 
@@ -154,6 +164,9 @@ func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSu
 }
 
 func (client *pubSubClient) topicPublish(ctx context.Context, request *TopicPublishRequest) error {
+	ctx, cancel := context.WithTimeout(ctx, client.requestTimeout)
+	defer cancel()
+
 	requestMetadata := internal.CreateMetadata(ctx, internal.Topic)
 	topicManager := client.getNextUnaryTopicManager()
 	var header, trailer metadata.MD
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -172,6 +172,7 @@ func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSu
 	})
 
 	if err != nil {
+		topicManager.NumActiveSubscriptions.Add(-1)
 		cancelFunction()
 		if subscribeClient != nil {
 			header, _ = subscribeClient.Header()
--- a/momento/pubsub_client.go
+++ b/momento/pubsub_client.go
@@ -21,16 +21,16 @@ const DEFAULT_NUM_STREAM_GRPC_CHANNELS uint32 = 4
 const DEFAULT_NUM_UNARY_GRPC_CHANNELS uint32 = 4
 
 type pubSubClient struct {
-	numUnaryChannels        uint32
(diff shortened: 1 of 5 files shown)


momento/pubsub_client.go
```diff
@@ -66,40 +68,67 @@ func newPubSubClient(request *models.PubSubClientRequest) (*pubSubClient, moment
...
+// Each grpc connection can multiplex 100 subscribe/publish requests.
+// Grpc channels that already have 100 subscriptions will silently queue up subsequent requests.
+// We must prevent subscription requests from queuing up if there are already numStreamChannels*100
+// concurrent streams as it causes the program to hang indefinitely with no error.
+func (client *pubSubClient) getNextStreamTopicManager() (*grpcmanagers.TopicGrpcManager, momentoerrors.MomentoSvcErr) {
+	// Max number of attempts = 100 * numStreamChannels
+	// This is in order to preserve the round-robin system (incrementing nextManagerIndex) but to not
+	// cut short the number of attempts in case there are many subscriptions starting up at the same time.
+	// TODO: do we want to set a different max number of attempts?
+	for i := 0; i < 100*int(client.numStreamChannels); i++ {
```
anitarua, line 98: open question: should a lower attempt number be set? or is this fine?

momento/topic_subscription.go
```diff
@@ -169,6 +167,11 @@ func (s *topicSubscription) Event(ctx context.Context) (TopicEvent, error) {
 	}
 }
 
+func (s *topicSubscription) decrementSubscriptionCount() {
+	s.momentoTopicClient.totalActiveSubscriptions.Add(-1)
+	s.topicManager.NumActiveSubscriptions.Add(-1)
+}
```
anitarua, line 173: I kept a counter for `totalActiveSubscriptions` to make it easier to log information (since we'd have to go through all topic managers and add up counts otherwise).

But I can remove this counter if the extra bookkeeping (like in this decrement helper function) seems not worth it

pgautier404: I was going to suggest we remove it just to keep things simple and avoid having to keep redundant counters in sync, but it looks like you're actually using this in `checkNumConcurrentStreams()` in the `pubsub_client`. Even given that, though, I think it's a lot safer to just sum the subscription counts of the topic managers in `checkNumConcurrentStreams()` as well than to track the sum manually separately.

momento/pubsub_client.go
pgautier404: I guess this is fine? If I were to suggest anything lower it would be totally arbitrary.

momento/topic_subscription.go
anitarua: sounds good, will remove the extra counter then

anitarua:
While testing locally against alpha with 101 subscriptions, I received "Topic subscription limit reached, checking to see if subscription is eligible for retry" and multiple retry attempts [from the topic client](https://github.com/momentohq/client-sdk-go/blob/main/momento/topic_client.go#L107).
This change came from here: https://github.com/momentohq/client-sdk-go/pull/419

Shouldn't we just return an error if the user has hit the default service limit of 100 subscriptions and ask them to contact us if they need more?
I think that with my PR, the problem described in https://github.com/momentohq/client-sdk-go/pull/419 would go away since we now count number of subscriptions per grpc channel and not across the pool of channels.
If in agreement, I can remove the attempted retries from the topic client.

pgautier404:
> While testing locally against alpha with 101 subscriptions, I received "Topic subscription limit reached, checking to see if subscription is eligible for retry" and multiple retry attempts [from the topic client](https://github.com/momentohq/client-sdk-go/blob/main/momento/topic_client.go#L107). This change came from here: #419
> 
> Shouldn't we just return an error if the user has hit the default service limit of 100 subscriptions and ask them to contact us if they need more? I think that with my PR, the problem described in #419 would go away since we now count number of subscriptions per grpc channel and not across the pool of channels. If in agreement, I can remove the attempted retries from the topic client.

That seems right to me, I agree those retries should be removed.

momento/pubsub_client.go
```diff
@@ -130,7 +172,6 @@ func (client *pubSubClient) topicSubscribe(ctx context.Context, request *TopicSu
 	})
 
 	if err != nil {
-		client.numGrpcStreams.Add(-1)
```
pgautier404, line 133: By the time we get this error, the `NumActiveSubscriptions` count on the topic manager has been incremented in `getNextStreamTopicManager`. I think that needs to be decremented here?

anitarua: yes, you're right, will update

pgautier404 (approved):
LGTM, thank you!

PR #593, opened 2025-03-05
chore: update topics loadgen example

anitarua:
Cleaned up the example program around the edges:

- Made sure there's a supervisor publishing thread that sends new publishes independent of response time
- Allow setting different numbers of subscribers and publishers (manually verified the numbers look right)
- Upgraded examples dep to latest sdk version


rishtigupta (commented):
lgtm! 
Will let @malandis take a look for a final review!

pgautier404 (approved):
This is great, thank you!

PR #594, opened 2025-03-06
feat: add middleware

pgautier404:
This commit adds a middleware framework to the request handling logic in `scs_data_client`. Custom middleware can use the framework to optionally filter processing by request type and will provide custom request handlers to implement`OnRequest()` and `OnResponse()` methods. A `GetMetadata()` method on the request handlers also allows for the modification of the outgoing request metadata. A sample middleware is provided to demonstrate a usage of the framework. Further examples will be added once we're able to release the SDK with these changes.

Note that most of the changed files in `momento` are request types to which I had to add a `getResponse()` accessor method. Do make sure not to gloss over the `scs_data_client` changes in that directory, as that's where the middleware is actually put to use.

This is the first step in enabling retry testing against `momento-local` which requires middleware to inject metadata into gRPC requests. I'm going to leave this as a draft PR for now as I use it to develop the momento local middleware on a branch based on this one, but it is no longer a work in progress and is good to go ahead and review.


anitarua (approved):
the changes make sense to me! the optional list of methods to apply middleware to is pretty neat!

PR #596, opened 2025-03-12
fix: add topic subscribe timeout

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1146
Subscribe request should return DeadlineExceeded error if first heartbeat not received within the client timeout duration.

Tested locally (allowed queueing up of subscribe requests again in pubsub client, disabled checks) and received the timeout error as expected when attempting to make 101 subscriptions on 1 grpc channel:

```
[2025-03-12T23:35:07Z] WARN (topic-client): WARNING: approaching grpc maximum concurrent stream limit, 1 remaining of total 100 streams

[2025-03-12T23:35:07Z] WARN (topic-client): WARNING: approaching grpc maximum concurrent stream limit, 0 remaining of total 100 streams

panic: TimeoutError: subscription did not receive first message within the expected time

goroutine 1 [running]:
main.main()
        /Users/anita/client-sdk-go/examples/topic-example/main.go:44 +0x274
exit status 2
```

### DIFF
--- a/momento/topic_client.go
+++ b/momento/topic_client.go
@@ -4,8 +4,8 @@ package momento
 import (
 	"context"
 	"fmt"
+	"time"
 
-	"github.com/momentohq/client-sdk-go/internal/grpcmanagers"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
@@ -31,16 +31,25 @@ type defaultTopicClient struct {
 	numChannels        uint32
 	pubSubClient       *pubSubClient
 	log                logger.MomentoLogger
+	requestTimeout     time.Duration
 }
 
 // NewTopicClient returns a new TopicClient with provided configuration and credential provider arguments.
 func NewTopicClient(topicsConfiguration config.TopicsConfiguration, credentialProvider auth.CredentialProvider) (TopicClient, error) {
 	numChannels := topicsConfiguration.GetNumGrpcChannels()
 
+	var timeout time.Duration
+	if topicsConfiguration.GetClientSideTimeout() < 1 {
+		timeout = defaultRequestTimeout
+	} else {
+		timeout = topicsConfiguration.GetClientSideTimeout()
+	}
+
 	client := &defaultTopicClient{
 		credentialProvider: credentialProvider,
 		numChannels:        numChannels,
 		log:                topicsConfiguration.GetLoggerFactory().GetLogger("topic-client"),
+		requestTimeout:     timeout,
 	}
 
 	pubSubClient, err := newPubSubClient(&models.PubSubClientRequest{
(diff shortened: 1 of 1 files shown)


momento/topic_client.go
```diff
@@ -126,7 +127,46 @@ func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscri
...
+	// If the first message is not received within this time, we will cancel the subscription.
+	firstMessageCtx, cancel := context.WithTimeout(ctx, c.requestTimeout)
+	defer cancel()
+	subChan := make(chan topicSubscription, 1)
+	errChan := make(chan error, 1)
+
+	// Send the subscribe request in a separate goroutine to avoid blocking the main thread.
+	// Here, we'll block until one of the select cases is triggered.
+	go c.sendSubscribe(ctx, request, subChan, errChan)
+	select {
```
anitarua, line 152: Go `select` statement will block until one of these cases is triggered (either context is done due to timeout or cancel function was called, subscription was received, or subscription error was received).
Note: no default case here because otherwise the select statement would not block, it would only select the default case.

https://go.dev/tour/concurrency/6

momento/topic_client.go
```diff
@@ -31,16 +31,25 @@ type defaultTopicClient struct {
...
 // NewTopicClient returns a new TopicClient with provided configuration and credential provider arguments.
 func NewTopicClient(topicsConfiguration config.TopicsConfiguration, credentialProvider auth.CredentialProvider) (TopicClient, error) {
 	numChannels := topicsConfiguration.GetNumGrpcChannels()
 
+	var timeout time.Duration
+	if topicsConfiguration.GetClientSideTimeout() < 1 {
+		timeout = defaultRequestTimeout
+	} else {
+		timeout = topicsConfiguration.GetClientSideTimeout()
+	}
```
malandis, line 46: In this case a user cannot specify "I don't want a timeout", correct?

momento/topic_client.go
```diff
@@ -126,7 +127,46 @@ func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscri
...
+	errChan := make(chan error, 1)
+
+	// Send the subscribe request in a separate goroutine to avoid blocking the main thread.
+	// Here, we'll block until one of the select cases is triggered.
+	go c.sendSubscribe(ctx, request, subChan, errChan)
+	select {
+	case <-ctx.Done():
+		return nil, momentoerrors.NewMomentoSvcErr(
+			momentoerrors.TimeoutError,
+			"subscription did not receive first message within the expected time",
```
malandis, line 156: Why is this a timeout error? Wouldn't this happen if the caller (or the service) cancels `ctx` before the first heartbeat is received, but not because of a timeout?

momento/topic_client.go
```diff
@@ -126,7 +127,46 @@ func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscri
 		log:                c.log,
 		cancelContext:      cancelContext,
 		cancelFunction:     cancelFunction,
-	}, nil
+	}
+}
+
+func (c defaultTopicClient) Subscribe(ctx context.Context, request *TopicSubscribeRequest) (TopicSubscription, error) {
```
malandis, line 133: Could you move this function above `sendSubscribe`? It's somewhat harder to follow with this below it.

momento/topic_client.go
```diff
@@ -57,30 +66,18 @@ func NewTopicClient(topicsConfiguration config.TopicsConfiguration, credentialPr
...
-	topicManager, subscribeClient, cancelContext, cancelFunction, err = c.pubSubClient.topicSubscribe(ctx, &TopicSubscribeRequest{
+	topicManager, subscribeClient, cancelContext, cancelFunction, err := c.pubSubClient.topicSubscribe(requestCtx, &TopicSubscribeRequest{
 		CacheName:                   request.CacheName,
 		TopicName:                   request.TopicName,
 		ResumeAtTopicSequenceNumber: request.ResumeAtTopicSequenceNumber,
 		SequencePage:                request.SequencePage,
 	})
 	if err != nil {
-		return nil, err
+		cancelFunction()
```
malandis, line 78: Why do we need to do this now and throughout this function?

momento/topic_client.go
```diff
@@ -57,30 +66,18 @@ func NewTopicClient(topicsConfiguration config.TopicsConfiguration, credentialPr
...
-	var topicManager *grpcmanagers.TopicGrpcManager
-	var subscribeClient pb.Pubsub_SubscribeClient
-	var cancelContext context.Context
-	var cancelFunction context.CancelFunc
-	var err error
-
+func (c defaultTopicClient) sendSubscribe(requestCtx context.Context, request *TopicSubscribeRequest, subChan chan topicSubscription, errChan chan error) {
 	var firstMsg *pb.XSubscriptionItem
-	topicManager, subscribeClient, cancelContext, cancelFunction, err = c.pubSubClient.topicSubscribe(ctx, &TopicSubscribeRequest{
+	topicManager, subscribeClient, cancelContext, cancelFunction, err := c.pubSubClient.topicSubscribe(requestCtx, &TopicSubscribeRequest{
```
malandis, line 71: 👍

momento/topic_client.go
anitarua: Oh yeah good catch, will return a cancelled error instead

momento/topic_client.go
anitarua: I figured there can be cases where `pubSubClient.topicSubscribe` creates a subscription in time but does not receive/process first heartbeat before the deadline, so we should be careful in trying to clean up streams when context times out

pgautier404 (approved):
:shipit:

momento/topic_client.go
malandis: Is it an error if the caller explicitly cancels the context?

anitarua: I figured returning an error would help indicate why the TopicClient.Subscribe() call did not complete.
Could `return nil, nil` instead but that feels more ambiguous to me.

@pgautier404 any thoughts on this?

pgautier404: I think it's fine to return a cancelled error. If a user decides to explicitly cancel the context they can explicitly handle the resulting error.

PR #598, opened 2025-03-13
feat: retry metrics middleware

pgautier404:


malandis:
I still need to review the middleware PR. Will get to that and then come back here.

momento/test_helpers/retry_middleware.go
```diff
@@ -0,0 +1,215 @@
...
+func (r *retryMetrics) GetTotalRetryCount(cacheName string, requestName string) int {
+	// The first timestamp is the original request, so we subtract 1
+	return len(r.data[cacheName][requestName]) - 1
+}
+
+// GetAverageTimeBetweenRetries returns the average time between retries in seconds.
+// TODO: what resolution are we looking for here? I'm using Unix epoch time, so am currently
+//
+//	limited to seconds, but I can obviously change that.
+func (r *retryMetrics) GetAverageTimeBetweenRetries(cacheName string, requestName string) int64 {
```
anitarua, line 63: I think seconds would work as long as we write the fixed delay tests so that it's >1s between each retry attempt

momento/test_helpers/retry_middleware.go
```diff
@@ -0,0 +1,215 @@
...
+		r.props,
+	), nil
+}
+
+// AddUnaryRetryInterceptor returns a unary interceptor that will retry the request based on the retry strategy. It is
+// essentially identical to the SDK retry interceptor, but it also collects metrics on the retry attempts and returns
+// them in a custom error type. If modifications are made to the retry interceptor in the SDK, they should be mirrored
+// here. Injecting the retry interceptor is a bit of a hack, but it keeps the retry metrics gathering out of the production
+// code.
+func (r *retryMetricsMiddleware) AddUnaryRetryInterceptor(s retry.Strategy) func(
```
anitarua, line 124: I might've missed it, but is there a similar comment over in the retry interceptor code?

momento/set.go
```diff
@@ -71,15 +71,11 @@ func (r *SetRequest) makeGrpcRequest(requestMetadata context.Context, client scs
 	if err != nil {
 		return nil, responseMetadata, err
 	}
-	r.grpcResponse = resp
+	//r.grpcResponse = resp
```
anitarua, line 74: extra comment

momento/update_ttl.go
```diff
@@ -55,28 +55,19 @@ func (r *UpdateTtlRequest) makeGrpcRequest(requestMetadata context.Context, clie
...
-
-	r.grpcResponse = resp
-
 	return resp, nil, nil
 }
 
-func (r *UpdateTtlRequest) interpretGrpcResponse() error {
-	grpcResp := r.grpcResponse
+func (r *UpdateTtlRequest) interpretGrpcResponse(resp interface{}) error {
+	myResp := resp.(*pb.XUpdateTtlResponse)
```
anitarua, line 62: is this always going to be a safe typecasting? 
is removing grpcResponse from the request just to clean it up? or related to the middleware changes? (I may not be recalling correctly)

momento/test_helpers/retry_middleware.go
pgautier404: No there isn't, great call!

PR #601, opened 2025-03-17
feat: add middleware and data client retry strategies and tests 

pgautier404:
This PR is split into a commits based on functional groupings. The initial commit adds the middleware framework code and the configuration to support it. The second commit contains the largest set of changes and includes all of the modification to the individual request types necessary for compatibility with the new framework. The third commit contains the changes to the data client and supporting types, introducing the logic that integrates with the middleware framework using the modified request types. The last of the initial commits adds the middleware for testing retry strategies against Momento local  and tests for our existing strategies.


config/retry/eligibility_strategy.go
```diff
@@ -11,6 +11,7 @@ type EligibilityStrategy interface {
 var retryableStatusCodes = map[codes.Code]bool{
 	codes.Internal:    true,
 	codes.Unavailable: true,
+	codes.Canceled:    true,
```
anitarua, line 14: I thought we did not want to retry on Canceled as per https://github.com/momentohq/client-sdk-go/pull/561?

momento/retry_test.go
```diff
@@ -1,14 +1,68 @@
...
+	Expect(err).To(BeNil())
+	createResponse, err := cacheClient.CreateCache(context.Background(), &momento.CreateCacheRequest{
+		CacheName: "cache",
+	})
+	Expect(err).To(BeNil())
+	Expect(createResponse).To(Not(BeNil()))
+	return cacheClient
+}
+
+var _ = Describe("cache-client retry eligibility-strategy", Label(CACHE_SERVICE_LABEL), func() {
```
anitarua, line 47: should we make a new label for momento-local tests since they would need to start up a momento-local instance?
wouldn't this currently fail cache tests when running the tests locally?

config/retry/eligibility_strategy.go
pgautier404: Oops, thanks. I thought this was an oversight. I'm going to set this to `false` here and add a comment so we don't keep doing this over and over.

momento/retry_test.go
pgautier404: Ok, I added a label for momento-local tests that I'm using as an exclusion filter for the existing make targets. I'm also adding new make targets that include them. For example, `prod-test` does not run momento-local tests but `prod-test-all` does.

.github/workflows/on-pull-request.yml
```diff
@@ -84,5 +84,11 @@ jobs:
       - name: Run docs examples
         run: make run-docs-examples
 
+      - name: Start Momento Local
```
anitarua, line 87: can be a follow up PR, but I think it would be nice to run momento-local tests as a separate job so that they can run in parallel to the alpha integration tests

anitarua (commented):
looking good overall! and thanks for reorganizing the commits!

going to pull this down and try it out before approving :)

.github/workflows/on-pull-request.yml
anitarua: tried running momento-local tests locally and the test setup requires a MOMENTO_API_KEY but momento-local doesn't, probably will need to set up another test context

Makefile
```diff
@@ -122,6 +131,11 @@ test-auth-service: install-ginkgo
 
 
 test-cache-service: install-ginkgo
```
anitarua, line 133: another note for follow-up PR: could make a `test-retry-strategies` target (or some other name) that a concurrently running CI job could call, separate from the other integration tests. Perhaps no need to tag retry tests with both cache and momento-local labels from that point on?

anitarua (approved):
lgtm!

.github/workflows/on-pull-request.yml
pgautier404: Good finds, thanks. I'll open a followup PR.

.github/workflows/on-pull-request.yml
```diff
@@ -84,5 +84,11 @@ jobs:
...
+        run: |
+          docker run -d -p 8080:8080 gomomento/momento-local
```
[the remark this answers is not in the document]
pgautier404: Thanks, I totally missed that in the js ci.

momento/scs_data_client.go
```diff
@@ -58,21 +68,92 @@ func (client scsDataClient) makeRequest(ctx context.Context, r requester) error
...
+			return momentoerrors.NewMomentoSvcErr(momentoerrors.ClientSdkError, err.Error(), err)
+		}
+
+		// Call the request handler OnRequest method and then add the handler to list of handlers to
+		// call OnResponse on when the response comes back.
+		newReq, err := newHandler.OnRequest(req)
+		if err != nil {
+			return momentoerrors.NewMomentoSvcErr(momentoerrors.ClientSdkError, err.Error(), err)
+		}
+		if newReq != nil {
```
malandis, line 106: Why would `newReq` be nil here?

momento/scs_data_client.go
```diff
@@ -58,21 +68,92 @@ func (client scsDataClient) makeRequest(ctx context.Context, r requester) error
...
+	// inspect the response and error results. Any error returned from the middleware OnResponse()
+	// method will be immediately returned as the actual error, skipping any outstanding response handlers.
+	// If none of the response handlers return an error, the original error (if any) will be returned after
+	// it is converted to a Momento service error.
+	var newResp interface{}
+	for i := len(middlewareRequestHandlers) - 1; i >= 0; i-- {
+		var requestHandlerError error
+		rh := middlewareRequestHandlers[i]
+		newResp, requestHandlerError = rh.OnResponse(resp, requestError)
+		if newResp != nil {
```
malandis, line 136: Can `newResp` be `nil`? Would that be an error?

momento/scs_data_client.go
```diff
@@ -58,21 +68,92 @@ func (client scsDataClient) makeRequest(ctx context.Context, r requester) error
...
 	ctx, cancel := context.WithTimeout(ctx, client.requestTimeout)
 	defer cancel()
 
-	requestMetadata := internal.CreateCacheMetadata(ctx, r.cacheName())
+	// Variable to gather request metadata into. This will be passed to middleware request handlers
+	// for potential modification. The final metadata will be passed to the grpc request as headers.
+	requestMetadata := make(map[string]string)
+
+	middlewareRequestHandlers := make([]middleware.RequestHandler, 0, len(client.middleware))
+	for _, mw := range client.middleware {
```
malandis, line 84: Given this block and the one below it are self contained, it would help readability to extract each of them to separate functions.

internal/grpcmanagers/data_manager.go
```diff
@@ -19,8 +20,21 @@ func NewUnaryDataGrpcManager(request *models.DataGrpcManagerRequest) (*DataGrpcM
 	endpoint := request.CredentialProvider.GetCacheEndpoint()
 	authToken := request.CredentialProvider.GetAuthToken()
 
+	// Check the middleware list for a "RetryMiddleware" and use the AddUnaryRetryInterceptor method from the
+	// first one found. This allows us to use the optimally efficient retry interceptor for production
+	// and a custom retry interceptor that tracks retry timestamps for testing. I don't love having this special
+	// exception for testing in the code here, but at least it's a straightforward dependency injection.
```
malandis, line 26: To clarify the comment: this block allows us to override the retry interceptor (normally set by way of the retry strategy) with a middleware. We would need this for testing but otherwise default to the retry strategy route.

momento/test_helpers/retry_middleware.go
```diff
@@ -0,0 +1,224 @@
...
+	if rh.props.ErrorRpcList != nil {
+		requestMetadata["error-rpcs"] = strings.Join(*rh.props.ErrorRpcList, " ")
+	}
+	if rh.props.ErrorCount != nil {
+		requestMetadata["error-count"] = fmt.Sprintf("%d", *rh.props.ErrorCount)
+	}
+
+	// add other stuff here :-)/
+
+	return requestMetadata
```
malandis, line 223: Should this be modifying the metadata in place vs returning an altered copy?

malandis (commented):
Overall very clear. I really appreciate how you staged the commits and the comments you left in `scs_data_client.go`.

Architecturally my only concern is either the naming of `GetMetadata` or how we are doing it. Should we be modifying the metadata in place and returning the modified metadata; or instead taking the metadata as input and returning a copy? AFAICT each middleware implementation holds on to a pointer to the original metadata, so this makes it unclear (to me) what it means to `GetMetadata` from a middleware once subsequent middlewares have had that function invoked. Could we have a more functional style here where the middleware has a `ModifyMetadata` that takes as input the metadata and returns a copy?

config/retry/fixed_count_retry_strategy.go
```diff
@@ -0,0 +1,104 @@
...
+	eligibilityStrategy EligibilityStrategy
+	maxAttempts         int
+	log                 logger.MomentoLogger
+}
+
+type FixedCountRetryStrategy interface {
+	Strategy
+	WithMaxAttempts(attempts int) Strategy
+	WithEligibilityStrategy(s EligibilityStrategy) Strategy
+}
```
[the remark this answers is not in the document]
pgautier404: Same.

momento/scs_data_client.go
pgautier404: It would be nil if the request handler didn't implement `OnRequest` because the base handler method returns `nil, nil`. That is the cheapest way to say "I don't have anything to add here." I updated the doc comment on `middleware.OnRequest` to give a better explanation.

internal/grpcmanagers/data_manager.go
pgautier404: That clarification isn't really accurate. Let me know if this makes sense and is more clear.

Both code paths use an interceptor that is provided the specified retry strategy. The middleware provides an alternative implementation of the interceptor that gathers request timing data and swaps the normal grpc responses types with custom response types defined in the middleware. These custom types package the actual grpc response (or error) alongside the timestamps. The retry middleware detects the custom types in its `OnResponse` method, feeds the timestamps to a metrics collector for analysis and reporting, and returns the original grpc response (or error) to the data client where it is converted into a Momento response object and returned to the cache client.

momento/test_helpers/retry_middleware.go
pgautier404: Yeah, it's funny, I spent quite a bit of time on Friday making sure that `OnRequest` and `OnResponse` had consistent and predictable calling conventions while completely ignoring this method which should clearly be called `OnMetadata` and work the same way. I've corrected that and `OnMetadata` is now consistent with the other `On*` methods and is documented to describe its usage. I also completely removed any reference to metadata from the middleware structs as it was completely unnecessary state. Finally, I added other stuff where it says "add other stuff here :-)" :-)

config/config.go
```diff
@@ -69,6 +75,18 @@ type Configuration interface {
...
+	// the below configuration will cause each GetRequest and SetRequest to be processed by the
+	// MyMiddleware middleware request handler:
+	//   loggerFactory := momento_default_logger.NewDefaultMomentoLoggerFactory(momento_default_logger.INFO)
+	//   myConfig := config.LaptopLatest().WithMiddleware([]middleware.Middleware{
+	//     NewMyMiddleware(middleware.Props{
+	//       Logger: loggerFactory.GetLogger("MyMiddleware"),
+	//       IncludeTypes: []interface{}{&momento.GetRequest{}, &momento.SetRequest{}},
+	//     }),
+	//   })
+	WithMiddleware(middleware []middleware.Middleware) Configuration
```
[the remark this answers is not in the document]
pgautier404: Totally, I missed that in JS.

internal/grpcmanagers/data_manager.go
pgautier404: I updated the comment to clarify with something a lot less verbose than what I have in that comment. Let me know what you think.

anitarua (approved):
lgtm!!

internal/grpcmanagers/data_manager.go
malandis: Much improved. Thanks!

config/retry/eligibility_strategy.go
```diff
@@ -11,7 +13,9 @@ type EligibilityStrategy interface {
 var retryableStatusCodes = map[codes.Code]bool{
 	codes.Internal:    true,
 	codes.Unavailable: true,
-	codes.Canceled:    true,
+	// this code is retryable in other SDKs, but because the client can generate this error code
+	// by cancelling the context, we do not retry it here.
+	codes.Canceled:    false,
```
malandis: 👍

config/middleware/middleware.go
```diff
@@ -0,0 +1,211 @@
...
+// the caller.
+func (rh *requestHandler) OnRequest(_ interface{}) (interface{}, error) {
+	return nil, nil
+}
+
+// OnMetadata is called before the request is made to the backend. It receives the current request
+// metadata map, which it may modify and return. Returning nil from this method leaves the metadata unchanged.
+func (rh *requestHandler) OnMetadata(map[string]string) map[string]string {
+	return nil
+}
```
malandis, line 170: Since `map` is a reference type, an implementer of `OnMetadata` could still modify the metadata and return `nil`. The middleware processor no replace the metadata map with the one returned by the function, but the function would have modified the other one anyway.

To fix this, the middleware processor would have to pass a copy the metadata to `OnMetadata`. Or we could just document this convention: ie that implementers of `OnMetadata` should return the metadata on modification or else the behavior is ill defined.

malandis (commented):
LGTM just one last comment re: metadata

config/middleware/middleware.go
pgautier404: An implementer can return nil if they want, which is exactly what the base implementation here does. It doesn't affect the map that was passed in. As documented, returning nil from this method leaves the original metadata that was passed in unchanged. Reference type here doesn't mean "can be clobbered by reassignment." If you reassign `requestMetadata` to nil in `OnMetadata`, the original value in the caller is not affected. Similarly, if you reassign `requestMetadata` to a new map and modify that it will not affect the original map.

If you are adding or deleting fields in `requestMetadata` in `OnMetadata` (or modifying the values held by existing keys), your observation holds. I can modify the metadata map like that and then return nil, but my modifications will have already been made to the map. That is functionally the same as modifying the map and returning the modified value, so I was fine with it, but I agree that it's a little annoying that returning the map is more of an advisory than a requirement.

Sooooo, I added a method to copy the map and am using a copy to pass into `OnMetadata` :-)

I do think it's a win that people can't accidentally mutate the metadata in their `OnMetadata` methods thinking they're not doing anything permanent if they return `nil`.

malandis (approved):

anitarua (dismissed):

.github/workflows/on-push-to-main.yml
```diff
@@ -82,8 +82,12 @@ jobs:
       - name: Build examples
         run: make build-examples
 
+      - name: Start Momento Local
+        run: |
+          docker run -d -p 8080:8080 gomomento/momento-local
```
anitarua, line 87: probably good to add `docker run --rm` here too

malandis (approved):

.github/workflows/on-push-to-main.yml
pgautier404: Good call, I'll add that to my current PR.

PR #603, opened 2025-03-20
feat: add new get/set if hash apis

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1198

Confirmed the APIs work as expected against alpha using the tests in https://github.com/momentohq/client-sdk-go/pull/612, but the tests will be merged at a later date to avoid unnecessary canary failures.

We should align on naming and any other public API implications on this PR:

- Use the proto names (e.g. SetIfPresentAnd**NotHashEqual**) or rearrange a little (e.g. SetIfPresentAnd**HashNotEqual**)?
- Unconditional get and set with hash names: GetWithHash and SetWithHash


pgautier404 (approved):
This looks good to me! I personally think the names you've used here are great and avoid some ambiguities inherent in the names of the protos.

PR #604, opened 2025-03-21
feat: timestamp channel

pgautier404:
This commit removes the custom interceptor used to gathering timestamp data for retry testing middleware. It instead adds an optional function parameter to the existing `interceptor.AddUnaryRetryInterceptor` function. Whereas the old interceptor gathered retry timestamp data and used custom response types to return that data to the middleware, the improved setup uses an optional callback in the interceptor to write timestamp data to a channel that is read by a goroutine started when the middleware is constructed. This allows us to use the same interceptor in production and under test without compromising performance in production and ensuring that test and production interceptor behavior won't drift.

Removing the custom error types that needed to be processed in `OnResponse` also allowed the removal of any and all potential error handling from `OnResponse` handlers, which gets rid of a potential point of fragility and confusion for authors of middleware.

This commit also fixes the response handling portion of the middleware layer (and removes the last bit of unnecessary state from the `requester` objects) by returning Momento responses from `interpretGrpcResponse` instead of locking them away in a `requester`. This allows the handing Momento responses instead of gRPC responses to the `OnResponse` handlers, as was intended from the beginning.

This commit also makes the middleware responsible for constructing its own metrics collector instead of requiring users to pass in a metrics collector to the middleware, which was non-optimal.

### DIFF
--- a/momento/scs_data_client.go
+++ b/momento/scs_data_client.go
@@ -129,20 +129,15 @@ func (client scsDataClient) applyMiddlewareResponseHandlers(
 	r requester,
 	middlewareRequestHandlers []middleware.RequestHandler,
 	resp grpcResponse,
-	requestError error,
 	responseMetadata []metadata.MD,
 ) (grpcResponse, error) {
-	var newResp interface{}
 	for i := len(middlewareRequestHandlers) - 1; i >= 0; i-- {
 		var requestHandlerError error
 		rh := middlewareRequestHandlers[i]
-		newResp, requestHandlerError = rh.OnResponse(resp, requestError)
-		if requestHandlerError != nil {
+		newResp, err := rh.OnResponse(resp)
+		if err != nil {
 			return nil, momentoerrors.ConvertSvcErr(requestHandlerError, responseMetadata...)
 		}
-
-		// The request handler returned a nil error, so we nil out the original error.
-		requestError = nil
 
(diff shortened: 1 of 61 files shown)


anitarua (approved):
I think it looks great! I think the channel-based approach makes sense.

My only nit is to rename `RetryMetricsMiddleware` to `MomentoLocalMiddleware` to be consistent with the JS and Java sdks (but maybe after this is merged with the consolidated-middleware branch)

momento/scs_data_client.go
```diff
@@ -123,56 +123,33 @@ func (client scsDataClient) applyMiddlewareRequestHandlers(
...
-	// we need to convert it to a Momento service error before returning it back.
-	if requestError != nil {
-		requestError = momentoerrors.ConvertSvcErr(requestError, responseMetadata...)
-	}
-	return resp, requestError
+	return resp, nil
 }
 
-func (client scsDataClient) makeRequest(ctx context.Context, r requester) error {
+func (client scsDataClient) makeRequest(ctx context.Context, r requester) (interface{}, error) {
```
anitarua, line 145: I think it would be nice in a future PR to be able to avoid returning `interface{}` from `makeRequest`, maybe using generics or by creating a `MomentoResponse` interface they all implement or smth

malandis (approved):

PR #609, opened 2025-03-26
chore: fix make target in github workflow

pgautier404:
This commit fixes the make target to run tests on push to main.

### DIFF
--- a/.github/workflows/on-push-to-main.yml
+++ b/.github/workflows/on-push-to-main.yml
@@ -82,10 +82,6 @@ jobs:
       - name: Build examples
         run: make build-examples
 
-      - name: Start Momento Local
-        run: |
-          docker run -d -p 8080:8080 gomomento/momento-local
-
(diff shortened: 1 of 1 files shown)


anitarua (commented):
prod-test doesn't include momento-local tests, right? ([ref](https://github.com/momentohq/client-sdk-go/blob/main/Makefile#L118-L120))

could remove the docker run command too then

pgautier404:
> prod-test doesn't include momento-local tests, right? ([ref](https://github.com/momentohq/client-sdk-go/blob/main/Makefile#L118-L120))
> 
> could remove the docker run command too then

Good catch, thanks. I've removed the docker command and am personally fine with skipping momento local tests for this workflow. If you have any concerns about that, though, I could add them here as well.

anitarua (approved):
yeah looks good, I think we only include momento-local tests in on-pull-request in other SDKs anyway

PR #610, opened 2025-03-29
feat: topics retries

pgautier404:
This commit adds support for middleware for the topic client. The middleware layer provides a way to add header metadata to requests to momento-local as well as a way to register a callback to handle the various events a topic subscription might encounter (heartbeat, item, discontinuity, error, reconnect). The MomentoLocalMiddleware has been extended to implement the `TopicEventCallbackMiddleware` interface, and it registers a simple callback to count each type of event that the subscription encounters. Finally, a number of tests have been added to verify that the topic subscription reconnection logic is working as intended.


config/topics_configurations.go
```diff
@@ -22,5 +24,9 @@ func TopicsDefaultWithLogger(loggerFactory logger.MomentoLoggerFactory) TopicsCo
 				client_timeout: 5 * time.Second,
 			}),
 		}),
+		RetryStrategy: retry.NewFixedCountRetryStrategy(retry.FixedCountRetryStrategyProps{
```
anitarua, line 27: should we keep the default behavior today using an "AlwaysRetryStrategy"?

momento/topic_subscription.go
```diff
@@ -187,16 +204,42 @@ func (s *topicSubscription) Event(ctx context.Context) (TopicEvent, error) {
...
+			GrpcMethod:     "/cache_client.pubsub.Pubsub/Subscribe",
+			AttemptNumber:  attempt,
+		})
+
+		if retryBackoffTime == nil {
+			s.log.Warn("Retry strategy determined that we should not retry, returning error")
+			return err
+		}
+
+		s.onTopicEvent("Subscribe", middleware.RECONNECT)
```
anitarua, line 235: should this event be sent later after the reconnect is confirmed?

momento/test_helpers/momento_local_middleware.go
```diff
@@ -108,53 +67,118 @@ func NewMomentoLocalMiddleware(props MomentoLocalMiddlewareProps) middleware.Mid
...
+		requestMetadata["return-error"] = *mw.metadataProps.ReturnError
+	}
+
+	if mw.metadataProps.ErrorRpcList != nil {
+		requestMetadata["error-rpcs"] = strings.Join(*mw.metadataProps.ErrorRpcList, " ")
+	}
+
+	if mw.metadataProps.ErrorCount != nil {
+		requestMetadata["error-count"] = fmt.Sprintf("%d", *mw.metadataProps.ErrorCount)
+	}
```
anitarua, line 159: these are valid for subscribe requests too, would just affect the initial subscription establishment iirc

anitarua (commented):
some comments from initial pass, will pull branch down for more in-depth review soon

momento/topic_subscription.go
pgautier404: This is tracking reconnect attempts and needs to account for unsuccessful attempts as well.

config/topics_configurations.go
pgautier404: Well this was a great suggestion :-) It perfectly addresses the backwards compat concerns we discussed yesterday. Thanks for the excellent idea!

anitarua (dismissed):
lgtm!

config/retry/always_retry_strategy.go
```diff
@@ -0,0 +1,68 @@
...
+func (r *alwaysRetryStrategy) String() string {
+	return fmt.Sprintf(
+		"alwaysRetryStrategy{reconnectMs: %d, log: %s}\n",
+		*r.reconnectMs,
+		r.log,
+	)
+}
+
+// NewAlwaysRetryStrategy is a retry strategy that always retries any request after a fixed delay.
+// This maintains compatibility with the legacy behavior of the client, but is not the optimal
```
malandis, line 50: To clarify the comment: on the topic client for subscriptions.

config/topics_configurations.go
```diff
@@ -15,12 +17,17 @@ func TopicsDefault() TopicsConfiguration {
...
 		LoggerFactory: loggerFactory,
 		TransportStrategy: NewTopicsStaticTransportStrategy(&TopicsTransportStrategyProps{
 			GrpcConfiguration: NewTopicsStaticGrpcConfiguration(&TopicsGrpcConfigurationProps{
 				client_timeout: 5 * time.Second,
 			}),
 		}),
+		RetryStrategy: retry.NewAlwaysRetryStrategy(retry.AlwaysRetryStrategyProps{
+			LoggerFactory: loggerFactory,
+			ReconnectMs:   &reconnectMs,
+		}),
```
malandis, line 31: We should clarify this is the subscription retry strategy. There are no retries for publishes (and if there were, the settings would likely be different than those for subscriptions)

malandis (commented):
Looks good. Just recommend a rename on the retry strategy to clarify it's for subscriptions.

config/retry/always_retry_strategy.go
pgautier404: Good callout . . . I'm actually not sufficiently separating concerns here. A retry strategy is broadly applicable to any operation that can fail and be tried again, so mentioning "the client" here is really ambiguous. I'll fix up the comment to make it more clear that this is only intended for a specific use case. I'm also going to go ahead and deprecate it since we really should discourage people from casually using it. Let me know what you think.

config/topics_configurations.go
pgautier404: I actually think this would be good as it is if I just renamed `ReconnectMs` to `RetryMs`. If I were to write an `AlwaysRetryStrategy` for the data client (or for topic publish for that matter), it would be identical to this one inasmuch as returning a single constant forever is the completely naive solution to `DetermineWhenToRetry()`, the single function in the `Strategy` interface.

config/topics_config.go
```diff
@@ -27,6 +32,11 @@ type TopicsConfigurationProps struct {
 	// TransportStrategy is responsible for configuring network tunables for the GRPC client,
 	// including the number of stream and unary grpc channels that should be used.
 	TransportStrategy TopicsTransportStrategy
+
+	RetryStrategy retry.Strategy
```
anitarua, line 36: should there be a comment explaining RetryStrategy on the topic client applies only to subscriptions?

config/topics_configurations.go
malandis: I understand the interface is the same. I mean that the settings for subscription retries are unlike any unary API we have -- it retries forever. If we were to add a strategy for publishes, it would certainly be a different strategy. So better to call this one something specific to subscriptions.

pgautier404: I have deprecated this as a strategy that no client going forward would ever use under any circumstances and recommended that users select a different strategy. If we were to add a strategy for *anything* it would not be this one. To your point, the naming should reflect the specificity of its use, so I've renamed it `LegacyTopicSubscriptionRetryStrategy`, which should render it sufficiently radioactive :-)

config/topics_config.go
pgautier404: Sure, I can add something.

anitarua (approved):

malandis (dismissed):
🎉

PR #612, opened 2025-03-31
chore: add integration tests for new get/set if hash methods

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1209

To avoid canary failures, we'll merge these new integration tests after we've confirmed the new apis have been rolled out to all prod cells.

### DIFF
--- a/momento/scalar_with_hash_test.go
+++ b/momento/scalar_with_hash_test.go
@@ -1,0 +1,518 @@
+package momento_test
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


rishtigupta:
do we need to add them to makefile for canaries to pick up these tests or are they included in an existing test-target?

anitarua:
they should be included in cache tests due to the `Label(CACHE_SERVICE_LABEL)` in the test. I also double checked that CI ran them too as part of the cache tests:

Screenshot 2025-04-21 at 11 23 14 AM

PR #616, opened 2025-04-03
feat: add FixedTimeoutRetryStrategy and tests

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1205

- Adds FixedTimeoutRetryStrategy and tests
- Updated timestamp collection to be unix time in milliseconds instead of seconds to better support the new tests
- Fix: momento-local middleware was sending metadata header `delay-millis` instead of `delay-ms`

Because go sdk uses context.WithTimeout / context.WithDeadline instead of a grpc call option, we are responsible for catching a retry's deadline exceeded error and passing along a new context object with updated deadline.

### DIFF
--- a/config/retry/fixed_timeout_retry_strategy.go
+++ b/config/retry/fixed_timeout_retry_strategy.go
@@ -33,10 +33,6 @@ type FixedTimeoutRetryStrategy interface {
 	WithRetryTimeoutMillis(timeout int) Strategy
 	WithRetryDelayIntervalMillis(delay int) Strategy
 	WithEligibilityStrategy(s EligibilityStrategy) Strategy
-
-	// CalculateRetryDeadline calculates the deadline for a retry attempt based on retryTimeoutMillis,
-	// but clips it to the overall deadline if the overall deadline is sooner.
-	CalculateRetryDeadline(overallDeadline time.Time) time.Time
 }
 
 type FixedTimeoutRetryStrategyProps struct {
@@ -101,13 +97,15 @@ func (r *fixedTimeoutRetryStrategy) WithEligibilityStrategy(strategy Eligibility
 	}
 }
 
-func (r *fixedTimeoutRetryStrategy) CalculateRetryDeadline(overallDeadline time.Time) time.Time {
+// CalculateRetryDeadline calculates the deadline for a retry attempt using the retry timeout,
+// but clips it to the overall deadline if the overall deadline is sooner.
+func (r *fixedTimeoutRetryStrategy) CalculateRetryDeadline(overallDeadline time.Time) *time.Time {
 	deadlineOffset := time.Duration(r.retryTimeoutMillis) * time.Millisecond
 	deadline := time.Now().Add(deadlineOffset)
 	if deadline.After(overallDeadline) {
 		deadline = overallDeadline
 	}
-	return deadline
+	return &deadline
 }
 
 func (r *fixedTimeoutRetryStrategy) DetermineWhenToRetry(props StrategyProps) *int {
--- a/config/retry/fixed_timeout_retry_strategy.go
+++ b/config/retry/fixed_timeout_retry_strategy.go
@@ -3,8 +3,10 @@ package retry
 import (
 	"fmt"
 	"math/rand"
+	"time"
 
 	"github.com/momentohq/client-sdk-go/config/logger"
+	"google.golang.org/grpc/codes"
 )
 
 const (
@@ -88,12 +90,30 @@ func (r *fixedTimeoutRetryStrategy) WithEligibilityStrategy(strategy Eligibility
 	}
 }
 
+func (r *fixedTimeoutRetryStrategy) GetResponseDataReceivedTimeoutMillis() *int {
+	return &r.responseDataReceivedTimeoutMillis
+}
+
 func (r *fixedTimeoutRetryStrategy) DetermineWhenToRetry(props StrategyProps) *int {
 	r.log.Debug(
 		"Determining whether request is eligible for retry; status code: %s, "+
 			"request type: %s, attemptNumber: %d",
 		props.GrpcStatusCode, props.GrpcMethod, props.AttemptNumber,
 	)
+
+	// If a retry attempt's timeout has passed but the client's overall timeout has not yet passed,
+	// we should reset the deadline and retry.
+	if props.AttemptNumber > 0 && props.GrpcStatusCode == codes.DeadlineExceeded && props.OverallDeadline.After(time.Now()) {
+		timeoutWithJitter := addJitter(r.retryDelayIntervalMillis)
+		r.log.Debug(
+			"Determined request is retryable; retrying after %d ms: [method: %s, status: %s, attempt: %d]",
+			timeoutWithJitter,
+			props.GrpcMethod,
+			props.GrpcStatusCode.String(),
+			props.AttemptNumber,
+		)
+		return &timeoutWithJitter
+	}
 
 	if !r.eligibilityStrategy.IsEligibleForRetry(props) {
 		r.log.Debug(
--- a/config/retry/fixed_timeout_retry_strategy.go
+++ b/config/retry/fixed_timeout_retry_strategy.go
@@ -1,0 +1,128 @@
+package retry
+
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/momentohq/client-sdk-go/config/logger"
+)
+
+const (
+	DefaultResponseDataReceivedTimeoutMillis = 1000 // 1 second default timeout for retry attempts
+	DefaultRetryDelayIntervalMillis          = 100  // Schedule retry attempt for 100ms later +/- jitter
+)
+
+type fixedTimeoutRetryStrategy struct {
+	eligibilityStrategy               EligibilityStrategy
+	log                               logger.MomentoLogger
+	responseDataReceivedTimeoutMillis int
+	retryDelayIntervalMillis          int
+}
+
+type FixedTimeoutRetryStrategy interface {
+	Strategy
+	WithResponseDataReceivedTimeoutMillis(timeout int) Strategy
+	WithRetryDelayIntervalMillis(delay int) Strategy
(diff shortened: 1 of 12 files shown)


momento/retry_test.go
```diff
@@ -545,6 +551,277 @@ var _ = Describe("retry eligibility-strategy", Label(RETRY_LABEL, MOMENTO_LOCAL_
...
+
+				// Should immediately receive errors and retry every DefaultRetryDelayIntervalMillis
+				// until the client timeout is reached.
+				maxAttempts := CLIENT_TIMEOUT_MILLIS / retry.DefaultRetryDelayIntervalMillis
+				Expect(metricsCollector.GetTotalRetryCount(cacheName, "Get")).To(BeNumerically("<=", maxAttempts))
+				Expect(metricsCollector.GetTotalRetryCount(cacheName, "Get")).To(BeNumerically(">", 0))
+
+				// Jitter will be +/- 10% of the retry delay interval
+				Expect(metricsCollector.GetAverageTimeBetweenRetries(cacheName, "Get")).To(BeNumerically("<=", retry.DefaultRetryDelayIntervalMillis*1.1))
+				Expect(metricsCollector.GetAverageTimeBetweenRetries(cacheName, "Get")).To(BeNumerically(">=", retry.DefaultRetryDelayIntervalMillis*0.9))
```
pgautier404, line 648: This is cool :-)

momento/test_helpers/retry_metrics_collector.go
```diff
@@ -41,10 +41,8 @@ func (r *retryMetrics) GetTotalRetryCount(cacheName string, requestName string)
 	return 0, fmt.Errorf("request name '%s' is not valid", requestName)
 }
 
-// GetAverageTimeBetweenRetries returns the average time between retries in seconds.
-//
-//	Limited to second resolution, but I can obviously change that if desired.
-//	This tracks with the JS implementation.
+// GetAverageTimeBetweenRetries returns the average time between retries in milliseconds.
+// OnInterceptorRequest in momento_local_middleware.go records Unix timestamps in milliseconds.
```
pgautier404, line 45: 👍

pgautier404 (dismissed):
🚢

internal/interceptor/retry_interceptor.go
```diff
@@ -14,13 +14,32 @@ import (
...
 		for {
 			// This is currently used for testing purposes only by the RetryMetricsMiddleware.
 			if onRequest != nil {
 				onRequest(ctx, method)
 			}
+
+			// If the FixedTimeoutRetryStrategy is used, overwrite the deadline using
+			// the configured retry timeout. Otherwise, use the given context.
+			retryCtx := ctx
+			if s, ok := s.(retry.FixedTimeoutRetryStrategy); ok {
```
anitarua, line 34: Then the retry interceptor need only check here to see if the context's deadline needs to be adjusted due to this one retry strategy

config/retry/fixed_timeout_retry_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+func (r *fixedTimeoutRetryStrategy) DetermineWhenToRetry(props StrategyProps) *int {
+	r.log.Debug(
+		"Determining whether request is eligible for retry; status code: %s, "+
+			"request type: %s, attemptNumber: %d",
+		props.GrpcStatusCode, props.GrpcMethod, props.AttemptNumber,
+	)
+
+	// If a retry attempt's timeout has passed but the client's overall timeout has not yet passed,
+	// we should reset the deadline and retry.
+	if props.AttemptNumber > 0 && props.GrpcStatusCode == codes.DeadlineExceeded && props.OverallDeadline.After(time.Now()) {
```
anitarua, line 115: couldn't find a way around passing in the overall deadline here, but I added a comment to explain why it's there

config/retry/fixed_timeout_retry_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+type FixedTimeoutRetryStrategy interface {
+	Strategy
+	WithResponseDataReceivedTimeoutMillis(timeout int) Strategy
+	WithRetryDelayIntervalMillis(delay int) Strategy
+	WithEligibilityStrategy(s EligibilityStrategy) Strategy
+
+	// CalculateRetryDeadline calculates the deadline for a retry attempt based on responseDataReceivedTimeoutMillis,
+	// but clips it to the overall deadline if the overall deadline is sooner.
+	CalculateRetryDeadline(overallDeadline time.Time) time.Time
+}
```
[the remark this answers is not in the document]
anitarua: And I was able to get rid of the `GetResponseDataReceivedTimeoutMillis()` method entirely as a result

config/retry/fixed_timeout_retry_strategy.go
```diff
@@ -0,0 +1,157 @@
...
+
+type FixedTimeoutRetryStrategyProps struct {
+	EligibilityStrategy               EligibilityStrategy
+	LoggerFactory                     logger.MomentoLoggerFactory
+	ResponseDataReceivedTimeoutMillis int
```
anitarua, line 38: I had named it `ResponseDataReceivedTimeoutMillis` to be consistent with the JS sdk but it could be renamed `RetryTimeoutMillis` or similar instead?

config/retry/retry.go
```diff
@@ -17,4 +20,7 @@ type Strategy interface {
 	//
 	// Returns The time in milliseconds before the next retry should occur or nil if no retry should be attempted.
 	DetermineWhenToRetry(props StrategyProps) *int
+
+	// GetResponseDataReceivedTimeoutMillis returns the timeout for a retry attemptin milliseconds.
+	GetResponseDataReceivedTimeoutMillis() *int
```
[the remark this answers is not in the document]
anitarua: Good suggestion, thanks! 

Was still using my JS brain and there was some reason why we had to add that method to the overall retry strategy interface and I think I had assumed the same here 😅

config/retry/fixed_timeout_retry_strategy.go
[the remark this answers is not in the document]
pgautier404: My only concern there is that it doesn't give us a reusable interface that future middleware could also implement. However, I can't really think of another general application for `CalculateRetryDeadline` off the top of my head, and if one ever popped up we could easily refactor into a new reusable interface. I think keeping it simple is the right choice here, and I'm super happy you could get rid of `GetResponseDataReceivedTimeoutMillis()`!!!

config/retry/fixed_timeout_retry_strategy.go
pgautier404: I would not be mad at a simpler name for that, and your suggestion makes it clearer that it is the *retry* that would be timing out here. It would be helpful to copy over the doc comments from the JS SDK here. This really is just kind of a confusing middleware, so the more comments the better :-)

internal/interceptor/retry_interceptor.go
```diff
@@ -14,13 +14,32 @@ import (
 func AddUnaryRetryInterceptor(s retry.Strategy, onRequest func(context.Context, string)) func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 	return func(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
 		attempt := 1
+		var overallDeadline time.Time
+		deadline, ok := ctx.Deadline()
+		if ok {
+			overallDeadline = deadline
+		} else {
+			overallDeadline = time.Now().Add(5 * time.Second)
```
pgautier404, line 22: When would this condition be met? Did you find you were being passed contexts without deadlines? I'm fine with providing a default deadline if it's necessary in some cases, but ideally this value would match the client deadline from the active client configuration. We could add another argument to `AddUnaryRetryInterceptor` and pass in the `request.GrpcConfiguration.GetDeadline()` value when adding the interceptor in `data_manager`.

That said, I don't think it's an emergency and it would be fine to look into that as a followup (or open an issue for it). I don't want to put it off too long, though, as it could produce unexpected behavior for users.

anitarua: `ctx.Deadline()` returns `(deadline time.Time, ok bool)` and I figured it was safer to handle the `ok` and `!ok` cases explicitly. I think most of the time, `deadline` will exist, but just in case it doesn't, I set it to the usual default deadline. (I think I couldn't import an existing const from somewhere due to circular imports but can extract the const into another package maybe?).

anitarua: and yeah true, it should match the deadline from the config if nothing else, let me take another look into this

internal/interceptor/retry_interceptor.go
anitarua: Michael [commented](https://github.com/momentohq/client-sdk-dotnet/pull/606#discussion_r2034196622) in my dotnet PR that retry interceptor/middleware shouldn't know about the concrete implementations of retry strategies.
Now I'm thinking perhaps a modified CalculateRetryDeadline approach instead like he suggested would work here too? And we could then also avoid having to check explicitly for a concrete impl?

pgautier404: That's kind of what I was getting at when I suggested using a more general interface here for the type assertion instead of the specific type of your `FixedTimeoutRetryStrategy`. I would rather do some sort of duck typing here than have **every** retry strategy need to implement a `CalculateRetryDeadline` method that all but one of them is going to return `nil` for. We can't provide a default implementation for `CalculateRetryDeadline` unless we do something like the kind of composition I did with the middleware request handlers where we use a "base" handler with default method implementations to help compose the actual handler functionality.

I think it would be ideal to be able to get rid of the type assertion. This approach was a compromise motivated by the fact that go just doesn't have base classes to provide override-able default implementations. If you'd like to try to set it up so that we don't have to define a `CalculateRetryDeadline` function in strategies that don't need it, I would definitely not be opposed.

pgautier404 (approved):
This looks great, thank you!

PR #618, opened 2025-04-04
chore: add compression middleware example

anitarua:


examples/compression/compression_middleware.go
```diff
@@ -0,0 +1,101 @@
...
+		compressed := rh.encoder.EncodeAll([]byte(fmt.Sprintf("%v", rawData)), nil)
+		rh.GetLogger().Info("Compressed request %T: %d bytes -> %d bytes", req, len(fmt.Sprintf("%v", rawData)), len(compressed))
+		return &momento.SetIfAbsentOrHashEqualRequest{
+			CacheName: r.CacheName,
+			Key:       r.Key,
+			Value:     momento.String(compressed),
+			HashEqual: r.HashEqual,
+		}, nil
+	default:
+		rh.GetLogger().Info("[compressionMiddleware] No action for OnRequest type: %T", req)
```
pgautier404, line 65:
```suggestion
		rh.GetLogger().Info("No action for OnRequest type: %T", req)
```

The logger will take care of that.

examples/compression/compression_middleware.go
```diff
@@ -0,0 +1,101 @@
...
+			return nil, fmt.Errorf("failed to decompress response: %v", err)
+		}
+		rh.GetLogger().Info(
+			"Decompressed response %T: %d bytes -> %d bytes",
+			resp, len(rawData), len(decompressed),
+		)
+		newGetWithHashResponse := responses.NewGetWithHashHit(decompressed, r.HashByte())
+		return newGetWithHashResponse, nil
+	default:
+		rh.GetLogger().Info("[compressionMiddleware] No action for OnResponse type: %T", resp)
```
pgautier404, line 98:
```suggestion
		rh.GetLogger().Info("No action for OnResponse type: %T", resp)
```

examples/compression/compression_middleware.go
anitarua: ah that's right, was leftover from testing with fmt.Printf statements

examples/compression/main.go
```diff
@@ -0,0 +1,142 @@
...
+		CacheName: cacheName,
+		Key:       momento.String("key-1"),
+	})
+	if err != nil {
+		momentoError, ok := err.(momento.MomentoError)
+		if ok {
+			log.Printf("[GetWithHash] Error: %s\n", momentoError.Message())
+		} else {
+			log.Printf("[GetWithHash] Error: %s\n", err.Error())
+		}
```
pgautier404, line 95:
```suggestion
		log.Printf("[GetWithHash] Error: %s\n", err)
```

I think we can get away with just stringifying whatever error we get here to keep this simple.

pgautier404 (approved):
:shipit:

PR #628, opened 2025-04-14
feat: add base and gzip compression middleware for scalar get and set requests

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1219

Creates a `CompressionMiddleware` interface that can accept a `CompressionStrategy` for gzip or zstd. 
The gzip version is folded into the `config/middleware/impl` folder since gzip is part of the go standard library.
The zstd version will be published as a different module named `client-sdk-go-compression-zstd` (mirroring the JS version) so that users can opt into the zstd dependency. The zstd package includes some tests to verify that it works.

Enabling gzip compression would look like:
```go
config := config.LaptopLatestWithLogger(loggerFactory).WithMiddleware(
		[]middleware.Middleware{
			impl.NewGzipCompressionMiddleware(impl.GzipCompressionMiddlewareProps{
				CompressionStrategyProps: compression.CompressionStrategyProps{
					CompressionLevel: compression.CompressionLevelDefault,
					Logger:           loggerFactory.GetLogger("compression-middleware"),
				},
			}),
		},
	)
```

<details>
<summary>First draft</summary>

Basically moves the zstd compression example into the sdk proper under the `config/middlewares/impl` folder with some revisions:

- renamed to ZstdCompressionMiddleware for clarity in case we add another compression algorithm in future
- NewZstdCompressionMiddleware accepts a CompressionLevel parameter now
- OnRequest updated to compress Set and SetIf requests by default
- OnResponse updated to decompress Get and GetWithHash requests by default
- Added some happy-path tests to verify that adding compression middleware will not break get/set requests.
    - Currently no way to verify that values are actually compressed besides viewing logs. Would probably require another middleware to pass compressed size to a channel and stuff.

Creating a client with this compression middleware with most basic options would look like:
```go
myConfig := config.LaptopLatestWithLogger(loggerFactory).WithMiddleware(
    []middleware.Middleware{
	    impl.NewZstdCompressionMiddleware(impl.ZstdCompressionMiddlewareProps{
		    Logger:           loggerFactory.GetLogger("compression-middleware"),
		    CompressionLevel: zstd.SpeedBestCompression,
	    }),
    },
)
```

If someone wants to compress only certain requests, then they could specify the `IncludeTypes` option:
```go
myConfig := config.LaptopLatestWithLogger(loggerFactory).WithMiddleware(
    []middleware.Middleware{
	    impl.NewZstdCompressionMiddleware(impl.ZstdCompressionMiddlewareProps{
		    Logger:           loggerFactory.GetLogger("compression-middleware"),
		    CompressionLevel: zstd.SpeedBestCompression,
		    IncludeTypes: []interface{}{
			    momento.SetRequest{},
			    momento.GetRequest{},
			    momento.SetIfAbsentOrHashEqualRequest{},
			    momento.GetWithHashRequest{},
		    },
	    }),
    },
)
```
</details>


config/middleware/impl/zstd_compression.go
```diff
@@ -0,0 +1,152 @@
...
+type zstdCompressionMiddleware struct {
+	middleware.Middleware
+	encoder *zstd.Encoder
+	decoder *zstd.Decoder
+}
+
+type ZstdCompressionMiddlewareProps struct {
+	Logger           logger.MomentoLogger
+	IncludeTypes     []interface{}
+	CompressionLevel zstd.EncoderLevel
```
malandis, line 22: We should create an enum of our own to not expose the third party library types.

config/middleware/impl/zstd_compression.go
```diff
@@ -0,0 +1,152 @@
...
+func (rh *zstdCompressionMiddlewareRequestHandler) OnResponse(resp interface{}) (interface{}, error) {
+	// We still need to use a switch statement to be able to access the response objects
+	// as the specific response types in order to access the Value field.
+	switch r := resp.(type) {
+	case *responses.GetHit:
+		decompressed, err := rh.decompress("GetHit", r.ValueByte())
+		if err != nil {
+			return nil, fmt.Errorf("failed to decompress response: %v", err)
+		}
+		return responses.NewGetHit(decompressed), nil
```
malandis, line 141: Should this be a failure or should we just pass through? I would use the JS SDK as a reference. For the zstd plugin, we peek at the data to see if it has the magic prefix indicating it was compressed, then we try decompressing; otherwise pass through (since it must not have been compressed). cc @nand4011

malandis (commented):
We should follow the JS pattern of modeling a compression plugin, then make a zstd implementation that is a new package. With that and the comments below should be gtg 👍

malandis (commented):
The shape looks good to me. Are we testing this as well?

anitarua:
@malandis There are some tests for zstd, but not yet for gzip. Should I add for gzip too?

malandis:
> @malandis There are some tests for zstd, but not yet for gzip. Should I add for gzip too?

We should specifically test the scenarios where we selectively decompress -- when the original data wasn't compressed, the application returns the data as is and doesn't error out.

We should also trace log when not decompressing because the magic symbol doesn't match.

client-sdk-go-compression-zstd/zstd_compression.go
```diff
@@ -44,10 +44,12 @@ func (c zstdCompressor) Compress(data []byte) ([]byte, error) {
 	return c.encoder.EncodeAll(data, nil), nil
 }
 
-func (c zstdCompressor) Decompress(data []byte) ([]byte, error) {
+func (c zstdCompressor) Decompress(data []byte, logger logger.MomentoLogger) ([]byte, error) {
```
malandis: We should have the logger be part of a prop passed to the constructor as opposed to part of the interface.

config/compression/compression_strategy.go
```diff
@@ -3,8 +3,8 @@ package compression
 import "github.com/momentohq/client-sdk-go/config/logger"
 
 type CompressionStrategy interface {
-	Compress(data []byte) ([]byte, error)
-	Decompress(data []byte, logger logger.MomentoLogger) ([]byte, error)
+	Compress(data []byte, logger logger.MomentoLogger, requestType string) ([]byte, error)
+	Decompress(data []byte, logger logger.MomentoLogger, requestType string) ([]byte, error)
```
malandis, line 7: Not a fan of passing the request type here. The compression middleware can be aware of that but the compression strategy doesn't need to be.

config/middleware/impl/gzip_compression.go
```diff
@@ -36,37 +36,47 @@ type gzipCompressor struct {
...
 		return nil, err
 	}
 
 	err = gzWriter.Close()
 	if err != nil {
+		logger.Error("Failed to close gzip writer: %v", err)
 		return nil, err
 	}
 
+	logger.Info("Compressed request %s: %d bytes -> %d bytes", requestType, len(data), buf.Len())
```
malandis, line 59: This should be debug/trace.

config/middleware/impl/gzip_compression.go
```diff
@@ -36,37 +36,47 @@ type gzipCompressor struct {
...
 
 	if isGzipCompressed(data) {
 		logger.Trace("Decompressing gzip compressed data")
-		return io.ReadAll(reader)
+		decompressed, err := io.ReadAll(reader)
+		if err != nil {
+			logger.Error("Failed to read decompressed data: %v", err)
+			return nil, err
+		}
+		logger.Info("Decompressed response %s: %d bytes -> %d bytes", requestType, len(data), len(decompressed))
```
malandis, line 78: debug/trace

config/compression/compression_strategy.go
anitarua: it's was just used for logging, can remove

pgautier404 (approved):
:shipit:

PR #640, opened 2025-04-18
feat: add base momento request and response interfaces

anitarua:
Work towards https://github.com/momentohq/dev-eco-issue-tracker/issues/1221

Opening this draft for early review!

Goal: add base type for momento request and response objects to minimize usage of `interface{}` (basically an `any` type), including when specifying what requests middlewares should act on (`IncludeTypes` parameter).

- Added new interfaces in `base_response_types.go` and updated the response objects' interfaces to include the base interface too. Excluded storage-related objects from this effort since it's being deprecated.
- Added new interfaces in new package `momento_request_base` in `momento/request_base` to avoid circular import issues. Added the `MomentoCacheRequest` base interface to the internal `requester` interface and updated cache data client request objects accordingly. 
- Updated middleware interface to accept `IncludeTypes` parameter as type `[]momento_response_base.MomentoCacheRequest`, but I think this is a breaking change. I don't think anyone is passing in this argument yet to their middlewares (if they're even using middlewares), but per recent team discussions, I don't want to cavalierly make unnecessary breaking changes.
- Updated middleware interface so that `OnRequest` and `OnResponse` can also make use of the new base types where appropriate to reduce usage of `interface{}`.

Tested locally with the middlewares-related examples and verified that updating the type of `IncludeTypes` works.

### DIFF
--- a/momento/request_base/base_request_types.go
+++ b/momento/request_base/base_request_types.go
@@ -1,0 +1,7 @@
+package momento_request_base
+
+type MomentoCacheRequest interface {
+	GetRequestName() string
+}
+
+type MomentoTopicRequest interface{}
--- a/responses/base_response_types.go
+++ b/responses/base_response_types.go
@@ -1,0 +1,9 @@
+package responses
+
+type MomentoCacheResponse interface{}
+
+type MomentoTopicResponse interface{}
+
+type MomentoLeaderboardResponse interface{}
+
+type MomentoAuthResponse interface{}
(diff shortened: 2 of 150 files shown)


momento/request_base/base_request_types.go
```diff
@@ -0,0 +1,7 @@
+package momento_request_base
+
+type MomentoCacheRequest interface {
+	GetRequestName() string
+}
+
+type MomentoTopicRequest interface{}
```
anitarua, line 7: base request type specified here

I added one for topic too but maybe not needed right now

responses/base_response_types.go
```diff
@@ -0,0 +1,9 @@
+package responses
+
+type MomentoCacheResponse interface{}
+
+type MomentoTopicResponse interface{}
+
+type MomentoLeaderboardResponse interface{}
+
+type MomentoAuthResponse interface{}
```
anitarua, line 9: added all the base response types here

was a more natural change for the responses since each already had a base interface type that then branched into Success/Error, Hit/Miss, etc

figured it could be handy to have responses categorized by client type, but can remove all but cache if preferred

PR #642, opened 2025-04-22
feat: support WithMaxSubscriptions config on topic client

anitarua:
Closes https://github.com/momentohq/dev-eco-issue-tracker/issues/1163

Implemented changes to support `WithMaxSubscriptions()` that should create one grpc channel to start and dynamically add more as subscriptions increased until the maximum number of channels is reached. The maximum number of channels equals `maxSubscriptions / 100`.

**Internal changes**

- `internal/grpcmanagers/topic_manager.go`: cleaned up grpc channel bookkeeping by creating `StaticUnaryManagerList`, `StaticStreamManagerList`, and `DynamicStreamManagerList`. Each struct is responsible for returning a grpc topic manager when requested. The static ones initialize the pool (list) of grpc channels once, then the list becomes read-only. The dynamic one contains a read-write lock in order to be thread-safe when adding a new grpc channel.
    - Edit: The dynamic and static stream manager lists now use an [unbuffered channel](https://go.dev/doc/effective_go#channels) to gatekeep / limit concurrency. The list objects will continually put the next available stream grpc manager on the channel for a pubsub client to pull as needed to service a burst of subscriptions without concurrency issues. As a result, the read-write lock no longer seemed necessary.
- `momento/pubsub_client.go`: simplified the internal pubsub client now that the bookkeeping is better encapsulated by the manager list structs. It only needs to read the correct configs upon initialization based on whether it received a config with `maxSubscriptions` or `numStreamChannels` defined on it

**Public-facing changes**

Creating a `TopicsDynamicTransportStrategy` with `MaxSubscriptions` option vs `TopicsStaticTransportStrategy` with `NumStreamChannels` option would be the most correct implementation, but I found it led to a cascade of many breaking changes due to the existing topics config structure. 

This PR currently contains minimal changes to the topics config code in order to support a `WithMaxSubscriptions` method without breaking backwards compatibility for our current golang topics users. The `maxSubscriptions` concept exists only at the `TopicConfiguration` level, not the transport or grpc config levels.

- Breaking change: moved the `const MAX_CONCURRENT_STREAMS_PER_CHANNEL` from `momento` package (in `momento/pubsub_client.go`) to `config` package in `config/topics_config.go`. Not sure if anyone but us was using it, but still a breaking change. Currently used only by the topic grpc manager, but keeping it in the `momento` package would have caused an `import cycle not allowed` error. Also had changed the type from int to uint32 but could stick with int if needed.
- Non-breaking change(?): un-deprecates `TopicConfiguration.GetMaxSubscriptions()`
- Non-breaking change(?): un-deprecates `TopicConfiguration.WithMaxSubscriptions()` -- sets `maxSubscriptions`, but also if `numStreamChannels` is also set, it'll warn that `maxSubscriptions` takes precedence
- Non-breaking change: `TopicConfiguration.WithNumStreamGrpcChannels()` -- sets `numStreamChannels`, but also if `maxSubscriptions` is also set, it'll warn that `maxSubscriptions` takes precedence
- Non-breaking change: moves definition of `TopicsStaticGrpcConfiguration` into the `config/topics-grpc-configuration.go` file instead for better organization
- Non-breaking change: marks `GetNumGrpcChannels()` as deprecated since `NumGrpcChannels` is deprecated. The stream/unary versions should be used instead. Removed unnecessary call to `GetNumGrpcChannels()` in the topic client file as well.

**Tests**

- Added a topic client using `WithMaxSubscriptions()` config and copied a test to sanity-check that the topic client can still pub and sub
- Added tests to confirm configs will prefer maxSubscriptions if both or neither `WithMaxSubscription()` and `WithNumStreamGrpcChannels()` are used
- Added `topic_manager_test.go` file to fuzz test the new stream manager list structs. These test run against momento-local so I labeled them as retry tests as well.


internal/grpcmanagers/topic_manager.go
```diff
@@ -47,3 +52,278 @@ func (topicManager *TopicGrpcManager) Close() momentoerrors.MomentoSvcErr {
...
+
+// Helper function to help sanity check number of concurrent streams before starting a new subscription
+func (list *DynamicStreamManagerList) checkNumConcurrentStreams() momentoerrors.MomentoSvcErr {
+	list.rwLock.Lock()
+	defer list.rwLock.Unlock()
+
+	numActiveStreams := list.CountNumberOfActiveSubscriptions()
+	list.logger.Debug("Current number of active subscriptions: %d", numActiveStreams)
+
+	if numActiveStreams >= int64(list.currentMaxConcurrentStreams) && len(list.grpcManagers) >= list.maxManagerCount {
```
malandis, line 274: Is `len` `O(1)` in Go?

anitarua: From quick research, I'm not sure. Regardless, I extract `len(list.grpcManagers)` into a variable since it's used multiple times within `checkNumConcurrentStreams()`

config/topics_config.go
```diff
@@ -168,12 +174,15 @@ func (s *topicsConfiguration) GetNumStreamGrpcChannels() uint32 {
 }
 
 func (s *topicsConfiguration) WithNumStreamGrpcChannels(numStreamGrpcChannels uint32) TopicsConfiguration {
-	// maxSubscriptions and numGrpcChannels are deprecated, not included in the override
+	if s.maxSubscriptions > 0 {
+		s.loggerFactory.GetLogger("TopicsConfiguration").Warn("WithMaxSubscriptions will override NumStreamGrpcChannels. Use only NumStreamGrpcChannels if you want to specify a static number of gRPC channels for subscriptions. Use only WithMaxSubscriptions if you want the topic client to dynamically create gRPC channels to support the number of subscriptions up to the specified limit.")
+	}
```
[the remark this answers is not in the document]
anitarua: I think that's what it's doing already. Calling `WithNumStreamGrpcChannels` would be setting a non-default `NumStreamGrpcChannels`, and `maxSubscriptions > 0` indicates a non-default value, no?

anitarua:
Added fuzz tests for the static and dynamic stream topic manager lists. Haven't encountered deadlocks but they're flaky in situations when a burst of subscriptions is started and sometimes the full number of subscriptions is not reached. Looking into fixing that, putting in draft for now

anitarua:
Update: The dynamic and static stream manager lists now use an [unbuffered channel](https://go.dev/doc/effective_go#channels) to gatekeep / limit concurrency. The list objects will continually put the next available stream grpc manager on the channel for a pubsub client to pull as needed to service a burst of subscriptions without concurrency issues. As a result, the read-write lock no longer seemed necessary and I removed it.

Did some additional testing locally to verify no goroutines are leaked. Was unable to add those goroutine checks to the tests (seemed like the goroutine check for each test was occurring just as another test was starting/running so there were always "leaked" goroutines).

PR #646, opened 2025-06-06
fix: make sure cache client closes all underlying clients upon close

anitarua:
This will prevent leaked goroutines from the cache client


momento/cache_client.go
```diff
@@ -1067,8 +1067,12 @@ func (c defaultScsClient) Ping(ctx context.Context) (responses.PingResponse, err
 }
 
 func (c defaultScsClient) Close() {
+	defer c.pingClient.Close()
 	defer c.controlClient.Close()
 	defer c.getNextDataClient().Close()
+	for _, client := range c.dataClients {
+		defer client.Close()
+	}
```
malandis, line 1075: With your change here do we still need the line above this?

anitarua: oh we wouldn't, meant to remove

Issue #571, opened 2025-01-05
RefreshApiKey returns a JWT for the apiKey, not an actual apiKey


joseph-wortmann:
The call to RefreshApiKey should return an apiKey, but It returns a JWT instead.

This makes It impossible to actually rotate keys, as the response that you get back is not a key.

This behavior occurs for both Super User and Fine Grained keys.

joseph-wortmann:
So I figured this out.

The API Key that the Credential Provider is expecting is actually a base 64 encoded JSON document that contains the "endpoint" and the "api_key", with the latter being a JWT. Both of these are returned separately in the RefreshApiKeySuccess structure as Endpoint and ApiKey, respectively, and must be combined by the user into the base64 encoded JSON as previously mentioned.

I can find nowhere in the documentation that this is called out. It is quite different than the JSON that is given for API key creation in the console. Really needs to be changed or documented.

alubbad:
The bug has been fixed in version 1.38.1.

PR #652, opened 2025-09-03
fix: refreshApiKey changed to match Credential Provider API key"

alubbad:
Solves the following [issue](https://github.com/momentohq/client-sdk-go/issues/571#issue-2769122473).

I changed the ApiKey that is returned by RefreshApiKey to be a 64-bit encoded JSON document in the same manner as GenerateApiKey.


anitarua (approved):
🎉

PR #654, opened 2025-09-04
chore: Adding Debug Logs

alubbad:
Added debug logs to makeRequest so that all methods report any errors they see. 

Log output looks like this: 
`[2025-09-04T21:59:30Z] DEBUG (data-client): context created with timeout=5s`
`[2025-09-04T21:59:31Z] DEBUG (fixed-count-retry-strategy): Request is not retryable: [method: /cache_client.Scs/DictionarySet, status: NotFound]`
`[2025-09-04T21:59:31Z] ERROR (data-client): gRPC request failed: rpc error: code = NotFound desc = Cache not found, responseMetadata=[map[] map[access-control-allow-credentials:[true] access-control-expose-headers:[grpc-status,grpc-message,grpc-encoding,grpc-accept-encoding,err] content-type:[application/grpc] date:[Thu, 04 Sep 2025 21:59:30 GMT] err:[momento_general_err] vary:[origin, access-control-request-method, access-control-request-headers]]]
panic: NotFoundError: Cache not found
        rpc error: code = NotFound desc = Cache not found`

### DIFF
--- a/momento/scs_data_client.go
+++ b/momento/scs_data_client.go
@@ -143,7 +143,10 @@ func (client scsDataClient) applyMiddlewareResponseHandlers(
 }
 
 func (client scsDataClient) makeRequest(ctx context.Context, r requester) (interface{}, error) {
+	logger := client.loggerFactory.GetLogger("data-client")
+	logger.Debug("%v request made on cache %v", r.requestName(), r.cacheName())
 	if _, err := prepareCacheName(r); err != nil {
+		logger.Error("failed to prepare cache name due to error", err)
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


momento/scs_data_client.go
```diff
@@ -143,7 +143,10 @@ func (client scsDataClient) applyMiddlewareResponseHandlers(
 }
 
 func (client scsDataClient) makeRequest(ctx context.Context, r requester) (interface{}, error) {
+	logger := client.loggerFactory.GetLogger("data-client")
+	logger.Debug("%v request made on cache %v", r.requestName(), r.cacheName())
 	if _, err := prepareCacheName(r); err != nil {
+		logger.Error("failed to prepare cache name due to error", err)
```
anitarua, line 149: I think this log probably isn't needed, prepareCacheName should return an InvalidArgumentError that explains the error clearly enough to the caller

momento/scs_data_client.go
```diff
@@ -143,7 +143,10 @@ func (client scsDataClient) applyMiddlewareResponseHandlers(
 }
 
 func (client scsDataClient) makeRequest(ctx context.Context, r requester) (interface{}, error) {
+	logger := client.loggerFactory.GetLogger("data-client")
```
anitarua, line 146: is it possible to move the logger creation to the constructor instead so it's not initialized with each request?

momento/scs_data_client.go
```diff
@@ -152,33 +155,44 @@ func (client scsDataClient) makeRequest(ctx context.Context, r requester) (inter
 	var err error
 	middlewareRequestHandlers, r, requestMetadata, err = client.applyMiddlewareRequestHandlers(r, requestMetadata)
 	if err != nil {
+		logger.Error("failed to apply middleware request handlers: %v", err)
 		return nil, err
 	}
-
+	logger.Debug("applied middleware request handlers, metadata=%v", requestMetadata)
```
anitarua, line 161: probably don't need several of these affirmative log lines:

- applied middleware request handlers
- request initialized
- context created
- grpc request succeeded
- interpreted grpc response
- applied middleware response handlers
- overall request succeeded

We want to add just enough logs to make debugging easier, so I think targeting the error scenarios make more sense here

anitarua (commented):
I think we can simplify a bit, but otherwise looking good!

PR #678, opened 2026-05-20
fix(topics): bound the subscribe handshake and stop leaking pool slots on timeout

eaddingtonwhite:
Fixes a permanent pool slot leak in the Subscribe handshake and makes its timeout behavior deterministic.

When Subscribe timed out waiting for the first heartbeat, `sendSubscribe` kept running in the background. A late heartbeat would build a subscription that nobody read and its slot leaked permanently; repeated timeouts during an outage could eat the whole pool until the process restarted.

What changed:

- `sendSubscribe` creates the stream context up front and arms a watchdog across the whole handshake, so the deadline also bounds connection establishment. Previously, against an unreachable endpoint, Subscribe returned `TimeoutError` while the reserved slot sat there until gRPC's ~20s connect timeout.
- a `handshakeDecided` CAS decides the race between the watchdog and the first message: a stream the watchdog cancelled is never handed to the caller (it fails with `TimeoutError` and releases the slot), and a delivered stream is never cancelled by the watchdog. When the deadline ties with a successful delivery, Subscribe returns the live subscription instead of a spurious timeout
- `drainAndCloseSubscription` releases the slot of a subscription that completes after Subscribe already returned via the timeout/cancel path
- error codes are deterministic now: canceling your ctx during Subscribe returns `CanceledError` (with `ctx.Err()` as the cause), a handshake timeout returns `TimeoutError`. It used to be a coin flip when both raced.

CI: a new race-tests job runs all the plain unit tests under `-race`, using `-skip '^TestMomento$'` rather than a `-run` allowlist (the allowlist silently dropped tests whose names didn't match the prefix). Also bumps checkout/setup-go to the current node24 majors ahead of GitHub's June 16 forced migration.

Tests cover slot release on timeout, on hung establishment (fails against the old structure), and on a late heartbeat; the cancel error code; the drainer; plus a 64x50 stress test asserting a delivered subscription is never watchdog-cancelled.

Stacked on #677. The race job and the action bumps first run in CI once this retargets to main.

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -131,3 +131,22 @@ jobs:
         env:
           MOMENTO_PORT: 8080
         run: make test-retry
+
+  race-tests:
+    needs: build
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Setup repo
+        uses: actions/checkout@v3
+
+      # Newer Go for better race-detector diagnostics; the rest of the workflow
+      # pins the declared minimum.
+      - name: Install Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.21.x
+
+      # Excludes the Ginkgo entry point (TestMomento) so this job needs no
+      # Momento credentials.
+      - name: Run unit tests under -race
+        run: go test -race -count=1 -run '^Test(Topic|Subscribe|Reservation)' ./internal/... ./momento/...
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -146,7 +146,10 @@ jobs:
         with:
           go-version: 1.21.x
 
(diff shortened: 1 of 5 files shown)


tylerburdsall (approved):

.github/workflows/on-pull-request.yml
```diff
@@ -131,3 +131,22 @@ jobs:
...
         run: make test-retry
+
+  race-tests:
+    needs: build
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Setup repo
+        uses: actions/checkout@v3
+
+      # Newer Go for better race-detector diagnostics; the rest of the workflow
```
anitarua, line 142: nice!

anitarua (approved):

tylerburdsall (approved):

